PR #9, opened 2019-06-24
renames for cli doc generation

kewegner:
We are responding to Helen's requested changes in the earlier PR.

### DIFF
--- a/cmd/completion.go
+++ b/cmd/completion.go
@@ -26,14 +26,15 @@ var bashCompletionCmd = &cobra.Command{
 	Use:   "completion",
 	Short: "Generates bash tab completions",
 	Long: `Running the completion command:
-
+	Command line completion automatically fills in partially typed commands.  This functionality can be added for appsody commands by doing the following:
+	
 	appsody completion > "your-directory"/appsody  
 	Linux
 	
 	1. On a current Linux OS (in a non-minimal installation), bash completion should be available.  
-	2. Place the completion script generated above in your bash completions directory.  
+	2. Place the completion script generated above in your Linux bash completions directory, e.g. /etc/bash_completion.d/. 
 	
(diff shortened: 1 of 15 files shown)


cmd/completion.go
```diff
@@ -26,14 +26,15 @@ var bashCompletionCmd = &cobra.Command{
 	Use:   "completion",
 	Short: "Generates bash tab completions",
 	Long: `Running the completion command:
```
kylegc: How about this:
```
	Long: `Outputs a bash completion script for appsody to stdout.

Bash completion is optionally available for your convenience. It helps you fill out appsody commands when you type the [TAB] key. 

To Install on macOS:
	1. brew install bash-completion
	2. Make sure to update your ~/.bash_profile as instructed
	3. appsody completion > /usr/local/etc/bash_completion.d/appsody

To install on linux:
	1. ??? do you have specific install instructions for linux?`
```

cmd/init.go
```diff
@@ -40,16 +40,16 @@ var whiteListDotFiles = []string{"git", "project", "DS_Store", "classpath", "fac
 
 // initCmd represents the init command
 var initCmd = &cobra.Command{
-	Use:   "init [stack]",
-	Short: "Initialize an appsody project with a stack and template app",
-	Long: `This creates a new appsody project in a local directory or sets up the local dev environment of an existing appsody project. 
+	Use:   "init <stack>",
```
kylegc: Stack is optional, this should go back to `[stack]`

cmd/completion.go
kylegc: It would be nice if the markdown generator would highlight the commands, like using back-ticks in markdown, but I'm not sure how to do that, especially since this code is already surrounded by back-ticks.

PR #12, opened 2019-06-24
Bug fixes

kylegc:
Better error messages on init process
Only search the /project path
Remove the appsody-controller binary


kylegc:
Fixed

Issue #37, opened 2019-07-01
appsody stop command should support different container names


kewegner:
Per @kylegc 
The run/debug/test commands support the --name option to change the default container name. If this option is used, then the appsody stop command won't work because it only uses the default name. We should add support for specifying the container name to appsody stop:

appsody stop my-project-dev-custom
OR
appsody stop --name my-project-dev-custom

kewegner:
We will add a --name flag to handle this.

kewegner:
The stop.go file was modified to add a persistent flag and logic to deal with the --name flag so that when a name is used, the appsody stop command can stop that container with that name name, which must have been specified on the corresponding appsody run.

kewegner:
Finished under the PR above.

PR #54, opened 2019-07-09
add --name to appsody stop 

kewegner:
Added --name parameter to appsody stop


kewegner:
#37 is resolved with this PR.  A test case is added for appsody stop as well.

cmd/stop_test.go
```diff
@@ -0,0 +1,172 @@
...
+		runChannel <- err
+	}()
+
+	// defer the appsody stop to close the docker container
+	defer func() {
+
+		stopOutput, errStop := cmdtest.RunAppsodyCmdExec([]string{"stop"}, projectDir)
+
+		//docker[stop appsody-stop-test
+		if !strings.Contains(stopOutput, "docker[stop appsody-stop-test") {
```
kylegc: Checking that the stop command ran is okay, but I think we really should run a `docker ps` to verify the container was stopped.

cmd/stop.go
```diff
@@ -32,14 +32,19 @@ var stopCmd = &cobra.Command{
...
+		}
 		//dockerRemove(imageName) is not needed due to --rm flag
 		os.Exit(1)
 	},
 }
+var stopContainerName string
 
 func init() {
 	rootCmd.AddCommand(stopCmd)
+	stopCmd.PersistentFlags().StringVar(&stopContainerName, "name", "", "The name of your development container.")
```
kylegc: We should try to reuse the `name` flag from dev_common. See the addDevCommonFlags() func as an example. We'll need to pull out the name from there into a different common flag.

cmd/cmdtest/testutils.go
```diff
@@ -209,3 +209,67 @@ func AddLocalFileRepo(repoName string, repoFilePath string) (string, func(), err
 
 	return repoURL, cleanupFunc, err
 }
+
+// RunDockerCmdExec runs the docker command with the given args in a new process
+// The stdout and stderr are captured, printed, and returned
+// args will be passed to the docker command
+// workingDir will be the directory the command runs in
+func RunDockerCmdExec(args []string, workingDir string) (string, error) {
+	execDir, err := os.Getwd()
```
kylegc: `docker` commands don't rely on the working directory, so you can remove all the code related to this.

cmd/stop.go
```diff
@@ -32,14 +32,18 @@ var stopCmd = &cobra.Command{
 		Info.log("Stopping development environment")
 		projectDir := getProjectDir()
 		imageName := fmt.Sprintf("%s-dev", filepath.Base(projectDir))
-
-		dockerStop(imageName)
+		if containerName == "" {
+			dockerStop(imageName)
```
kylegc: containerName will always be set, you can remove this if check and the imageName and always use containerName

PR #67, opened 2019-07-10
Remove deploy condition

kylegc:


chilanti:
Errored out - need to revise travis.yml

PR #89, opened 2019-07-18
Projectname refactor

chilanti:
Scrubbed all the CLI code for references to projectName and projectDir. Modified all instances where the projectName was populated directly from projectDir - to use getProjectName(), which returns lowercased base projectDir.

### DIFF
--- a/cmd/dev_common.go
+++ b/cmd/dev_common.go
@@ -54,10 +54,10 @@ func buildCommonFlags() {
 
 		if perr != nil {
 			if pmsg, ok := perr.(*NotAnAppsodyProject); ok {
-				Debug.log("Not a valid Appsody project - continuing: ", perr)
-				projectName = ""
+				Debug.log("Cannot retrieve the project name - continuing: ", perr)
 			} else {
 				Error.log("Error occurred retrieving project name... exiting: ", pmsg)
+				os.Exit(1)
 			}
 		}
 		//defaultName := filepath.Base(curDir) + "-dev"
@@ -92,7 +92,7 @@ func addDevCommonFlags(cmd *cobra.Command) {
 func commonCmd(cmd *cobra.Command, args []string, mode string) {
 	projectDir, perr := getProjectDir()
 	if perr != nil {
-		Error.log("The current directory is not a valid appsody project. Run appsody init <stack> to create one: ", perr)
+		Error.log(perr)
 		os.Exit(1)
 	}
 	projectConfig := getProjectConfig()
--- a/cmd/dev_common.go
+++ b/cmd/dev_common.go
@@ -50,12 +50,21 @@ func buildCommonFlags() {
 		//	Error.log("Error getting current directory ", err)
 		//	os.Exit(1)
 		//}
+		projectName, perr := getProjectName()
 
+		if perr != nil {
+			if pmsg, ok := perr.(*NotAnAppsodyProject); ok {
+				Debug.log("Not a valid Appsody project - continuing: ", perr)
+				projectName = ""
+			} else {
+				Error.log("Error occurred retrieving project name... exiting: ", pmsg)
+			}
+		}
 		//defaultName := filepath.Base(curDir) + "-dev"
-		defaultName := getProjectName() + "-dev"
+		defaultName := projectName + "-dev"
 		nameFlags.StringVar(&containerName, "name", defaultName, "Assign a name to your development container.")
 		//defaultDepsVolume := filepath.Base(curDir) + "-deps"
-		defaultDepsVolume := getProjectName() + "-deps"
+		defaultDepsVolume := projectName + "-deps"
 		commonFlags.StringVar(&dockerNetwork, "network", "", "Specify the network for docker to use.")
 		commonFlags.StringVar(&depsVolumeName, "deps-volume", defaultDepsVolume, "Docker volume to use for dependencies. Mounts to APPSODY_DEPS dir.")
 		commonFlags.StringArrayVarP(&ports, "publish", "p", nil, "Publish the container's ports to the host. The stack's exposed ports will always be published, but you can publish addition ports or override the host ports with this option.")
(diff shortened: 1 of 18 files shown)


chilanti:
Made those changes - reused the isHelpCommand() in getProjectName.

cmd/utils.go
```diff
@@ -186,31 +194,36 @@ func mountExistsLocally(mount string) bool {
...
 		Error.log("Error getting current directory ", err)
-		os.Exit(1)
+		return "", err
 	}
 	appsodyConfig := filepath.Join(dir, ConfigFile)
 	projectDir, err := exists(appsodyConfig)
 	if err != nil {
 		Error.log(err)
-		os.Exit(1)
+		return "", err
```
kylegc: I think this should return `dir, err`

cmd/utils.go
```diff
@@ -186,31 +194,36 @@ func mountExistsLocally(mount string) bool {
...
 
-		Error.log("Current dir is not an appsody project. " +
-			"Run `appsody init <stack>` to setup an appsody project. Run `appsody list` to see the available stacks.")
-
-		os.Exit(1)
+		Debug.log("Current dir is not an appsody project.")
+		// +
+		// "Run `appsody init <stack>` to setup an appsody project. Run `appsody list` to see the available stacks.")
+		var e NotAnAppsodyProject = "The current directory is not a valid appsody project."
+		return "", &e
```
kylegc: I think this should return `dir, &e`

cmd/utils.go
```diff
@@ -186,31 +194,36 @@ func mountExistsLocally(mount string) bool {
...
 	}
 	if !projectDir {
 
-		Error.log("Current dir is not an appsody project. " +
-			"Run `appsody init <stack>` to setup an appsody project. Run `appsody list` to see the available stacks.")
-
-		os.Exit(1)
+		Debug.log("Current dir is not an appsody project.")
+		// +
+		// "Run `appsody init <stack>` to setup an appsody project. Run `appsody list` to see the available stacks.")
```
kylegc: Reason for not including this text? I'm pretty sure it was added for clarity based on user feedback/issue.

cmd/utils.go
```diff
@@ -186,31 +194,36 @@ func mountExistsLocally(mount string) bool {
...
-	return dir
+	return dir, nil
 }
 
 func getProjectConfig() ProjectConfig {
 	if projectConfig == nil {
-		dir := getProjectDir()
+		dir, perr := getProjectDir()
+		if perr != nil {
+			Error.log("The current directory is not a valid appsody project. Run appsody init <stack> to create one: ", perr)
```
kylegc: Why not `Error.log(perr)`?

cmd/utils.go
```diff
@@ -225,10 +238,13 @@ func getProjectConfig() ProjectConfig {
 	}
 	return *projectConfig
 }
-func getProjectName() string {
-	projectDir := getProjectDir()
+func getProjectName() (string, error) {
+	projectDir, err := getProjectDir()
+	if err != nil {
+		return "", err
```
kylegc: I think this should still return the project name if it can, which means we can continue this function and return `err` at the end.

cmd/dev_common.go
```diff
@@ -42,20 +42,30 @@ var nameFlags *flag.FlagSet
...
+		//	os.Exit(1)
+		//}
+		projectName, perr := getProjectName()
+
+		if perr != nil {
+			if pmsg, ok := perr.(*NotAnAppsodyProject); ok {
+				Debug.log("Not a valid Appsody project - continuing: ", perr)
+				projectName = ""
+			} else {
+				Error.log("Error occurred retrieving project name... exiting: ", pmsg)
```
kylegc: Did you mean to use an os.Exit() after this?

cmd/dev_common.go
```diff
@@ -42,20 +42,30 @@ var nameFlags *flag.FlagSet
...
+		//if err != nil {
+		//	Error.log("Error getting current directory ", err)
+		//	os.Exit(1)
+		//}
+		projectName, perr := getProjectName()
+
+		if perr != nil {
+			if pmsg, ok := perr.(*NotAnAppsodyProject); ok {
+				Debug.log("Not a valid Appsody project - continuing: ", perr)
+				projectName = ""
```
kylegc: No need for this, use what getProjectName() returned even if it returns a perr

cmd/dev_common.go
```diff
@@ -80,13 +90,17 @@ func addDevCommonFlags(cmd *cobra.Command) {
 }
 
 func commonCmd(cmd *cobra.Command, args []string, mode string) {
-
+	projectDir, perr := getProjectDir()
+	if perr != nil {
+		Error.log("The current directory is not a valid appsody project. Run appsody init <stack> to create one: ", perr)
```
kylegc: Maybe just print the perr? I'm worried about duplicate messages...

cmd/extract.go
```diff
@@ -32,6 +32,14 @@ var extractCmd = &cobra.Command{
 	Long: `This copies the full project, stack plus app, into a local directory
 in preparation to build the final docker image.`,
 	Run: func(cmd *cobra.Command, args []string) {
+		// Make sure we are in an Appsody project
+		projectDir, perr := getProjectDir()
+
+		if perr != nil {
+			Error.log("The current directory is not a valid appsody project. Run appsody init <stack> to create one: ", perr)
```
kylegc: Same comment as above, I'm worried about duplicate messages.

cmd/extract.go
```diff
@@ -171,11 +176,22 @@ in preparation to build the final docker image.`,
...
+	//		Error.log("Error getting current directory ", err)
+	//	os.Exit(1)
+	//}
+	//defaultName := filepath.Base(curDir) + "-extract"
+	projectName, perr := getProjectName()
+
+	if perr != nil {
+		if pmsg, ok := perr.(*NotAnAppsodyProject); ok {
+			Debug.log("Not a valid Appsody project - continuing: ", perr)
+			projectName = ""
```
kylegc: Same comments as above, don't set the projectName here, and missing os.Exit()

cmd/init.go
```diff
@@ -147,7 +147,11 @@ func init() {
 //Runs the .appsody-init.sh/bat files if necessary
 func install() {
 	Info.log("Setting up the development environment")
-	projectDir := getProjectDir()
+	projectDir, perr := getProjectDir()
+	if perr != nil {
+		Error.log("The current directory is not a valid appsody project. Run appsody init <stack> to create one: ", perr)
```
kylegc: Same comment as above, duplicate message?

cmd/utils.go
```diff
@@ -131,7 +135,11 @@ func getVolumeArgs() []string {
 		homeDir = homeDirOverride
 		homeDirOverridden = true
 	}
-	projectDir := getProjectDir()
+	projectDir, perr := getProjectDir()
+	if perr != nil {
+		Error.log("The current directory is not a valid appsody project. Run appsody init <stack> to create one: ", perr)
```
kylegc: Same comment as above, duplicate message?

chilanti: This has been resolved now.

PR #97, opened 2019-07-24
Make targets for dependency management

mhmxs:


Makefile
```diff
@@ -27,6 +28,16 @@ build_name = $(COMMAND)-$(VERSION)-$(os)-amd64
 build_binary = $(build_name)$(BINARY_EXT_$(os))
 package_binary = $(COMMAND)$(BINARY_EXT_$(os))
 
+.PHONY: deps
```
kylegc: I'm okay with adding the `dep` tool and an `ensure` target, but instead of a `deps` make target to install it, lets be consistent with how we install golangci-lint. That is, make a variable for DEP_BINARY and a make target for the binary itself. Then the ensure target can depend on the DEP_BINARY to install it without additional commands and only when needed.

Gopkg.toml
```diff
@@ -55,11 +55,3 @@
 [prune]
   go-tests = true
   unused-packages = true
-
-[[constraint]]
```
kylegc: Why remove these constraints? We hit issues in the past when these versions were not exactly lined up. Unless there's a good reason to remove these, I'd rather be safe than sorry and leave them here.

mhmxs: They were indirect dependencies, my 6th sense felt you had issues in the past, but i did a try :) I will remove this from my PR thanks.

mhmxs:
@kylegc I updateted the pull request base on your comments. Is it good for you?

Makefile
```diff
@@ -82,11 +84,23 @@ $(GOLANGCI_LINT_BINARY):
 	# see https://github.com/golangci/golangci-lint
 	curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(GO_PATH)/bin $(GOLANGCI_LINT_VERSION)
 
+# not PHONY, installs deps if it doesn't exist
+$(DEP_BINARY):
+	curl https://raw.githubusercontent.com/golang/dep/master/install.sh | DEP_RELEASE_TAG=$(DEP_RELEASE_TAG) sh
+
+.PHONY: ensure
```
kylegc: Instead of using the `ifeq` check, can't you use make target dependencies like this?
```
.PHONY: ensure
ensure: $(DEP_BINARY)
	$(DEP_BINARY) ensure
```

mhmxs: Sure and fixed

Issue #36, opened 2019-07-01
Remove os.Exit() from cli so that we can call unit tests without exiting the test framework


kewegner:
Per @kylegc 

We need to remove os.Exit() from the cli so that we can call unit tests without exiting the test framework.

PR #113, opened 2019-07-25
Remove os.exit(1) from repo_remove.go

kewegner:
Partial fulfillment of #36 removes os.exit(1) call


kewegner:
Issue with changed files.  Will do again

PR #123, opened 2019-07-26
removed os.exit(1) from repo_remove, added test case #2

kewegner:


kewegner:
PR in error

kewegner:
We need to hold off on this PR until Monday.  We may not need it, but its removal cause build issues.

chilanti:
This is empty - closing.

PR #124, opened 2019-07-26
Remove os exit from docs and list #3

kewegner:
Partial fulfillment of #36 
Removed os.exit from list.go and docs.go


kewegner:
Incorrect PR

PR #135, opened 2019-07-30
removed os.exit(1) from extract

kewegner:
Partial fulfillment of #36 
Removed os.exit from extract.go
Changed init.go and build.go to use extractCmd.RunE.

### DIFF
--- a/cmd/extract.go
+++ b/cmd/extract.go
@@ -114,9 +114,8 @@ in preparation to build the final docker image.`,
 			err = execAndWaitReturnErr(cmdName, cmdArgs, Debug)
 			if err != nil {
 
-				Error.log("docker create command failed: ", err)
 				dockerRemove(extractContainerName)
-				return err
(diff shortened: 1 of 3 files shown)


cmd/extract.go
```diff
@@ -117,11 +112,12 @@ in preparation to build the final docker image.`,
 			cmdArgs = append([]string{"create"}, cmdArgs...)
 			cmdArgs = append(cmdArgs, stackImage)
 			err = execAndWaitReturnErr(cmdName, cmdArgs, Debug)
-
 			if err != nil {
+
 				Error.log("docker create command failed: ", err)
 				dockerRemove(extractContainerName)
-				os.Exit(1)
+				return err
```
chilanti: Why is the pattern different here? I see we aren't wrapping `err` in an `errors.Errorf`... and we left the `Error.log()`... is there a good reason?

PR #136, opened 2019-07-31
Remove os exit from utils.go and parent callers

kewegner:
Partial fullfillment of #36 
Removed os.exit calls from utils.go and callers of utils.go

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -49,7 +49,10 @@ var buildCmd = &cobra.Command{
 		}
 		cmdName := "docker"
 		cmdArgs := []string{"build", "-t", buildImage, "-f", dockerfile, extractDir}
-		execAndWait(cmdName, cmdArgs, DockerLog)
+		execError := execAndWait(cmdName, cmdArgs, DockerLog)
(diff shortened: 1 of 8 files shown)


cmd/build.go
```diff
@@ -46,7 +49,10 @@ var buildCmd = &cobra.Command{
 		}
 		cmdName := "docker"
 		cmdArgs := []string{"build", "-t", buildImage, "-f", dockerfile, extractDir}
-		execAndWait(cmdName, cmdArgs, DockerLog)
+		execError := execAndWait(cmdName, cmdArgs, DockerLog)
+		if execError != nil {
+			return execError
```
chilanti: I think this should be wrapped.

kewegner: The problem with the wrapping is that it makes the stack look like it came from the caller rather than the function that has the original error.
In the future we need to discuss wrapping of original golang type errors perhaps.

chilanti: OK - let's leave this as it is right now - we'll revisit tracing in another occasion.

Issue #133, opened 2019-07-30
Include repo name in `appsody list`


ianpartridge:
**Is your feature request related to a problem? Please describe.**

Currently, when you run `appsody list` you see output like:

```
NAME             	VERSION	DESCRIPTION
nodejs-express   	0.2.2  	Express web framework for Node.js
swift            	0.1.0  	Swift runtime
java-microprofile	0.2.3  	Microprofile using AdoptOpenJDK and Maven
java-spring-boot2	0.3.2  	Spring Boot using IBM Java SDK and Maven
nodejs           	0.2.2  	Node.js Runtime
```

This does not indicate which repo the stacks came from. If you have installed additional repos through `appsody repo add` you will have stacks originating from multiple repos and it's not clear which repo the stacks came from.

**Describe the solution you'd like**

Indicate which repo the stack came from, perhaps through an additional REPO column:

```
REPO       NAME             	VERSION	DESCRIPTION
appsodyhub nodejs-express   	0.2.2  	Express web framework for Node.js
appsodyhub swift            	0.1.0  	Swift runtime
appsodyhub java-microprofile	0.2.3  	Microprofile using AdoptOpenJDK and Maven
appsodyhub java-spring-boot2	0.3.2  	Spring Boot using IBM Java SDK and Maven
appsodyhub nodejs           	0.2.2  	Node.js Runtime
```

An alternative would be to group the output by repo, for example list the stacks from repo1 followed by the stack from repo2.

Note also that currently if you have two stacks with the same name from multiple repos, only one is shown in `appsody list` and it's not clear how to choose between them with `appsody init`.

ianpartridge:
It might be best to group the output by repo, as today most people aren't using `appsody repo add`.

ianpartridge:
Something like:

```
$ appsody list
REPO
appsodyhub

ID               	VERSION	DESCRIPTION
swift            	0.1.0  	Swift runtime
java-microprofile	0.2.3  	Microprofile using AdoptOpenJDK and Maven
java-spring-boot2	0.3.2  	Spring Boot using IBM Java SDK and Maven
nodejs           	0.2.2  	Node.js Runtime
nodejs-express   	0.2.2  	Express web framework for Node.js
```

Also, can we change the `ID` column header to be `STACK`? That's what they are.

Finally, can the ordering of the stacks be made consistent?  Currently they come out randomly each time you run `appsody list`.  This is odd, please alphabetise them.

PR #138, opened 2019-08-01
Added initial support for multiple repos

chilanti:
This PR introduces the following functionality:
1) Supports multiple repositories in an adequate way. Now, the same stack name may be present in multiple repositories.
1) Introduces the concept of *default repo*. The default repo is marked with an asterisk when you run `appsody repo list`. The default repo can be modified by manually editing `~/.appsody/repository/repository.yaml` and moving the `default:true` yaml element to the repo that you want to make the default repo.
1) Now, `appsody list` lists all the stacks that are available with an indication of the repo they come from.
1) There's a new `appsody list <repo>` variation of the command, which only shows the stacks in the specified repo (if it exists).
1) To create an appsody project, you can keep using the old `appsody init <stack>`, with the understanding that Appsody will try to find <stack> in the default repo.
1) If you want to initialize a project that is located in a repo other than the default one, you can now issue `appsody init <repo>/<stack>` (i.e. you can fully qualify the stack you want to initialize). Appsody will look into the `<repo>` you indicate.
1) The command `appsody repo remove <repo>` won't allow you to remove the default repo.
1) A new installation of Appsody sets the first repo as the default repo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ stages:
   - name: lint
   - name: test
   - name: deploy
-    if: branch = master AND tag IS present
+    if: tag IS present
 
 install-controller: &install-controller
   install:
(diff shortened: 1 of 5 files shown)


chilanti:
Fulfills #133

kewegner:
I reviewed per above, and the said functionality looks good.

Issue #141, opened 2019-08-01
Appsody should include the notion of "default repository"


chilanti:
**Is your feature request related to a problem? Please describe.**
Currently, there is no assumptions on which repository is the "default repository", which would force users to always qualify their `appsody init` requests against a specific repo.

**Describe the solution you'd like**
Appsody should allow the designation of a "default repository", against which requests are issued lacking a specific indication (ex: `appsody init <stack>` assumes you want to create a project from the <stack> in the default repo.)
`appsody repo list` should provide an indication of which repo is the default repo (by means of an asterisk by the repo name)
`appsody remove repo <repo>` should not allow you to remove the default repo.

Initially, it's ok to have the user manually edit `~/.appsody/repository/repository.yaml` to change the default repo.

Issue #140, opened 2019-08-01
Appsody list command should allow listing a single repo


chilanti:
**Is your feature request related to a problem? Please describe.**
Currently, `appsody list` takes no parms and lists all the stacks - no way to filter by repo.

**Describe the solution you'd like**
`appsody list` should include an optional parameter <repo> and list all the stacks in that repo.

Issue #139, opened 2019-08-01
Appsody init command to support multiple repos


chilanti:
**Is your feature request related to a problem? Please describe.**
Currently, there is no way to indicate what repo a stack should be instantiated from. `appsody init` only allows you to specify the stack name - not the repo.

**Describe the solution you'd like**
`appsody init` should allow a way to fully qualify the stack:
`appsody init <repo>/<stack>`. 
Lacking the repo indicator, Appsody should assume the request goes against the default repo.

PR #144, opened 2019-08-01
CLI repo and init: add support for V2 repository schema

chilanti:
This PR allows the CLI to deal with a mixture of V1 and V2 repositories.


kewegner:
Fullfillment of #139 #140 #141

PR #149, opened 2019-08-02
Handling missing default repo and situations where repo not found

chilanti:
"Legacy" CLI users (i.e. 0.2.8 and prior) may not have a default repo. This PR sets the default repo to either the only one they may have, or "appsodyhub" if it's there, or the first one in their list if they have multiple repos and no "appsodyhub".
Also handles situations in which the repo that's specified in requests isn't found.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ stages:
   - name: lint
   - name: test
   - name: deploy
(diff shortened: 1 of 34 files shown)


chilanti:
Erroneous PR - closing.

Issue #152, opened 2019-08-05
Add capability to set-default repository


mhamwala:
**Is your feature request related to a problem? Please describe.**
Currently, we do not have a CLI command that allows you to set your desired default repository.

**Describe the solution you'd like**
A capability to set your default repository with a command like the following:
```
appsody repo set-default <repo-name>
``` 

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

PR #158, opened 2019-08-05
Added option to set default repository

skoh7645:
Added capability to set your default repository with a command like the following:
```
appsody repo set-default <repo-name>
```
Fixes :https://github.com/appsody/appsody/issues/152


kewegner:
I would suggest we do some testing when the index.yaml is V1 as well as V2 for this.

skoh7645:
@kewegner I wasn't sure exactly how to test this, I added another yaml file with V2 for testing.  Can you please take a look at the travis build.

PR #167, opened 2019-08-06
Add version check for repository yaml files.

### LINKED ISSUE #157: Add warning message when apiLevel is not supported by cli
While reading various config files and indexes, the cli uses apiVersion field to determine how to best parse the contents. If we find that the apiVersion is now something we know of then the cli should issue a warning and also suggest upgrading the cli.

kewegner:
Fixes #157 

Adds version check for repository yaml.

Updates help text for multi-repo support.


cmd/list.go
```diff
@@ -21,11 +21,9 @@ import (
 
 // listCmd represents the list command
 var listCmd = &cobra.Command{
-	Use:   "list [repo]",
+	Use:   "list [repository]",
 	Short: "List the Appsody stacks available to init",
-	Long: `This command lists all the stacks available in your repositories, if you omit the 
-	optional [repo] parameter. If you specify the repository name [repo], only the stacks in that
-	repositories will be listed.`,
+	Long:  `This command lists all the stacks available in your repositories. If you omit the  optional [repository] parameter, the stacks for all the repositories are listed. If you specify the repository name [repo], only the stacks in that repositories will be listed.  An asterisk denotes the default repository.`,
```
tnixa: need to change [repo] to [repository]

Issue #98, opened 2019-07-24
Cannot find source file /var/folders/.../exe/appsody-controller to copy during test


mhmxs:
**Describe the bug**
During `make test` i receive the following error:
```
[Debug] Attempting to load the controller from /Users/mhmxs/.appsody/appsody-controller
[Debug] Didn't find the controller in .appsody - copying from the binary directory...
[Debug] Binary location /var/folders/z2/f__84fm11736mw001xd06r4m0000gn/T/go-build846635544/b001/exe
[Debug] Attempting to copy the source controller from: /var/folders/z2/f__84fm11736mw001xd06r4m0000gn/T/go-build846635544/b001/exe/appsody-controller
[Error] Cannot find source file /var/folders/z2/f__84fm11736mw001xd06r4m0000gn/T/go-build846635544/b001/exe/appsody-controller to copy
```

**To Reproduce**
1. Remove `~/.appsody/appsody-controller`
2. Execute `make test`

**Expected behavior**
The test must compile on every machine not just where `~/.appsody/appsody-controller` exists.

**Actual behaviour**
Doesn't compile because `make test` doesn't create `appsody-controller` during the test.

**Environment Details (please complete the following information):**
 - OS: macOS 10.14.6 

**If applicable please specify:**
 - CLI version: latest master

chilanti:
Hi - this is expected. The controller needs to be built separately. We've documented it in https://github.com/appsody/appsody/blob/master/build.md. Also - notice that the controller is a Linux build - always - because it is meant to run inside the stack containers (as opposed to the host system). The build process for the controller is detailed over here: https://github.com/appsody/controller. Let me know if you have comments or questions. Thanks!

kylegc:
Another option is to have the makefile check for the controller and download it if it doesn't exist. @tnixa is working on something similar to run the tests on Travis CI. We'll see if we can incorporate it in the local testing as well.

mhmxs:
I also wanted to suggest the extra check in makefile. But downloading it is also a nice way. Thanks, i will be happy with that.

neeraj-laad:
Closing this issue for now. We can re-open if we think of better ideas in future.

PR #181, opened 2019-08-09
hard code controller url

tnixa:
the $GH_ORG usage in the stacks travis build was not working so trying to hard code it


kewegner:
We will discuss on Monday when folks return.  I have a proposal re: the controller URL.  
I have created some Makefile code that checks if the controller exists in .appsody and  if GH_ORG is set and the controller binary doesn't exists and  GH_ORG is not set  it will construct a URL with "appsody" in it, otherwise if it is defined and the controller does not exist, it constructs the controller URL using GH_ORG.  This would solve the issue in this PR, and it would also address #98.  It should also be compatible with .travis.yaml calling install-controller.

PR #183, opened 2019-08-12
Copy during update corrupts controller binary on MacOS, add checksum to prevent this

kewegner:
appsody CLI now does checksum on controller binaries to see if the copy to .appsody is really necessary to lessen the likelihood of this scenario occuring.


cmd/utils.go
```diff
@@ -831,3 +834,41 @@ func execAndWaitWithWorkDirReturnErr(command string, args []string, logger appso
...
+
+func checksum256TestFile(newFileName string, oldFileName string) (bool, error) {
+	var checkValue bool
+	fmt.Println("Checksum 256")
+	fmt.Println("oldFile", oldFileName)
+	fmt.Println("newFile", newFileName)
+
+	oldFile, err := os.Open(oldFileName)
+	if err != nil {
+		return false, errors.Errorf("File open failed for %s controller binary: %v", oldFileName, err)
```
neeraj-laad: Perhaps log in a util function could be more generic and not refer to `controller binary`. We might be able to reuse this function for diff files in future

cmd/utils.go
```diff
@@ -831,3 +834,41 @@ func execAndWaitWithWorkDirReturnErr(command string, args []string, logger appso
 	}
 	return err
 }
+
+func checksum256TestFile(newFileName string, oldFileName string) (bool, error) {
+	var checkValue bool
+	fmt.Println("Checksum 256")
```
neeraj-laad: Do these get printed on console when user runs appsody cli? Should these be info logs ?

cmd/utils.go
```diff
@@ -831,3 +834,41 @@ func execAndWaitWithWorkDirReturnErr(command string, args []string, logger appso
...
+
+	oldFile, err := os.Open(oldFileName)
+	if err != nil {
+		return false, errors.Errorf("File open failed for %s controller binary: %v", oldFileName, err)
+
+	}
+	defer oldFile.Close()
+
+	oldSha256 := sha256.New()
+	if _, err := io.Copy(oldSha256, oldFile); err != nil {
```
neeraj-laad: Could we extract the bit about get checksum for a file in a separate util function? I see we do this twice here. and might make the code cleaner.

cmd/dev_common.go
```diff
@@ -153,21 +153,37 @@ func commonCmd(cmd *cobra.Command, args []string, mode string) error {
...
-			chmodErr := os.Chmod(destController, 0755)
-			if chmodErr != nil {
-				return errors.Errorf("Cannot make the controller  executable - exiting: %v", chmodErr)
+		controllerExists, existsErr := exists(destController)
+		if existsErr != nil {
+			return existsErr
+		}
+		Debug.log("appsody-controller exists: ", controllerExists)
+
+		checksum, checksumErr := checksum256TestFile(filepath.Join(binaryLocation, "appsody-controller"), destController)
```
neeraj-laad: calling this variable checksumMatch or filesMatch might make it easier to follow the code.

PR #189, opened 2019-08-15
Support templates +  URL. Fixes #188

### LINKED ISSUE #188: Allow templates to be fetched from inside the image
See appsody/stacks#181

Is your feature request related to a problem? Please describe.

As stacks are customized and distributed, it becomes increasingly likely that templates and stacks will get out of sync (since they are published as separate artifacts). Templates are comparatively thin, but often depend on the build behavior of the stack (at least in a "this template requires this minimum version of the stack" sense).

Describe the solution you'd like

I think the templates should be packaged inside the stack. That way, you can extract/use a template from any stack image, and you never get into the situation where a project can't be created because the two separate publishing mechanisms are out of sync.

From an index.yaml POV, this could look something like this:

  - id: java-spring-boot2
    name: Spring Boot®
    version: 0.4.0
    description: Spring Boot using OpenJ9 and Maven
    license: Apache-2.0
    maintainers:
      - name: Erin Schnabel
        email: schnabel@us.ibm.com
        github-id: ebullient
    default-template: default
    templates:
      - id: default
        url: image:/templates/default.tgz
      - id: kotlin
        url: image:/templates/kotlin.tgz
    image: appsody/java-spring-boot2:0.4.0

If the image: URL is present, then the image tag is used to find the templates (this assumes one image per appsody stack).

ebullient:
POC to support `image:` URLs for in-stack templates.

```  - id: java-spring-boot2
    name: Spring Boot®
    version: 0.4.0
    description: Spring Boot using OpenJ9 and Maven
    license: Apache-2.0
    maintainers:
      - name: Erin Schnabel
        email: schnabel@us.ibm.com
        github-id: ebullient
    default-template: default
    templates:
      - id: default
        url: image:/templates/default.tgz
      - id: kotlin
        url: image:/templates/kotlin.tgz
    image: appsody/java-spring-boot2:0.4.0
```

This works against PR appsody/stacks#180

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -85,7 +85,10 @@ setup the local dev environment.`,
 		var index *RepoIndex
 
 		if len(args) >= 1 {
-			var projectName string
+			var templateURL string = ""
+			var stackImage string = ""
+			var templateName string = "none"
+
 			projectParm := args[0]
 
 			repoName, projectType, err := parseProjectParm(projectParm)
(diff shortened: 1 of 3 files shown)


ebullient:
Too many changes.. will come at this a different way.

PR #190, opened 2019-08-15
Add support for customizing the project directory

seabaylea:
The `appsody extract` and therefore `appsody build` commands currently make the assumption that the application sits inside the `/project` directory inside the container. Whilst this has been true for all Appsody Stacks so far, it may not be in the future.

An example of this is "Traditional WebSphere", where the standard docker images provide assets in `/work/config`.

This adds an optional `APPSODY_PROJECT_DIR` environment variable, which defaults to `/project`.

### DIFF
--- a/cmd/extract.go
+++ b/cmd/extract.go
@@ -107,7 +107,10 @@ in preparation to build the final docker image.`,
 			return dockerPullErr
 		}
 
-		containerProjectDir := "/project"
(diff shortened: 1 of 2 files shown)


neeraj-laad:
@seabaylea @kewegner  We should update the environment variable docs https://github.com/appsody/website/blob/master/content/docs/stacks/environment-variables.md along with this change

PR #196, opened 2019-08-16
--docker-options flag added

### LINKED ISSUE #116: Add docker run options for appsody run
In order to enable more complex development environments, developers need the possibility to add more options for the `docker run` that is done internally. Setups that, for example, require more complex networking, would need to specify a `--network`. In order to be more flexible, we should allow to specify additional run arguments.

The ability to specify additional docker run arguments, such as `--network`, `--memory`, `--privileged`, and potentially others.

kewegner:
Fixes #116 

Adds the --docker-options flag to allow users to send options to docker command (run, build) via the appsody command line.

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -16,10 +16,13 @@ package cmd
 
 import (
 	"path/filepath"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
+
+var dockerBuildOptions string
 
 // buildCmd provides the ability run local builds, or setup/delete Tekton builds, for an appsody project
 var buildCmd = &cobra.Command{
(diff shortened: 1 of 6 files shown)


cmd/build.go
```diff
@@ -16,11 +16,30 @@ package cmd
...
 	"path/filepath"
+	"regexp"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
+func checkDockerBuildOptions(options []string) error {
+	buildOptionsTest := "(^((-t)|(--tag)|(-f)|(--file))(=?$)|(=.*))"
```
chilanti: I am not sure we should try and match for `=`.
Example: `docker build --file Dockerfile --tag mytag` would result in no matches for `--file Dockerfile --tag mytag`) - or am I confused?

chilanti: I did some testing - and this seems to yield matches for the blacklisted options: `(^((-t )|(--tag )|(-f )|(--file ))(.*))`

cmd/dev_common.go
```diff
@@ -38,10 +38,25 @@ var depsVolumeName string
...
 var publishAllPorts bool
 var dockerNetwork string
-
+var dockerOptions string
 var nameFlags *flag.FlagSet
 var commonFlags *flag.FlagSet
 
+func checkDockerRunOptions(options []string) error {
+	fmt.Println("testing docker options", options)
+	runOptionsTest := "(^((-p)|(--publish)|(--publish-all)|(-P)|(-u)|(--user)|(--name)|(--network)|(-t)|(--tty)|(--rm)|(--entrypoint)|(-v)|(--volume)|(-e)|(--env))(=?$)|(=.*))"
```
chilanti: Same as above?

cmd/build.go
kewegner: The reason the code works is the string is split. and each individual piece is tested, not the whole string at once.

chilanti: Sorry - I had missed that.

PR #202, opened 2019-08-20
Add support for the Appsody-Operator

seabaylea:
This adds support for the Appsody-operator into the deploy commands by:
* Adding a new `operator` command with `install` and `uninstall` to support installing and uninstalling the operator
* Updating the `deploy` command to work with a default `app-deploy.yaml` that is extracted from inside the Appsody Stack associated with the project
  * This falls back to the existing Knative Serving approach if no file is found in the stack
* Adding a `deploy --generate` flag to generate the `app-deploy.yaml` but not deploy the app
* Add a new `delete` command to remove the deployed project


cmd/deploy.go
```diff
@@ -15,107 +15,395 @@
...
 
-var namespace, tag string
-var push bool
+var configFile, namespace, operatorspace, watchspace, tag string
+var generate, force, push bool
 
 var deployCmd = &cobra.Command{
 	Use:   "deploy",
-	Short: "Build and deploy your Appsody project on a local Kubernetes cluster",
+	Short: "Build and deploy your Appsody project to Kubernetes cluster",
```
kewegner: to "a" or "your" Kubernetes cluster???

cmd/deploy.go
```diff
@@ -15,107 +15,395 @@
...
-				if err != nil {
-					Warning.log("The exposed port is not a valid integer. The service will not be accessible.")
-					port = 0
+		}
+
+		exists, err := exists(configFile)
+		if err != nil {
+			return errors.Errorf("Error checking status of %s", configFile)
+		}
+		if !exists || (exists && force == true) {
```
kewegner: force== true triggers a lint error

cmd/deploy.go
```diff
@@ -15,107 +15,395 @@
...
+			if os.IsNotExist(err) {
+				return errors.Errorf("Config file does not exist %s. ", configFile)
+			}
+			return errors.Errorf("Failed reading file %s", configFile)
+		}
+		defer yamlFile.Close()
+		scanner := bufio.NewScanner(yamlFile)
+		scanner.Split(bufio.ScanLines)
+		var txtlines []string
+		for scanner.Scan() {
```
kewegner: I assume the yaml structure is too complicated to create structs to parse it like we do the index.yaml files?

cmd/deploy.go
```diff
@@ -15,107 +15,395 @@
...
+		if portsErr != nil {
+			return portsErr
+		}
+		if len(portsStr) == 0 {
+			//No ports exposed
+			Warning.log("This container exposes no ports. The service will not be accessible.")
+			port = 0 //setting this to 0
+		} else {
+			portStr := portsStr[0]
+			Warning.log("Picking the first exposed port as the KNative service port. This may not be the correct port.")
```
kewegner: I agree with your comment.

cmd/deploy.go
```diff
@@ -15,107 +15,395 @@
...
+
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
+	go func() {
+		<-c
+		err := dockerStop(containerName)
+		if err != nil {
+			Error.log(err)
+		}
+		//dockerRemove(containerName) is not needed due to --rm flag
```
kewegner: I do not see any --rm flag, is it required?  
There is no wait after dockerStop, is one needed (similar to appsody run?)

cmd/deploy.go
```diff
@@ -15,107 +15,395 @@
...
+		os.Exit(1)
+	}()
+	cmdName = "docker"
```
kewegner: Future: no change needed, this code looks like the code we have in extract, I wonder if there is a potential here for future refactoring?

cmd/utils.go
```diff
@@ -542,38 +542,21 @@ func GenKnativeYaml(yamlTemplate string, deployPort int, serviceName string, dep
 		return "", err
```
kewegner: FUTURE: I think in future refactoring we might want to consider moving the k8s functions out of utils.go into a k8utils.go etc, like we are planning on doing for docker commands.

cmd/operator.go
```diff
@@ -0,0 +1,242 @@
...
+
+	"github.com/pkg/errors"
+
+	"github.com/spf13/cobra"
+)
+
+var operatorYamlName = "appsody-app-operator.yaml"
+var appsodyCRDName = "appsody-app-crd.yaml"
+var crdURL = "https://raw.githubusercontent.com/appsody/appsody-operator/master/deploy/releases/0.1.0/appsody-app-crd.yaml"
+var operatorURL = "https://raw.githubusercontent.com/appsody/appsody-operator/master/deploy/releases/0.1.0/appsody-app-operator.yaml"
```
chilanti: Are we going with these URLs? I thought you said we'd point to the latest release.

cmd/deploy.go
```diff
@@ -15,107 +15,395 @@
...
+		return errors.Errorf("Failed to write local application configuration file: %s", err)
+	}
+
+	Info.log("Created deployent manifest: ", configFile)
+	return nil
+}
+
 func init() {
 	rootCmd.AddCommand(deployCmd)
+	deployCmd.PersistentFlags().BoolVar(&generate, "generate", false, "Only generate the deployment configuration file. Do not deploy the project.")
```
chilanti: Would `--generate-only` or `--generate-manifest` or `--generate-yaml` be more meaningful?

cmd/deploy.go
seabaylea: Fixed.

cmd/deploy.go
seabaylea: Fixed.

cmd/deploy.go
seabaylea: The challenge is that the CRD spec for the app-deploy.yaml is currently v1alpha1 so it likely to change (whereas the use of the specific `applicationImage:` field is unlikely to).

Ultimately we'd like to import the YAML definition as a struct from the appsody-operator project and then parse it using that directly.

cmd/deploy.go
seabaylea: This is actually a copy/paste from the original deploy command - so presumably its correct ;-P

cmd/operator.go
seabaylea: These are currently correct and work - yes, we will need to switch this to the formal Git release when its present (we should ideally get that PR in before we cut a new release of the CLI).

cmd/deploy.go
seabaylea: Switched to use `--generate-only`

seabaylea:
@chilanti @kewegner Thanks for the review. I think I've addressed everything raised.

Can you take another look?

chilanti:
After talking to Chris - there is enough `dry-run` support (it's buried in the lower level functions such as `KubeGet` as it should be).

kewegner:
Merging we need to reenable the requirement for tests to pass before merging
so we can merge i

PR #216, opened 2019-08-22
Move delete cmd to subcmd under deploy

ianpartridge:
As discussed with @seabaylea @neeraj-laad and @tomleah - we felt it made more sense for `appsody delete` to be `appsody deploy delete`.  Otherwise it could be confusing as to what exactly is being deleted.  With `appsody deploy delete` it's clear that it's the deployment that's being deleted.

My first contribution - all feedback very welcome - hopefully this is correct.


ianpartridge:
Not that I'm aware of.

PR #222, opened 2019-08-23
Windows deploy fix

### LINKED ISSUE #220: appsody deploy on windows fails: "Error attempting to run copy command cp -rfL ..."
**Describe the bug**
appsody deploy via operator is working on mac but failing on windows with...
```
Error attempting to run copy command cp -rfL /config/app-deploy.yaml /tmp/config/app-deploy.yaml on image appsody/nodejs:0.2
```

**To Reproduce**
0. start docker desktop and its k8s
1. `mkdir nodejs`
2. `cd nodejs`
3. `appsody init nodejs`
4. `appsody operator install`
5. `appsody deploy`

**Expected behavior**
deploy should finish without errors

**Actual behaviour**
```
Error attempting to run copy command cp -rfL /config/app-deploy.yaml /tmp/config/app-deploy.yaml on image appsody/nodejs:0.2
```

 - OS: windows 10

 - CLI version: 0.3.0 plus fixes
 - Stack you are using: nodejs

**Additional context**
`appsody extract` works...
```
C:\Users\appsody\projects082219\nodejs2>appsody extract
Extracting project from development environment
Running command: docker[pull appsody/nodejs:0.2]
[Warning] The stack image does not contain APPSODY_PROJECT_DIR. Using /project
Running command: docker[run --name nodejs2-extract -v C:/Users/appsody/projects082219/nodejs2/:/project/user-app --entrypoint /bin/bash appsody/nodejs:0.2 -c cp -rfL /project /tmp/project]
Running command: docker[cp nodejs2-extract:\tmp\project C:\Users\appsody\.appsody\extract\nodejs2]
Running command: docker[rm nodejs2-extract -f]
```

the docker command fails...
```
C:\Users\appsody\projects082219\nodejs2>docker run --name nodejs-extract --entrypoint /bin/bash appsody/nodejs:0.2 -c cp -rfL /config/app-deploy.yaml /tmp/app-deploy.yaml
cp: missing file operand
Try 'cp --help' for more information.
```

if i manually create c:\tmp the docker command does not return an error but noting gets copied to c:\tmp

kewegner:
Fixes #220.

This is a critical fix, appsody deploy fails on windows without it.

The solution is to use the same cp logic for windows as the other os's.
Also I had to defer the remove of the temp file because it caused problems on windows.

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -21,8 +21,6 @@ import (
 	"io/ioutil"
 	"os"
 	"os/signal"
-	"path/filepath"
-	"runtime"
 	"strconv"
 	"strings"
 	"syscall"
@@ -114,8 +112,8 @@ generates a deployment manifest (yaml) file if one is not present, and uses it t
 		if err != nil {
 			return err
 		}
-		defer file.Close()
 
+		defer closeAndRemoveFile(file, tempConfigFile)
 		w := bufio.NewWriter(file)
 		for _, line := range txtlines {
 			fmt.Fprintln(w, line)
@@ -135,11 +133,6 @@ generates a deployment manifest (yaml) file if one is not present, and uses it t
 		}
 		Info.log("Deployment succeeded.")
 
-		err = os.Remove(tempConfigFile)
-		if err != nil {
-			return err
-		}
-
 		// Ensure hostname and IP config is set up for deployment
 		time.Sleep(1 * time.Second)
 
@@ -154,6 +147,14 @@ generates a deployment manifest (yaml) file if one is not present, and uses it t
 	},
 }
 
+func closeAndRemoveFile(file *os.File, fileName string) {
+	file.Close()
+	Debug.log("Removing file: ", fileName)
+	err := os.Remove(fileName)
+	if err != nil {
+		Warning.logf("Failed to remove %s.", fileName)
+	}
+}
 func deployWithKnative(cmd *cobra.Command, args []string) error {
 	buildErr := buildCmd.RunE(cmd, args)
 	if buildErr != nil {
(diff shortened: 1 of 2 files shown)


chilanti:
What has changed on Windows that can allow you to skip the navigation of the symlinks? The code that was removed had a reason to be there for some of the stacks, where symlinks were present, the `cp` command would fail.

kewegner:
I think the difference here, is it isn't trying to copy a tree of stuff, it is only trying to copy the single file /config/app-deploy.yaml.  Which was added for this new support.  I will say I only tried it with nodejs-express. We can try it with the other stacks.

The extract and build stuff is still handled by the build and extract logic. And still works.  Still undetermined why the same logic can't copy one file.  One difference is there no mounts.
I suspect it is a permissions problem of some sort but could not get it to work after several hrs of trying, but I had been manually trying things and got the solution in the PR to work, and then realized that that was what the non-windows path did.
I had tried adding a volume mount for /tmp and on the command line that worked, but running the same docker command inside the cli didn't work.

cmd/deploy.go
```diff
@@ -154,6 +147,14 @@ generates a deployment manifest (yaml) file if one is not present, and uses it t
...
 
+func closeAndRemoveFile(file *os.File, fileName string) {
+	file.Close()
+	Debug.log("Removing file: ", fileName)
+	err := os.Remove(fileName)
+	if err != nil {
+		Warning.logf("Failed to remove %s.", fileName)
+	}
+}
 func deployWithKnative(cmd *cobra.Command, args []string) error {
```
neeraj-laad: is this function for all deploys? or just for knative?

chilanti:
The windows dependent code was NOT redundant. There were problems with the Java stacks on Windows, because of symlinks that were present in the stack and that wouldn't copy on Windows. This needs to be tested thoroughly with liberty and spring.

cmd/deploy.go
kewegner: the close and remove function is only for the non knative path

kewegner: So the reason we could do away with it was there was only one file to copy and it was not a symlink.  In the case of build/extract they are trying to copy an entire stack in which there can be symlinks.
So for just app-deploy.yaml we can live without the windows specific code.
But for a whole tree of a stack image, we'd need it.  So it needs to remain in the build and deploy path.
Rather than fight the battle of why the Windows path didn't work (my guess permissions and the /tmp directory), I elected to remove it, once I realized we could live without it for this cp of app-deploy.yaml.

Issue #268, opened 2019-09-04
"Running command:" argument list surrounded by brackets


ykoyfman:
Remove the `[` and `]` surrounding the args in the CLI output that shows what commands are being run.  Users not familiar with kubectl CLI may not know to remove the brackets before attempting commands.

For example:
```
#  appsody deploy delete
Deleting deployment using deployment manifest app-deploy.yaml                                                                                                                                                   
Attempting to delete resource from Kubernetes...                                                                                                                                                                
Running command: kubectl[delete -f app-deploy.yaml --namespace default]                                                                                                                              
Deployment deleted
```

kewegner:
Unfortunately the fix corrupts the output and removes the colored output, so we will need another solution.

kylegc:
Reopend. Sandy, we cannot remove the brackets from every single log message. That code has been reverted in PR #293. We need to go through the calls to exec and only remove the brackets around commands before/while sending to the logger.

PR #273, opened 2019-09-05
cmd: removed brackets from argument list

skoh7645:
Removed the `[` and `]` surrounding the args in the CLI output that shows what commands are being run.

fixes: #268


kewegner:
The ports_test.go and stop_test.go explicitly look for the output docker[ so that will need to change.

Issue #250, opened 2019-08-28
Update repo from V1 index to use the V2 index


mhamwala:
I recently added some documentation to the FAQ section of the website addressing part of this issue. 

Let's use this issue to discuss whether or not we want to implement a way to automatically upgrade users who are using the old V1 index to use the V2 index.

Additionally how and when we will delete the V1 index

related: #223

ebullient:
@gastaldi Should be good that `f8 launcher catalog` adopt this index v2 spec to unify our efforts and been able in the future to propose perhaps different repos but adhering to the same spec that `appsody repo list` could use and next `appsody init` to scaffold a project. WDYT ?

_Originally posted by @cmoulliard in https://github.com/appsody/appsody/issues/223#issuecomment-528445971_

gastaldi:
@ebullient @cmoulliard yes, that sounds like a good idea

cmoulliard:
> a good idea

Where is the index v2 spec defined @ebullient ?

ebullient:
I actually don't think we're writing it as a spec anywhere, @neeraj-laad ? 

v2 is what is now produced by appsody stacks build tooling: 
https://github.com/appsody/stacks/releases/download/python-flask-v0.1.3/incubator-index.yaml

v1 is the format of the manually maintained index.yaml in appsody/stacks: https://github.com/appsody/stacks/blob/82dca1afd73f6bc901b4bbf6bbe725075fbfb26b/index.yaml

PR #276, opened 2019-09-05
new index: switching index from v1 to v2

mhamwala:
Reads the users local appsody `repository.yaml` and switches the v1 index to use the v2 index.

fixes: https://github.com/appsody/appsody/issues/250

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -1015,3 +1015,22 @@ func checkTime() {
 		}
 	}
 }
+
+func setNewIndexURL() {
+
+	var repoFile = getRepoFileLocation()
+	var oldIndexUrl = "https://raw.githubusercontent.com/appsody/stacks/master/index.yaml"
+	var newIndexUrl = "https://github.com/appsody/stacks/releases/latest/download/incubator-index.yaml"
+
+	data, err := ioutil.ReadFile(repoFile)
+	if err != nil {
+		Warning.log("Unable to read repository file")
+	}
+
+	replaceUrl := bytes.Replace(data, []byte(oldIndexUrl), []byte(newIndexUrl), -1)
+
+	if err = ioutil.WriteFile(repoFile, replaceUrl, 0666); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+}
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -1016,6 +1016,8 @@ func checkTime() {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


cmd/utils.go
```diff
@@ -1045,3 +1045,22 @@ func checkTime() {
...
+
+	data, err := ioutil.ReadFile(repoFile)
+	if err != nil {
+		Warning.log("Unable to read repository file")
+	}
+
+	replaceURL := bytes.Replace(data, []byte(oldIndexURL), []byte(newIndexURL), -1)
+
+	if err = ioutil.WriteFile(repoFile, replaceURL, 0666); err != nil {
+		fmt.Println(err)
```
kylegc: Lets turn this into `Warning.log(err)`

cmd/utils.go
```diff
@@ -1045,3 +1045,22 @@ func checkTime() {
...
+		os.Exit(1)
```
kylegc: I think we should remove the os.Exit
If this fails, the user may still be able to use appsody with the old index.

cmd/utils.go
```diff
@@ -1045,3 +1045,22 @@ func checkTime() {
 		}
 	}
 }
+
+func setNewIndexURL() {
```
kylegc: Please add some comments to this function to describe its purpose. Especially in this case, I'm looking for something that says this is temporary code and should be removed in the future.

cmd/utils.go
```diff
@@ -1045,3 +1045,22 @@ func checkTime() {
...
+	var newIndexURL = "https://github.com/appsody/stacks/releases/latest/download/incubator-index.yaml"
+
+	data, err := ioutil.ReadFile(repoFile)
+	if err != nil {
+		Warning.log("Unable to read repository file")
+	}
+
+	replaceURL := bytes.Replace(data, []byte(oldIndexURL), []byte(newIndexURL), -1)
+
+	if err = ioutil.WriteFile(repoFile, replaceURL, 0666); err != nil {
```
kylegc: Also, I suggest using 0644 to be consistent with the way we write the repo file today.
See https://github.com/appsody/appsody/blob/master/cmd/repo.go#L476

cmd/utils.go
mhamwala: yep this makes sense!

cmd/utils.go
mhamwala: change addressed

mhamwala:
Changes addressed :)

PR #283, opened 2019-09-07
buildah: use bind mount and localy copy

gireeshpunathil:
1. 

When stacks whose Dockerfile defines less privileged users
as the owner of the container, extraction fails in CRI-O
deployments because the mount point is owned by privileged
user who started the parent container, and the less privileged
user does not have write permission into it.
    
So we go by the old bind mount route where:
  - don't run the child container, just get its mount point,
  - copy the project folder from the mount to the extract folder.

2. 

A class of systems (e.g:- RHEL 7.6) exhibit situations wherein
the bindmount volumes are not propagated to the child containers
Accommodate those systems as well, by performing local copies
for the locations that are resident in the host.
ref: https://github.com/containers/buildah/issues/1821

### DIFF
--- a/cmd/extract.go
+++ b/cmd/extract.go
@@ -17,7 +17,7 @@ package cmd
 import (
 	"fmt"
 	"os"
-	s "strings"
+	"strings"
 
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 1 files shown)


gireeshpunathil:
cc @chilanti @ianpartridge @neeraj-ladd @kilnerm

cmd/extract.go
```diff
@@ -17,6 +17,7 @@ package cmd
 import (
 	"fmt"
 	"os"
+	s "strings"
```
neeraj-laad: I would prefer not to use `s` and instead use `strings` to make the code clearer.

neeraj-laad:
Res of the PR looks ok to me.. as it mainly only impacts buildah code. it should not break other Appsody features.

gireeshpunathil:
I just pushed a change to address the review comments; thanks!

Issue #102, opened 2019-07-24
Appsody does not pass the `-v` flag on to the controller


kylegc:
When appsody is run with the `-v` flag for verbose debugging, it should also pass this flag on to the controller for its verbose debugging as well.

PR #285, opened 2019-09-09
run: propagate verbose flag to the controller

gireeshpunathil:
Fixes: https://github.com/appsody/appsody/issues/102

cc @kylegc  @kewegner @neeraj-laad

### DIFF
--- a/cmd/dev_common.go
+++ b/cmd/dev_common.go
@@ -259,6 +259,9 @@ func commonCmd(cmd *cobra.Command, args []string, mode string) error {
 		cmdArgs = append(cmdArgs, dockerOptionsCmd...)
 	}
 	cmdArgs = append(cmdArgs, "-t", "--entrypoint", "/appsody/appsody-controller", platformDefinition, "--mode="+mode)
+	if verbose {
+		cmdArgs = append(cmdArgs, "-v")
+	}
 	Debug.logf("Attempting to start image %s with container name %s", platformDefinition, containerName)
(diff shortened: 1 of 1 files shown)


kewegner:
The controller outputs a lot of verbose debug.  We may want to revisit the logging the controller does.  That was the reason I held off on doing this enhancement.

gireeshpunathil:
here is a sample output from `nodejs` stack with this flag propagation:
```
[Container] 33m39m
[Container] I0909 09:36:47.553463       1 main.go:351   Debug  Install complete
[Container] I0909 09:36:47.553523       1 main.go:361  Running: npm start --node-options --require=appmetrics-dash/attach
[Container] I0909 09:36:47.553543       1 main.go:363   Debug  Set workdir:  /project/user-app
[Container] I0909 09:36:47.555035       1 main.go:375   Debug  New process created with pid 48
[Container] I0909 09:36:47.555096       1 main.go:520   Debug  Started Server process
[Container] I0909 09:36:47.555112       1 main.go:526   Debug  mutex unlocked
[Container] I0909 09:36:47.555124       1 main.go:381   Debug  waitProcess ENTRY
[Container] 
[Container] > nodejs-simple@0.1.0 start /project/user-app
[Container] > node app.js
```
while logging verbosity within the container may be a subject of revision, I don't think that will affect this switch in any ways. Moreover, this is output only when verbose is turned on in the `appsody run` .

kylegc:
I agree and have merged this. We can address the output of the controller in a different issue if desired. 

Thanks Gireesh!

Issue #292, opened 2019-09-10
0.4.3 escapes control characters in output


ebullient:
**Describe the bug**

In _0.4.2_ of the CLI, command output (e.g. from maven) showed colors added by build and other tools.. which is especially nice for test output, etc.

![image](https://user-images.githubusercontent.com/808713/64634867-3b1c2480-d3cc-11e9-8c2c-c629b1bd31f6.png)

0.4.3 has escaped or scrubbed the output (possibly by routing it through a logger), displaying escaped (read: ugly) control characters.

![image](https://user-images.githubusercontent.com/808713/64634810-19bb3880-d3cc-11e9-903b-d5fa8858a031.png)

**To Reproduce**

Perform any appsody operation like `run` or `test` with a 0.4.2 (or earlier) client.
Compare with 0.4.3

The java-spring-boot2 stack gives a reasonable demonstration of the difference.

Working with commands (like `test` or `run`) should be as transparent as possible, and should retain interactive niceties provided by the stack tooling in the most native/natural way possible.

An aside:  I don't see what value a `[Container]` prefix provides, and it just adds visual clutter. Why do I care that the container is building it? I just want to know that maven/node/whatever is running, and I can see that from the output.

kylegc:
Fixed by #293

PR #293, opened 2019-09-10
temporary fix to restore [] in output

kewegner:
Remove the code (PR #273 ) which removed "[]" from the output.
That issue will be dealt with in another issue/pull request.


kylegc:
Fixes #292

Issue #290, opened 2019-09-10
Option to disable watcher for tests


ebullient:
**Is your feature request related to a problem? Please describe.**

Stacks are inconsistent in the behavior of `appsody test` (some run through and quit, others run and then watch). For a script or other automation, you want to be able to run the tests and quit with the appropriate result based on failures. There isn’t anyway to specify this behavior via the appsody command line at the moment. 

Doing any kind of manipulation on test output to determine when tests are done is language/stack specific and fragile. Trying to force all stacks to behave identically with `appsody test` is also not useful, as we expect further customization of stacks...

**Describe the solution you'd like**

`appsody test --no-watcher` could work. 

I want the stack specified tests to run, but for the process to terminate and return with the result of the executed command.
 
Related to appsody/stacks#308 —> appsody/stacks#323

PR #297, opened 2019-09-10
added --no-watcher flag to disable file watching regardless of Stack Environment variable settings.

kewegner:
This PR adds a --no-watcher flag to disable file watching by the controller regardless of the Docker Environment variables in the stack.

In order for this PR to function correctly, a new release of the controller will be needed.
That function is part of Issue https://github.com/appsody/controller/issues/21 of the appsody controller component and its corresponding pull request.


ebullient:
Fixes #290

PR #302, opened 2019-09-11
cmd: removed [] from argument list

skoh7645:
Removed the "[" and "]" surrounding the args in the CLI output only in cases where output contains "docker command:" or "Running command:"

Fixes: #268


skoh7645:
@kewegner Yes, I understand how changing it globally could affect things in scenarios we do not realise. I was trying to avoid doing it individually as it seemed inefficient. I have addressed the changes you wanted in the most recent commit

kewegner:
I reviewed the next round of changes it looks much better. Requesting Kyle for 2nd review.

Issue #305, opened 2019-09-11
Appsody list: handle HTTP errors nicely


chilanti:
**Is your feature request related to a problem? Please describe.**
If your repo list includes repositories that become inaccessible for some reasons, `appsody list` fails even though other repositories in the list are still accessible.

**Describe the solution you'd like**
`appsody list` (and `appsody init`) should keep working if some repos are "dead". Also - the timeout could be trimmed to less than the 120 secs (to be discussed).

PR #308, opened 2019-09-12
cmd: handle HTTP errors nicely

skoh7645:
`appsody list` (and `appsody init`) will keep working even if some repos are inaccessible.

Fixes: #305

### DIFF
--- a/cmd/repo.go
+++ b/cmd/repo.go
@@ -476,20 +476,41 @@ func (r *RepositoryFile) WriteFile(path string) error {
 	return ioutil.WriteFile(path, data, 0644)
 }
 
+type indexError struct {
+	indexName string
+	theError error
+ }
+ 
+ func (e indexError) Error() string {
+	return e.indexName + " " + fmt.Sprintf("%s", e.theError)
+ }
+ 
+ type indexErrors struct {
+	listOfErrors []indexError
+ }
+ 
+ func (e indexErrors) Error() []string {
+	var result []string
+	for _, err := range e.listOfErrors {
+		result = append(result, fmt.Sprintf("%s", err))
+	}
+	return result
+ }
+
 func (r *RepositoryFile) GetIndices() (RepoIndices, []string, error) {
-	
-	names := []string{}
+	names := make([]string, 0)
 	indices := make(map[string]*RepoIndex)
 	for _, rf := range r.Repositories {
 		var index, err = downloadIndex(rf.URL)
 		if err != nil {
-			names := append(names, rf.Name)
-			return indices, names, err
+			names = append(names, rf.Name)
+		} else {
+			indices[rf.Name] = index
 		}
-		indices[rf.Name] = index
 	}
-	return indices, nil, nil
+	return indices, names, nil
 }
+
 func (index *RepoIndex) buildStacksFromIndex(repoName string, Stacks []Stack) ([]Stack, error) {
 
 	for id, value := range index.Projects {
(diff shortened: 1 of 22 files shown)


cmd/repo.go
```diff
@@ -476,17 +476,19 @@ func (r *RepositoryFile) WriteFile(path string) error {
...
+func (r *RepositoryFile) GetIndices() (RepoIndices, []string, error) {
+	
+	names := []string{}
 	indices := make(map[string]*RepoIndex)
 	for _, rf := range r.Repositories {
 		var index, err = downloadIndex(rf.URL)
 		if err != nil {
-			return indices, err
+			names := append(names, rf.Name)
+			return indices, names, err
```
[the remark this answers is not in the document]
skoh7645: No, that is correct. I wasn't sure how to keep the loop going as we have to return the `err` message as well.

cmd/repo.go
```diff
@@ -477,17 +498,20 @@ func (r *RepositoryFile) WriteFile(path string) error {
...
-			return indices, err
+			repoErr := indexError{rf.Name, err}
+			brokenRepos = append(brokenRepos, repoErr)
+		} else {
+			indices[rf.Name] = index
 		}
-		indices[rf.Name] = index
 	}
-	return indices, nil
+	return indices, &indexErrors{brokenRepos}
```
chilanti: I think you should return  `&indexErrors{brokenRepos}` only if `len(brokenRepos) > 0`, otherwise the caller will think there's always an error. Let me know if I am mistaken.

kylegc: I agree, if there are no broken repos, you should return a `nil` error

cmd/repo.go
```diff
@@ -101,6 +101,27 @@ func findTemplateURL(projectVersion ProjectVersion, templateName string) string
...
+	indexName string
+	theError  error
+}
+
+type indexErrors struct {
+	listOfErrors []indexError
+}
+
+func (e indexError) Error() string {
+	return "- Repository: " + e.indexName + "\n  Reason: " + fmt.Sprintf("%s", e.theError)
```
kylegc: The `fmt.Sprintf()` is unnecessary here, you should just use `e.theError.Error()`

cmd/repo.go
```diff
@@ -101,6 +101,27 @@ func findTemplateURL(projectVersion ProjectVersion, templateName string) string
...
+}
+
+func (e indexErrors) Error() string {
+	var myerrors string
```
kylegc: Can you use the built in join method here? Something like:
`return strings.Join(e.listOfErrors, "\n")`

cmd/repo.go
```diff
@@ -476,17 +476,19 @@ func (r *RepositoryFile) WriteFile(path string) error {
 	return ioutil.WriteFile(path, data, 0644)
 }
 
-func (r *RepositoryFile) GetIndices() (RepoIndices, error) {
-
+func (r *RepositoryFile) GetIndices() (RepoIndices, []string, error) {
```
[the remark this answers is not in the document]
skoh7645: Thank you! This was very helpful

cmd/repo.go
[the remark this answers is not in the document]
skoh7645: New commits have addressed this issue

cmd/repo.go
skoh7645: Done

cmd/repo.go
skoh7645: Done

cmd/repo.go
skoh7645: I tried this and I cannot use `e.listOfErrors` with the strings join method as it is not of type `[]string`

kylegc: Of course, nevermind :-)

Issue #270, opened 2019-08-30
Validate Dockerfile-stack in Linter


Kamran64:
**Is your feature request related to a problem? Please describe.**
Currently the stacks linter will only verify that a Dockerfile-stack exists, but won't actually validate the contents of the file. As a result the lint test will pass even if the file is empty. 

The linter could validate the use of environment variables within the Dockerfile-stack and ensure e.g. directories set for WATCH_DIR exist.

If we could post rules that a Dockerfile-stack must conform to either here or in the [linter epic](https://github.com/appsody/stacks/issues/199)

Bazif-Khan:
These environment variables should be present:

**Error:**
- [ ] APPSODY_MOUNTS
- [ ] APPSODY_RUN
- [ ] APPSODY_RUN_ON_CHANGE
- [ ] APPSODY_RUN_KILL
- [ ] APPSODY_DEBUG
- [ ] APPSODY_DEBUG_ON_CHANGE
- [ ] APPSODY_DEBUG_KILL
- [ ] APPSODY_TEST
- [ ] APPSODY_TEST_ON_CHANGE
- [ ] APPSODY_TEST_KILL

These environment variables are optional:

**Warning:**
- [ ] APPSODY_DEPS
- [ ] APPSODY_WATCH_DIR

neeraj-laad:
@Bazif-Khan I believe we need some changes on this:

**What vars are allowed and when:**
- *ON_CHANGE: should be optional
- *_KILL: should be optional
- IF WATCH_DIR is preset we expect to have at least one ON_CHANGE
- IF WATCH_DIR is not preset we expect to not have any ON_CHANGE
- APPSODY_INSTALL should be deprecated and APPSODY_PREP should be used instead.

What values are allowed and when:
* *KILL should be trtue/false
* APPSODY_WATCH_REGEX - check value is a valid regex
* Validation on paths?
* APPSODY_PROJECT_DIR should be used?

kewegner:
Re: IF WATCH_DIR is present we expect to have at least one ON_CHANGE
Currently if there is no WATCH_DIR we default to the mount dirs if there is an ON_CHANGE setting.
I think what we should do is print an info message that if WATCH_DIR is not present and an ON_CHANGE is that we default to the MOUNTS_DIR

Re the regex. 

Given that true validation is kind of dependent on what is the source being watched, (e.g. what text we run the regex on..) I am not sure how deep a validation one could do.

[...]

PR #311, opened 2019-09-12
Linter: Validate Dockerfile-stack

Bazif-Khan:
fixes: https://github.com/appsody/appsody/issues/270

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,8 +57,7 @@ jobs:
       deploy:
         provider: releases
         skip_cleanup: true
-        api_key:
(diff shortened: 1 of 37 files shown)


Bazif-Khan:
> So this does provide some nice basic coverage. One unfortunate thing it will miss is if you do not have APPSODY_RUN, but do have, say, APPSODY_RUN_ON_CHANGE then it won't spot that (since APPSODY_RUN matches both). I guess you could look for "APPSODY_RUN=" or "APPSODY_RUN " (note the space) and if neither of those where present, then it's an error?

Thanks. I have changed to "APPSODY_RUN=".

henrynash:
Is the same not true for APPSODY_DEBUG and APPSODY_TEST? In fact, do be consisted why don't you test all with an "=" after them? However, a word of caution, is not

ENV APPSODY_RUN blablabla

also valid? (I.e. no '=' but white space instead) - it appears to be from docker documentation? If that is valid (you should check), that is why I suggested you need to test for the existence of exactly one of 'APPSODY_RUN=' or 'APPSODY_RUN '

PR #312, opened 2019-09-12
Handle --knative option for current stacks as well as stacks that do not contain app-deploy.yaml

kewegner:
If the --knative flag is present and there is a app-deploy.yaml containing an appsodyapplication, the createKnativeService: true tag will be added to the yaml file.

If the --knative flag is not present and createKnativeService: true is in the app-deploy.yaml file the createKnativeService element will be removed.  if it is createKnativeService: false, it will remain.

If the stack does not contain an app-deploy.yaml file containing and appsodyapplication, the --knative flag will have no impact.

Unless the appsody deploy is done with a --tag, -t option the imagename will be prefixed with dev.local/<projectname>

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -127,6 +127,12 @@ generates a deployment manifest (yaml) file if one is not present, and uses it t
 				foundCreateKnativeTag = true
 				localFoundCreateKnativeTag = true
 			}
+			if strings.Contains(line, "createKnativeService") && strings.Contains(line, "false") && knative {
+				//if knative we want to replace false with true in
+				line = strings.Replace(line, "false", "true", 1)
+				foundCreateKnativeTag = true
+				localFoundCreateKnativeTag = true
+			}
 			if strings.Contains(line, "kind:") && strings.Contains(line, "Service") {
 				isKnativeService = true
 			}
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -31,7 +31,7 @@ import (
 )
 
 var configFile, namespace, watchspace, tag string
-var generate, force, push bool
+var knative, generate, force, push bool
 
 var deployCmd = &cobra.Command{
 	Use:   "deploy",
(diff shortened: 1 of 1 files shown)


cmd/deploy.go
```diff
@@ -78,21 +78,27 @@ generates a deployment manifest (yaml) file if one is not present, and uses it t
...
 
-		//Retrieve the project name and lowercase it
-		projectName, perr := getProjectName()
-		if perr != nil {
-			return errors.Errorf("%v", perr)
-		}
-		deployImage := projectName // if not tagged, this is the deploy image name
+		var deployImage string
+		//	:= projectName // if not tagged, this is the deploy image name
 		if tag != "" {
```
chilanti: Is this check still needed? We check tag earlier and set it if not set....

cmd/deploy.go
```diff
@@ -103,26 +109,51 @@ generates a deployment manifest (yaml) file if one is not present, and uses it t
...
 		for scanner.Scan() {
+			var localFoundCreateKnativeTag bool
 			line := scanner.Text()
 			if strings.Contains(line, "applicationImage:") {
 				index := strings.Index(line, ": ")
 				start := line[0:(index + 2)]
 				line = start + deployImage
 				Info.log("Using applicationImage of: ", deployImage)
 			}
-			txtlines = append(txtlines, line)
```
chilanti: I would suggest the following:
```
for ... {
if strings.Contains(line, "createKnativeService") { //it's the infamous flag
 foundCreateKnativeTag = true
 if strings.Contains(line, "true") && !knative {
  //replace with false 
 } else if strings.Contains(line, "false") && knative {
  //replace with "true"
 }
} 
append()
} //end loop
//check if !found and append in that case

kewegner: I think that is good, one alternative might be to just check if the tag is there and then regen using the value of knative.  But the above is OK and probably simpler.

I think the bigger problem I just thought of is the append if createKnativeService is not there... Do we know that someone has not hand edited the file and added another k8s object defintiion?  I think we need to ask @neeraj-laad what the limits of editing app-deploy.yaml are.

cmd/deploy.go
chilanti: Understood you are setting a different tag.

cmd/deploy.go
kewegner: Changes have been made.  We will discuss the editing of app-deploy.yaml on the scrum and see if more work is needed.

PR #314, opened 2019-09-13
test: add a test for `appsody extract`

gireeshpunathil:
Extraction test logic:
~~Don't inspect the images, but just figure out the mount points from the debug log that is thrown out from extraction~~
 - Get the environment variables from the images that was just extracted - GetEnvVar. Switch the current directory to meet the API's needs. (refactor it so that this isn't needed)
 - For each mount that is a file, make sure the source and destination file sizes match.
 - For each mount that is a folder, make sure the source and destination folders have same content (file name match)
 - Skip mounts that were not extracted (e.g:- /mvn/...)
 - For all cases, make sure we have a `Dockerfile` extracted at the vortex of the extraction.

### DIFF
--- a/functest/extract_test.go
+++ b/functest/extract_test.go
@@ -1,0 +1,198 @@
+// Copyright © 2019 IBM Corporation and others.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package functest
+
+import (
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+	"reflect"
+	"regexp"
+	"strings"
+	"testing"
+
+	cmdtest "github.com/appsody/appsody/cmd/cmdtest"
+)
+
+// Simple test for appsody extract command.
+// A future enhancement is to extract with buildah.
+// I could have very well taken one single stack
+// and test, but due to the fact that most pain points
+// come from diverse mounts that exist for different
+// stacks, it is nice to loop through each and make sure
+// nothing is broken.
+
+func TestExtract(t *testing.T) {
+
+	log.Println("stacksList is: ", stacksList)
+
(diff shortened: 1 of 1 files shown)


functest/extract_test.go
```diff
@@ -0,0 +1,198 @@
...
+		// 1. Don't inspect the images, but just figure out the mount points
+		// from the debug log that is thrown out from extraction
+		// 2. For each mount that is a file, make sure the source and
+		// destination file sizes match.
+		// 3. For each mount that is a folder, make sure the source and
+		// destination folders have same content (file name match)
+		// 4. Skip mounts that were not extracted (e.g:- /mvn/...)
+		// 5. For all cases, make sure we have a Dockerfile extracted
+		// at the vortex of the extraction.
+		r, _ := regexp.Compile("APPSODY_MOUNTS Value: .*")
```
chilanti: What's the rationale behind using the debug output to find out what the mounts are? We like the test, but at the same time we think it's risky to depend on a debug message to figure out the mounts.

gireeshpunathil: @chilanti - I was concerned about this too! 

my rationale: if we want to obtain the environment of the stack, I would need to either do almost everything as `extract` do, or invoke APIs that take part in extraction logic. In ether way, any potential bug that affects `extract` can also infect into the test.

I changed it now, and took a new approach - basically run the `extract` in-process, temporarily switch the working folder the project location, get the environment through `GetEnvVar` - right now this is not exported, but a change is proposed to that effect through https://github.com/appsody/appsody/pull/322 , pls have a look!

gireeshpunathil:
Don't land this yes - it depends on changes in https://github.com/appsody/appsody/pull/322

functest/extract_test.go
chilanti: Thanks - will take a look - and I do see your point (it doesn't make sense to test `extract` with some of the same `extract` logic). But we'll take it.

functest/extract_test.go
```diff
@@ -0,0 +1,208 @@
...
+		dockerFile := filepath.Join(extractDir, "Dockerfile")
+		_, err = exists(dockerFile)
+		if err != nil {
+			t.Fatal("Extraction failure, Dockerfile was not extracted into ", extractDir)
+		}
+	}
+}
+
+// private exists fn.
+// TODO: Can this be re-used from cmd/utils.go?
```
chilanti: Yes - we should reuse the one in utils. Please make it public.

gireeshpunathil: @chilanti - raised in #328 , PTAL, thanks!

gireeshpunathil:
[ PR now blocked on #328 ]

Issue #313, opened 2019-09-12
Too many calls to check latest version URL


kylegc:
**Describe the bug**
The latest version URL https://github.com/appsody/appsody/releases/latest
Should only be called once every 24 hours.
I see it is being called on every CLI run.

Also, after it has been 24 hours, the URL gets called twice back-to-back in the same CLI run.

**To Reproduce**
Put a debug statement `Debug.log("Getting latest version ", LatestVersionURL)` in utils.go getLatestVersion() method.
Run the CLI with `-v`

**Expected behavior**
Only one call to LatestVersionURL the first time. Then no calls to LatestVersionURL for 24 hours.

 - CLI version: 0.4.4

PR #315, opened 2019-09-13
cmd: limited calls to check latest version available

skoh7645:
Fixes: #313

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -1008,19 +1008,16 @@ func doVersionCheck() {
 	if latest != "" && VERSION != "vlatest" && VERSION != latest {
 		switch os := runtime.GOOS; os {
 		case "darwin":
-			Info.logf("*\n*\n*\n\nA new CLI update is available.\nPlease run `brew upgrade appsody` to upgrade from %s --> %s.\n\n*\n*\n*", VERSION, latest)
-		case "linux":
-			Info.logf("*\n*\n*\n\nA new CLI update is available.\n1)Please download the most recently released Debian install package from the Appsody releases page.\n2)Open your terminal and enter the directory in which you downloaded the file.\n3)Run the command `sudo apt upgrade -f ./appsody_v.r.m_amd64.deb -y` to upgrade from %s --> %s.\n\n*\n*\n*", VERSION, latest)
-		case "windows":
-			Info.logf("*\n*\n*\n\nA new CLI update is available.\n1)Please download the Appsody binaries for Windows from the Appsody releases page.\n2)Move the file to the directory where you stored the existing Appsody binaries.\n3)Run the command `tar -xvf appsody-v.r.m-windows.tar.gz` to upgrade from %s --> %s.\n\n*\n*\n*", VERSION, latest)
+			Info.logf("\n*\n*\n*\n\nA new CLI update is available.\nPlease run `brew upgrade appsody` to upgrade from %s --> %s.\n\n*\n*\n*", VERSION, latest)
 		default:
-			Info.logf("*\n*\n*\n\nA new CLI update is available.\nPlease go to https://appsody.dev/docs/getting-started/installation and upgrade from %s --> %s.\n\n*\n*\n*", VERSION, latest)
+			Info.logf("\n*\n*\n*\n\nA new CLI update is available.\nPlease go to https://appsody.dev/docs/getting-started/installation and upgrade from %s --> %s.\n\n*\n*\n*", VERSION, latest)
 		}
-		cliConfig.Set("lastversioncheck", currentTime)
-		if err := cliConfig.WriteConfig(); err != nil {
-			Error.logf("Writing default config file %s", err)
+	}
 
-		}
+	cliConfig.Set("lastversioncheck", currentTime)
+	if err := cliConfig.WriteConfig(); err != nil {
+		Error.logf("Writing default config file %s", err)
+
 	}
 }
 
@@ -1031,18 +1028,14 @@ func getLastCheckTime() string {
 func checkTime() {
 	var lastCheckTime = getLastCheckTime()
 
-	if lastCheckTime == "none" {
+	lastTime, err := time.Parse("2006-01-02 15:04:05 -0700 MST", lastCheckTime)
+	if err != nil {
+		Debug.logf("Could not parse the config file's lastversioncheck: %v. Continuing with a new version check...", err)
 		doVersionCheck()
-	} else {
-		lastTime, err := time.Parse("2006-01-02 15:04:05 -0700 MST", lastCheckTime)
-		if err != nil {
-			Warning.log("Error", err)
-			doVersionCheck()
-		}
-		if time.Since(lastTime).Hours() > 24 {
-			doVersionCheck()
-		}
+	} else if time.Since(lastTime).Hours() > 24 {
+		doVersionCheck()
 	}
+
 }
 
 // TEMPORARY CODE: sets the old v1 index to point to the new v2 index (latest)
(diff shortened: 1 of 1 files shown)


skoh7645:
If the user is not connected to Internet, the timestamp will not be updated and they will get the "```Unable to check the most recent version of Appsody in GitHub.... continuing.```" message with every command. Do we want this, or should I change it to update the timestamp anyway (even though the version check hasn't actually been done)?

skoh7645:
I spoke to @neeraj-laad and we agreed to change the yaml variable to "lastCheckAttempt" and we are going to update the timestamp in the case we are not able to actually do the check to avoid sending the same message repeatedly

cmd/utils.go
```diff
@@ -1003,14 +1003,15 @@ func getLatestVersion() string {
 
 func doVersionCheck(data []byte, old string, new string, file string) {
 	var latest = getLatestVersion()
-	if VERSION != "vlatest" && VERSION != latest {
+	if latest != "none" && VERSION != "vlatest" && VERSION != latest {
 		Info.log("*\n*\n*\n\nA new CLI update is available.\nPlease go to " + LatestVersionURL + " and update from " + VERSION + " --> " + latest + ".\n\n*\n*\n*")
```
kylegc: I wonder if we should link to https://appsody.dev/docs/getting-started/installation

Actually it would be really nice to detect the client's OS and give specific instructions. For example, on mac we could say "Please run `brew upgrade appsody`."

cmd/utils.go
```diff
@@ -1003,14 +1003,15 @@ func getLatestVersion() string {
...
 	}
 	output := bytes.Replace(data, []byte(old), []byte(new), -1)
```
kylegc: It would probably be better to use cliConfig set and write methods here.

cmd/utils.go
```diff
@@ -1031,17 +1032,15 @@ func checkTime() {
...
 	currentTime = time.Now().Format("2006-01-02 15:04:05 -0700 MST")
 
-	if getLatestVersion() != "none" {
-		if lastCheckTime == "none" {
+	if lastCheckTime == "none" {
+		doVersionCheck(data, lastCheckTime, currentTime, configFile)
+	} else {
+		lastTime, err := time.Parse("2006-01-02 15:04:05 -0700 MST", lastCheckTime)
+		if err != nil {
+			fmt.Println(err)
```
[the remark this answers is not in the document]
skoh7645: Done

cmd/utils.go
```diff
@@ -1031,17 +1032,15 @@ func checkTime() {
 	lastCheckTime = getLastCheckTime()
 	currentTime = time.Now().Format("2006-01-02 15:04:05 -0700 MST")
 
-	if getLatestVersion() != "none" {
-		if lastCheckTime == "none" {
+	if lastCheckTime == "none" {
+		doVersionCheck(data, lastCheckTime, currentTime, configFile)
```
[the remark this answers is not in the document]
skoh7645: Done

cmd/utils.go
```diff
@@ -1031,17 +1032,15 @@ func checkTime() {
 	lastCheckTime = getLastCheckTime()
 	currentTime = time.Now().Format("2006-01-02 15:04:05 -0700 MST")
 
-	if getLatestVersion() != "none" {
-		if lastCheckTime == "none" {
+	if lastCheckTime == "none" {
```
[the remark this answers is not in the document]
skoh7645: In `root.go`, I set the default value to be "none"

cmd/utils.go
skoh7645: Done

cmd/utils.go
skoh7645: Done, although I am unsure how specific we want to be with our instructions

cmd/utils.go
```diff
@@ -991,56 +991,56 @@ func getLatestVersion() string {
...
+		case "darwin":
+			Info.logf("*\n*\n*\n\nA new CLI update is available.\nPlease run `brew upgrade appsody` to upgrade from %s --> %s.\n\n*\n*\n*", VERSION, latest)
+		case "linux":
+			Info.logf("*\n*\n*\n\nA new CLI update is available.\n1)Please download the most recently released Debian install package from the Appsody releases page.\n2)Open your terminal and enter the directory in which you downloaded the file.\n3)Run the command `sudo apt upgrade -f ./appsody_v.r.m_amd64.deb -y` to upgrade from %s --> %s.\n\n*\n*\n*", VERSION, latest)
+		case "windows":
+			Info.logf("*\n*\n*\n\nA new CLI update is available.\n1)Please download the Appsody binaries for Windows from the Appsody releases page.\n2)Move the file to the directory where you stored the existing Appsody binaries.\n3)Run the command `tar -xvf appsody-v.r.m-windows.tar.gz` to upgrade from %s --> %s.\n\n*\n*\n*", VERSION, latest)
+		default:
+			Info.logf("*\n*\n*\n\nA new CLI update is available.\nPlease go to https://appsody.dev/docs/getting-started/installation and upgrade from %s --> %s.\n\n*\n*\n*", VERSION, latest)
+		}
+		cliConfig.Set("lastversioncheck", currentTime)
```
kylegc: I thought we decided to always write the current time when we do a version check, not just when the version differed.

cmd/utils.go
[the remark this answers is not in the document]
kylegc: OK that is fine. I wonder though, do we even need this check? Or should we just let the time.Parse fail and run doVersionCheck() there?

cmd/utils.go
```diff
@@ -991,56 +991,56 @@ func getLatestVersion() string {
...
+	var lastCheckTime = getLastCheckTime()
 
 	if lastCheckTime == "none" {
-		doVersionCheck(data, lastCheckTime, currentTime, configFile)
+		doVersionCheck()
 	} else {
 		lastTime, err := time.Parse("2006-01-02 15:04:05 -0700 MST", lastCheckTime)
 		if err != nil {
-			fmt.Println(err)
+			Warning.log("Error", err)
```
kylegc: This may be better as a Debug log without the "Error" string. Something like:
Debug.log("Could not parse the config file's lastversioncheck. Continuing with a new version check.", err)

cmd/utils.go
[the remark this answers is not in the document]
kylegc: Looks much better!

cmd/utils.go
```diff
@@ -991,56 +991,56 @@ func getLatestVersion() string {
...
-	output := bytes.Replace(data, []byte(old), []byte(new), -1)
-	err := ioutil.WriteFile(file, output, 0666)
-	if err != nil {
-		Warning.log("Error writing to config file")
-	}
+	var currentTime = time.Now().Format("2006-01-02 15:04:05 -0700 MST")
+	configFile = getDefaultConfigFile()
+
+	if latest != "" && VERSION != "vlatest" && VERSION != latest {
+		switch os := runtime.GOOS; os {
```
kylegc: Thanks for writing this, I think it is really cool, however in talking to Michele, the longer instructions for linux and windows would be better to point to the website doc instead. Mainly because there are some additional notes in the docs and they may change in the future.
Also, there are two variations of linux (debian and rhel) with different instructions that you cannot detect here.

Net: keep the case for "darwin", but remove linux and windows to result in the default for them.

Thanks!

cmd/utils.go
kylegc: Agree, see new comment below.

cmd/utils.go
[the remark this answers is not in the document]
skoh7645: Very true...I have addressed this now

cmd/utils.go
skoh7645: Done

cmd/utils.go
skoh7645: Done

cmd/utils.go
skoh7645: Done

PR #318, opened 2019-09-13
test: drain stale repo before adding

gireeshpunathil:
tests that would have run before this and crashed could leave the repo in a bad state - mostly leading to: "a repo with this name already exists." this is because the failed tests do not get a chance to run the cleanup routines which is supposed to remove the repo.

so clean it up pro-actively, ignore any errors. 

Helps in minimizing turbulence in test runs.


gireeshpunathil:
Fixes: #330

Issue #301, opened 2019-09-11
appsody run --docker-options prohibits -e


GregDritschler:
**Describe the bug**
It is not possible to pass environment variables to the container.

>appsody run --docker-options "-e spring_profiles_active=dev"
Running development environment...
Running command: docker[pull appsody/java-spring-boot2:0.3]
testing docker options [-e spring_profiles_active=dev]
[Error] -e is not allowed in --docker-options

**To Reproduce**
appsody run --docker-options "-e spring_profiles_active=dev"

**Expected behavior**
Environment variable is passed to the container.

**Actual behaviour**
appsody disallows passing any environment variable to the container.

**Environment Details (please complete the following information):**
 - OS: Windows 10

**If applicable please specify:**
 - CLI version: 0.4.2
 - Stack you are using: java-spring-boot2

**Screenshots**

**Additional context**
1) It is very common to pass dynamic configuration into containers via environment variables.  For example, in Kubernetes the values of secrets can be passed as environment variables.  It is likely that developers will want to be able to imitate this behavior when using appsody run.

2) Profiles are a very popular Spring feature but there's no way to select one dynamically due to this restriction.

gireeshpunathil:
Agree that disallowing to set running environment will handicap developers heavily. IMO either of:
 - print a warning message if `-e` option is set irrespective of `APPSODY..` env is touched or not
 - print a warning message if `-e` option is set and `APPSODY..` env is touched

chilanti:
Fixed by PR #320

PR #320, opened 2019-09-13
Appsody to allow environment vars to be passed to Docker

chilanti:
Eliminated the filter that prevents `-e` or `--env` parms to be specified in `--docker-options`.


chilanti:
Fixes #301

PR #324, opened 2019-09-16
Cache and reuse the results of getEnvVar()

pushkarnk:
Motivation:
Commands may call getEnvVar() multiple times for different environment
variable names. Every invocation needs a `docker inspect` invocation followed
by unmarshalling of the json and a walk through the resulting array. Caching
the environment variables in a map can avoid repeating these steps and speed-up
command execution.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -50,6 +50,7 @@ type RootCommandConfig struct {
 	// package scoped, these are mostly for caching
 	setupConfigRun bool
 	imagePulled    map[string]bool
+	cachedEnvVars  map[string]string
 }
 
 // Regular expression to match ANSI terminal commands so that we can remove them from the log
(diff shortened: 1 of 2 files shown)


pushkarnk:
cc @gireeshpunathil

gireeshpunathil:
LGTM! this also solves the issue of too many logging of environment variables (right now it prints the whole list per invocation of this API) and restrict it to once per CLI invocation.

chilanti:
Hi Pushkar - thanks for the PR - can I ask you to open an issue and associate it with the PR? We are trying to get every PR associated w/ an issue. Also - opening an issue in advance would help us with planning (and is likely to prevent some duplication of work). Thanks.

pushkarnk:
@chilanti Created #335

pushkarnk:
Moved the cache from being a global var, into RootCommandConfig.

PR #325, opened 2019-09-16
test: wait for container to come up in debug test

gireeshpunathil:
Two improvements:
1. Right now the test is not waiting for the channel that
runs the goroutine for `appsody debug`, basically not checking
whether it was initiated / ran or not. Put up the container
check loop to make sure the container appears.

2. Name the container uniquely, deriving the name from the stack
so that back-to-back runs / turbulence in the test system do not
confuse the test logic - i.e., read and / or stop a wrong container.


chilanti:
@gireeshpunathil - can I ask you to create an issue and link it to this PR? We'd like to get into a modus operandi where all PRs have associated issues. Thanks.

gireeshpunathil:
Fixes: #329

Issue #298, opened 2019-09-11
Option to init git repo during appsody init


smcclem:
**Is your feature request related to a problem? Please describe.**
Feature request. The developer uses appsody init to create an appsody project on the local filesystem. Usually after that, they want to connect and push to a git repo, which they must do manually, outside of the appsody cli.  

**Describe the solution you'd like**
I would like a new , optional flag (-g ?) added to the appsody init command where a git repo can be specified. appsody init will run the basic git commands to push to a new repo.  So running the command :

appsody init [stack] -g https://github.com/smcclem/test.git

will result in these commands being run by appsody: 

git init
git add *
git commit -m "Appsody project for [stack]"
git remote add origin https://github.com/smcclem/test.git
git push -u origin master 

**Describe alternatives you've considered**
None. 

**Additional context**
None.

kylegc:
I think this is a cool idea. We could also use the git repo in other ways such as cloning an existing appsody project from git and forking the stacks repo to contribute a stack. 

Moving this to the backlog to get picked up.

pushkarnk:
In my opinion, we must restrict to the `git init` and `git remote add ...` and probably `git add ...`.  Not all users may appreciate an automated `git commit` and `git push`. Eager to know what others think.

pushkarnk:
Did a basic implementation in #326 I'm doing a `git init`, `git remote add...` and `git add` for now. I'm unsure if we should automate the `git commit ...` and `git push ..`. Open for suggestions.

neeraj-laad:
I am not sure if people are really after this. Moving to the backlog and bring this in if there is more activity on this.

PR #326, opened 2019-09-16
Initialize a git repo during appsody init

pushkarnk:
*Motivation*:
This is related to feature request
https://github.com/appsody/appsody/issues/298. Typically, after creating an
appsody project, Jane, the developer, may want to connect it to an external git
repository and push it there. The `appsody init` workflow can automate this by
invoking some of the common `git` commands on Jane's behalf.


chilanti:
@kylegc @neeraj-laad - this PR addresses the "fast track" onboarding of Jane. Currently, it does `git init` and `git add remote`, followed by a `git add *`. Doesn't do `commit` and `push`. Should we:
1) Leave it as it is
2) Always commit and push upon init, if Jane specifies `--git-repo`
3) Add another flag `--push` which would commit and push (the comment on `commit` could be something like : "Appsody initial commit"

pushkarnk:
@chilanti @neeraj-laad @kylegc I'd vote for not automating a commit and push because the flag name `--git-repo` makes it a little non-obvious. We don't want to surprise developers. The `--push` flag sounds good, but I'm keen on understanding what others think.

pushkarnk:
> @kylegc @neeraj-laad - this PR addresses the "fast track" onboarding of Jane. Currently, it does `git init` and `git add remote`, followed by a `git add *`. Doesn't do `commit` and `push`. Should we:
> 
> 1. Leave it as it is
> 2. Always commit and push upon init, if Jane specifies `--git-repo`
> 3. Add another flag `--push` which would commit and push (the comment on `commit` could be something like : "Appsody initial commit"

@neeraj-laad @chilanti @kylegc Keen to know your thoughts on which of the three alternatives makes the most sense.

neeraj-laad:
How about this:

`--git-repo` : init, add remote
`--git-repo --push`: init, add remote, add *,  commit and push

PR #327, opened 2019-09-16
added tests for linter

Kamran64:


kylegc:
@Kamran64 It looks like these tests are failing on Travis.

Kamran64:
@kylegc managed to fix on Windows but for some reason the Linux build is struggling to find the test-stack contents. Thought it was an issue with relative paths but changing them to the full path is still failing. Will continue to investigate.

Issue #31, opened 2019-07-01
Adopt a volume-mounting Docker syntax that allows colons in path names


kewegner:
This issue has to do with the fact that:

1. Windows path names are likely to include the drive ID and a colon separator

2. Even Unix-style path and file names are allowed to include colons (try touch foo:bar and you'll see what I mean).

For now: 
1.  We mount volumes with -v : which won't work in case has a colon in it. We need to switch to the --mount syntax (or use the Go Docker APIs appropriately).
2. The issue is been fixed "for Windows" - assuming that the only extra colon we get can occur after the drive ID (i.e. C:/:).

3. The controller also has this issue when it defaults the WATCH_DIR from MOUNTS

@kylegc did a little more research on this and discussed with @chilanti . Even if we switch to the --mount syntax, we still have the issue of splitting the list of mounts with the ';' separator char.

The Ultimate solution is as follows:

1. Allow either docker -v or --mount syntax in KNAPP_MOUNTS var with a default separator of ';'. We can determine that --mount syntax is used if "src=" and "dst=" exist.
2. We provide an optional KNAPP_SEPARATOR var that allows the user to change the default separator of ';' to any string sequence they'd like.

Until this solution is implemented,  it will be documented that ':' and ';' characters are not supported in the mount paths.

PR #338, opened 2019-09-18
stacks: make mount separator configurable

gireeshpunathil:
Right now, `;` is hard-coded to be the mount separator
in the stack-mounts, which inhibits this character from
being used for path names.

Introduce an appsody env APPSODY_MOUNT_SEPARATOR that can
be used by stacks that want to use a different separator.

Partially fixes: https://github.com/appsody/appsody/issues/31

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -188,7 +188,12 @@ func getVolumeArgs() ([]string, error) {
 		Warning.log("The stack image does not contain APPSODY_MOUNTS")
 		return volumeArgs, nil
 	}
-	stackMountList := strings.Split(stackMounts, ";")
+	mountSeparator, err := GetMountSeparator()
+	if err != nil {
+		Warning.log("Could not determine mount separator from The stack image")
+		return volumeArgs, err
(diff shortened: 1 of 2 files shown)


kewegner:
The Appsody controller also Inspects Appsody Mounts, so it would have to be modified I believe to handle this change.

chilanti:
I am not aware of any stacks hitting this - but Unix file systems to allow special characters in path and file names (including ":" which is a nasty problem to address). See also: #31 .

kewegner:
This will require issue #31 for appsody controller
In addition, I believe this line:
163	mountList := strings.Split(mountListSource, ";")
in stack_lint_dockerfile_stack.go will need to be updated.

neeraj-laad:
We expect more changes in this area and would prefer to get a clear design agreed across various components - stacks, cli, controller etc. Other priorities have not allowed us to focus on this use case as much.

I'm closing this for now, but once we have more clarity on how we want to enable this use case, we can re-open this PR again.

PR #343, opened 2019-09-18
Update default Controller version to 0.2.3

kewegner:
Updates Controller version to 0.2.3


kewegner:
This was done in #345

PR #351, opened 2019-09-19
support changing deployment name during appsody deploy

kewegner:
Searches for metadata: 
                       name: <name>
Rather than project name (directory of project)

Uses yaml parsing (structs) to find the right name in the app-deploy.yaml file.


kewegner:
User Id for commit not valid

Issue #346, opened 2019-09-19
`--docker-options` needs hardening


gireeshpunathil:
**Describe the bug**
The command line parser needs to make stricter rules allow only valid set of inputs

**To Reproduce**

```
#appsody run --docker-options=-t
Running development environment...
Running command: docker pull appsody/nodejs-express:0.2
testing docker options [-t]
[Error] -t is not allowed in --docker-options
```
 - good

```
#appsody run --docker-options=-it
Running development environment...
Running command: docker pull appsody/nodejs-express:0.2
testing docker options [-it]
Running docker command: docker run --rm -p 3000:3000 -p 9229:9229 --name foo-dev -v /Users/gireeshpunathil/Desktop/t/foo/:/project/user-app -v foo-deps:/project/user-app/node_modules -v /Users/gireeshpunathil/.appsody/appsody-controller:/appsody/appsody-controller -it -t --entrypoint /appsody/appsody-controller appsody/nodejs-express:0.2 --mode=run
[Container] the input device is not a TTY
[Error] Error waiting in 'appsody run' exit status 1
```
 - not so good

```
#appsody run --docker-options="ubuntu bash -c whoami"
Running development environment...
Running command: docker pull appsody/nodejs-express:0.2
testing docker options [ubuntu bash -c whoami]
Running docker command: docker run --rm -p 3000:3000 -p 9229:9229 --name foo-dev -v /Users/gireeshpunathil/Desktop/t/foo/:/project/user-app -v foo-deps:/project/user-app/node_modules -v /Users/gireeshpunathil/.appsody/appsody-controller:/appsody/appsody-controller ubuntu bash -c whoami -t --entrypoint /appsody/appsody-controller appsody/nodejs-express:0.2 --mode=run
[Container] Unable to find image 'ubuntu:latest' locally
[Container] latest: Pulling from library/ubuntu
[Container] 5667fdb72017: Pulling fs layer
...
[Container] Digest: sha256:b88f8848e9a1a4e4558ba7cfc4acc5879e1d0e7ac06401409062ad2627e6fb58
[Container] Status: Downloaded newer image for ubuntu:latest
[Container] root
#
```
 - it pulls the ubuntu image, not good

```
#appsody run --docker-options=--help
Running development environment...
Running command: docker pull appsody/nodejs-express:0.2
testing docker options [--help]
[...]

PR #355, opened 2019-09-20
cmd: harden `--docker-options` parsing

gireeshpunathil:
~~Disallow anything that does not start with a `-`.
(luckily all the docker run options do)~~

EDIT: As per discussion, this was not pursued due to maintenance overhead.

Disallow `--help`
(not the best place to do `docker run --help`)
(in addition, even if it is, the output is truncated, so!)

Fixes: https://github.com/appsody/appsody/issues/346

cc @kylegc @chilanti @kewegner


kewegner:
I think the change for --help is ok, I am not sure ^[^-].*)| is good though.  The reason is it is checking the whole string, not just the options so for instance if I have -m 4g  will this work if the regex runs against the '4g'.  I will test this out and get back to you.

gireeshpunathil:
@kewegner - I see! let me play around with regex a bit and see if there is a way; and meanwhile let use see what others think about this.

gireeshpunathil:
per conversation in the tracking issue, I will go just with blacklisting `--help` (makes sense to do, easy to do, good to do as its o/p is truncated atm)

gireeshpunathil:
adjusted the changes accordingly, @kewegner , @kylegc - ptal.

Issue #334, opened 2019-09-17
"sleeping for 60 seconds" message from CLI is confusing


henrynash:
When you run appsody stop (or CNTL C appsody run), the CLI says that it is sleeping for 60 seconds. I don't believe this is true. 

From @tnixa in duplicate #344:
**Describe the bug**
When i do an appsody run and kill is using ctrl c I get a message
`[Container] ^CClosing down development environment, sleeping 60 seconds: exit status 2`
which makes me thing it is sleeping for 60 seconds but it returns within a couple seconds. This message did not happen with the previous release in this scenario. If I do `appsody run -i`  and kill it with ctrl c I get the same message but it actually does a 60 second sleep before returning.

I'm also not sure what `exit status 2` is supposed to mean. I think I was getting `exit status 1` with the previous CLI release

**Environment Details (please complete the following information):**
 - OS: [e.g. iOS] Windows 10

**If applicable please specify:**
 - CLI version: 0.4.4 plus fixes
 - Stack you are using: nodejs

PR #356, opened 2019-09-20
[on hold] cmd: remove sleep in runner when controller is interrupted

gireeshpunathil:
1. Remove the 1 minute sleep in the runner
2. Adjust the messages accordingly.

Fixes: https://github.com/appsody/appsody/issues/334

cc @kylegc . It is not clear to me the need for `sleep`: In `tty` mode (ctrl+c intercepted by the controller) or not (ctrl+c intercepted by runner) , the `interrupt` signal is always obtained from the child (controller) , which means we don't need to wait anymore? Am I missing something? I tested in mac and linux, and did not see any issues with removal of sleep.

### DIFF
--- a/cmd/dev_common.go
+++ b/cmd/dev_common.go
@@ -24,7 +24,6 @@ import (
 	"runtime"
 	"strings"
 	"syscall"
-	"time"
 
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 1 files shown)


kylegc:
@kewegner was assigned #334 and he is investigating. What we know right now is that the sleep was there from the beginning of Appsody but we aren't sure why. We want to make sure we cleanup resources properly.

gireeshpunathil:
apologies @kylegc @kewegner - didn't see the assignment. For now I mark this PR as on hold, (can close too, if there was already a progress).

PR #370, opened 2019-09-25
--generate-only fixes for appsody deploy

kewegner:
Allows --tag option to take precedence for appsody deploy --generate-only
Allows --knative flag to be reflectd in appsody deploy --generate-only

Uses existing applicationImage tag in app-deploy.yaml for deployments, doesn't regen the tag if it is there.


henrynash:
Great!

Issue #287, opened 2019-09-09
Add --project-name option to init, add the name to .appsody-config.yaml, and validate it against k8s rules


mtamboli:
**Describe the bug**
If the name of the appsody directory is not what Kubernetes service likes, pod will not come up during deployment. Directory name for the appsody project is used for the service.

AppsodyApplication or KnativeService create ServiceSpec objects that must conform to RFC-1123 naming
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/
appsody init should check that the name conforms
https://github.com/kubernetes/kubernetes/blob/v1.15.3/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go

**To Reproduce**
1. mkdir jmp/1-2
2. cd jmp/1-2
3. appsody init Kabanero/ava-microprofile
4. appsody deploy --generate-only
5. app-deploy.xml file is created and it has the directory name "1-2" in that file.
Then the application was used with Kabanero pipeline, application pod did not start.
because of the deployment failure:
```
 Message:           Service "1-2" is invalid: metadata.name: Invalid value: "1-2": a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')
```

**Expected behavior**
I would expect error to be displayed during `appsody init` if directory name is going to be a problem. If not during appsody init, then definitely during `deploy --generate-only'. Need to catch the error earlier.

**Environment Details (please complete the following information):**
OpenShift 3.1.1
Kabanero release 0.1.2

**If applicable please specify:**
 - CLI version:
 - Stack you are using:

**Screenshots**

**Additional context**

chilanti:
This needs discussion - @neeraj-laad and @kyle. 
I think we have been sort of skirting the problem of special characters in Appsody paths for a while (in some cases we turn underscores into dashes, etc.) - but this is a bigger issue, with the potential for a number of consequences (see also #31).

skoh7645:
@chilanti In case 2 we would still need to check the project dir against the regex?

skoh7645:
@mtamboli we have to make sure the project name starts with an alphabetic char an ends on an alphanumeric char.
Perhaps we should only append/prepend if project name doesn't start/end on a valid char?

chilanti:
[...]

PR #375, opened 2019-09-26
Optional --project-name flag added to init command and validated project name added to .appsody-config.yaml

Bazif-Khan:
fixes: https://github.com/appsody/appsody/issues/287


cmd/init.go
```diff
@@ -37,6 +37,8 @@ type initCommandConfig struct {
 	noTemplate bool
 }
 
+var prjName string
```
chilanti: This needs to go into the `RootCommandConfig` or `initCommandConfig`- no more global vars.

cmd/utils.go
```diff
@@ -272,15 +273,54 @@ func getProjectConfig(config *RootCommandConfig) (ProjectConfig, error) {
...
 		err := v.ReadInConfig()
+
 		if err != nil {
 			var tempProjectConfig ProjectConfig
 			return tempProjectConfig, errors.Errorf("Error reading project config %v", err)
 
 		}
+
+		if prjName != "" {
+			prjName = strings.ToLower(prjName)
```
chilanti: The logic should be as we discussed in the issue: if the user provides a string, we should validate it against the Kubernetes regex. If that validation fails, we should simply tell the user that the name they provided can't be used. We shouldn't manipulate the string that is given to us.

cmd/utils.go
```diff
@@ -272,15 +273,54 @@ func getProjectConfig(config *RootCommandConfig) (ProjectConfig, error) {
...
+				return *config.ProjectConfig, errors.Errorf("This is not a valid project name")
+			}
+
+			v.Set("project-name", prjName)
+			err := v.WriteConfig()
+			if err != nil {
+				Error.log(err)
+			}
+
+			Info.log("Your project name is ", prjName)
```
chilanti: "Your Appsody project name is..."

cmd/utils.go
```diff
@@ -272,15 +273,54 @@ func getProjectConfig(config *RootCommandConfig) (ProjectConfig, error) {
...
+			match, _ := regexp.MatchString("[a-z]([-a-z0-9]*[a-z0-9])?", prjName)
+
+			if !match {
+				return *config.ProjectConfig, errors.Errorf("This is not a valid project name")
+			}
+
+			v.Set("project-name", prjName)
+			err := v.WriteConfig()
+			if err != nil {
+				Error.log(err)
```
chilanti: Do we carry on even if this errors out? I think we should return the err.

cmd/utils.go
```diff
@@ -272,15 +273,54 @@ func getProjectConfig(config *RootCommandConfig) (ProjectConfig, error) {
...
+			reg, err := regexp.Compile("[^a-z0-9]+")
+			if err != nil {
+				Error.log(err)
+			}
+			projectName := reg.ReplaceAllString(project, "-")
+
+			v.Set("project-name", projectName)
+			err = v.WriteConfig()
+			if err != nil {
+				Error.log(err)
```
chilanti: Same as above.

cmd/utils.go
```diff
@@ -272,15 +273,54 @@ func getProjectConfig(config *RootCommandConfig) (ProjectConfig, error) {
...
+			}
+
+			Info.log("Your project name is ", prjName)
+		} else {
+			project := "appsody-" + strings.ToLower(filepath.Base(dir)) + "-app"
+			reg, err := regexp.Compile("[^a-z0-9]+")
+			if err != nil {
+				Error.log(err)
+			}
+			projectName := reg.ReplaceAllString(project, "-")
```
chilanti: Even after the substitution, we should check that the name is still compliant with the K8S regex. I understand that it should be at this point - but better double check.

cmd/utils.go
```diff
@@ -301,13 +341,16 @@ func getOperatorHome(config *RootCommandConfig) string {
...
 	}
-	projectName := strings.ToLower(filepath.Base(projectDir))
-	projectName = strings.ReplaceAll(projectName, "_", "-")
-	return projectName, nil
+
+	appsodyConfig := filepath.Join(dir, ConfigFile)
+	v := viper.New()
+	v.SetConfigFile(appsodyConfig)
+	err = v.ReadInConfig()
+	return v.GetString("project-name"), err
```
chilanti: @kylegc and I discussed a bit here. This is a get method - but... it may happen that someone manually manipulates the configuration file. So - we should do two things:
* check that the name is still compliant with the k8s regex. If it isn't, Error.log and error out.
* if we don't find anything (i.e. someone happily deleted the project name from the file) we should re-gen it (just like we did in `init`) and set it in the file. Warning.log("Your project name is now...").

cmd/utils.go
Bazif-Khan: I have changed it; however, I was manipulating the string because I was thinking it was better to manipulate the string and the let user know the name of the project rather than displaying the error.

cmd/utils.go
Bazif-Khan: changed

cmd/init.go
Bazif-Khan: Moved into the initCommandConfig

cmd/utils.go
Bazif-Khan: changed

cmd/utils.go
Bazif-Khan: changed

cmd/utils.go
Bazif-Khan: changed

cmd/utils.go
Bazif-Khan: Done

cmd/utils.go
```diff
@@ -263,6 +264,62 @@ func getProjectDir(config *RootCommandConfig) (string, error) {
...
 
+func setProjectName(config *RootCommandConfig, projectName string) error {
+	dir, perr := getProjectDir(config)
+	if perr != nil {
+		return errors.Errorf("The current directory is not a valid appsody project. Run appsody init <stack> to create one: %v", perr)
+	}
+	appsodyConfig := filepath.Join(dir, ConfigFile)
+	v := viper.New()
+	v.SetConfigFile(appsodyConfig)
+	err := v.ReadInConfig()
```
EnriqueL8: Do not need to read, since we are only writing

cmd/utils.go
```diff
@@ -301,13 +363,46 @@ func getOperatorHome(config *RootCommandConfig) string {
...
-	projectName := strings.ToLower(filepath.Base(projectDir))
-	projectName = strings.ReplaceAll(projectName, "_", "-")
-	return projectName, nil
+
+	appsodyConfig := filepath.Join(dir, ConfigFile)
+	v := viper.New()
+	v.SetConfigFile(appsodyConfig)
+	err = v.ReadInConfig()
+
+	match, _ := regexp.MatchString("[a-z]([-a-z0-9]*[a-z0-9])?", v.GetString("project-name"))
```
EnriqueL8: Still need to error out if the `project-name` exists, but does not match the regex.

cmd/init.go
```diff
@@ -255,23 +256,28 @@ func install(config *initCommandConfig) error {
 		return errors.Errorf("%v", perr)
 
 	}
-	projectConfig, configErr := getProjectConfig(config.RootCommandConfig)
-	if configErr != nil {
-		return configErr
-	}
-	platformDefinition := projectConfig.Platform
+	err := setProjectName(config.RootCommandConfig, config.projectName)
+	if err == nil {
```
EnriqueL8: Suggestion: keep consistent error checking with `if err != nil`

cmd/utils.go
Bazif-Khan: We need ReadInConfig() for writing into the file as well.

cmd/init.go
Bazif-Khan: changed

cmd/utils.go
Bazif-Khan: changed

cmd/init.go
```diff
@@ -258,6 +256,10 @@ func install(config *initCommandConfig) error {
 		return errors.Errorf("%v", perr)
 
 	}
+	err := setProjectName(config.RootCommandConfig, config.projectName)
```
chilanti: I would suggest you avoid passing in the rootConfig here. The only reason why you are doing it is because `setProjectName` then calls `dir, perr := getProjectDir(config)`. I would suggest doing that call here, and passing in the projectDir as a string to setProjectName. I want to avoid the proliferation of functions that require the entire configuration struct to be passed in even though it's needed just to get something else.

cmd/utils.go
```diff
@@ -264,6 +264,62 @@ func getProjectDir(config *RootCommandConfig) (string, error) {
 	return config.ProjectDir, nil
 }
 
+func setProjectName(config *RootCommandConfig, projectName string) error {
+	dir, perr := getProjectDir(config)
```
chilanti: As mentioned previously, I would remove this call from here and do it in the caller. Pass in two strings - `projectDir` and `projectName`.

cmd/utils.go
```diff
@@ -350,7 +372,52 @@ func getProjectName(config *RootCommandConfig) (string, error) {
...
+		return "my-project", err
+	}
+
+	projectName := v.GetString("project-name")
+
+	if projectName != "" {
+		match, err := regexp.MatchString("[a-z]([-a-z0-9]*[a-z0-9])?", projectName)
+
+		if !match {
+			v.Set("project-name", "my-project")
```
chilanti: I thought we agreed that if this isn't a match we would just error out with a message that explains that the project name must conform to those rules. Here we are changing it to `my-project` if I am not mistaken.

cmd/init.go
Bazif-Khan: Done

cmd/utils.go
Bazif-Khan: Done

cmd/init.go
```diff
@@ -74,11 +75,11 @@ setup the local dev environment.`,
 
 	initCmd.PersistentFlags().BoolVar(&config.overwrite, "overwrite", false, "Download and extract the template project, overwriting existing files.  This option is not intended to be used in Appsody project directories.")
 	initCmd.PersistentFlags().BoolVar(&config.noTemplate, "no-template", false, "Only create the .appsody-config.yaml file. Do not unzip the template project. [Deprecated]")
+	initCmd.PersistentFlags().StringVar(&config.projectName, "project-name", "", "Project Name for Kubernetes Service")
```
kylegc: It would be nice to set the default here, so the `appsody help init` command will display the proper default. 
See `run` and `extract` commands for examples of how to do that.

cmd/utils.go
Bazif-Khan: changed

cmd/init.go
Bazif-Khan: If I set the default value of projectName in the way you are setting in extract.go for extractContainerName then the default value is always set to my-project because we are setting projectName based on current directory name and we can't get the directory name until we init the project because getProjectDir() function gives you the dir name only if rootConfig has the value and it has the value when project is initialized.

kylegc: The `defaultExtractContainerName()` method works today, such that `appsody help extract` will lists the correct default name based on the projectDir, not "my-project". 

Are you saying this PR breaks that?

I don't see a reason why you can’t run the same logic (reusing code) to calculate the default `init --project-name` for the help text in a similar way to extract's name. The `config.ProjectDir` should be set by then. 

I feel pretty strongly about this because it impacts the user’s first impression of appsody.

Bazif-Khan: I have changed it. Now I'm also getting the default name similar to the way you are getting default Name. You are also getting name always "my-project" but you are just appending "-extract" at the end of "my-project".

cmd/init.go
```diff
@@ -576,3 +577,42 @@ func parseProjectParm(projectParm string, config *RootCommandConfig) (string, st
...
+
+		if err != nil {
+			Error.log(err)
+			os.Exit(1)
+		}
+
+		indexPathSeparator := strings.LastIndex(projectDirPath, string(os.PathSeparator))
+
+		projectName := projectDirPath[indexPathSeparator+1:]
+
```
chilanti: I believe that these two lines of code are unnecessary, because `filepath.Base` already gets you the leaf.

cmd/init.go
```diff
@@ -576,3 +577,42 @@ func parseProjectParm(projectParm string, config *RootCommandConfig) (string, st
...
+		match, _ = regexp.MatchString("[a-z]([-a-z0-9]*[a-z0-9])?", projectName)
+
+		if match {
+			return projectName
+		}
+
+		return "my-project"
+	}
+	return projectName
+}
```
chilanti: This logic (or a small variation thereof) is repeated three times - perhaps we can check this in as it is now and refactor later? @kylegc what do you think...?

Bazif-Khan: I have refactored it

tnixa:
I tested this code and found a few problems...
1. `appsody deploy` will fail with a project name of `python-project_Test` due to docker not liking the capital letter
2. `appsody deploy` will fail with a project name of `python-project_test` due to kubectl apply not liking the underscore
3. with a project directory of `node.js` i get an error with `appsody deploy` due to a problem with the dot. i believe the code is supposed to convert dots to dashes but this doesn't appear to be working

Bazif-Khan:
> I tested this code and found a few problems...
> 
> 1. `appsody deploy` will fail with a project name of `python-project_Test` due to docker not liking the capital letter
> 2. `appsody deploy` will fail with a project name of `python-project_test` due to kubectl apply not liking the underscore
> 3. with a project directory of `node.js` i get an error with `appsody deploy` due to a problem with the dot. i believe the code is supposed to convert dots to dashes but this doesn't appear to be working

Sorry for not testing these cases. I forgot/missed to check these cases.

@kylegc @chilanti @tnixa I was using regex which was in this comment: https://github.com/appsody/appsody/issues/287#issuecomment-530952985 so it was failing. I have changed the regex. Now the rules which I'm using are:
- Length: less than 128 characters
- Start: must start with the lowercase letter
- End: must end with lowercase letter or digit.

I have created a utility function for validating the name called validateKubernetesResourceName() which I'm using for validating project name and stack name.

PR #385, opened 2019-10-01
appsody stack validate and package commands

tnixa:


cmd/stack_validate_build.go
```diff
@@ -0,0 +1,76 @@
...
+	"strings"
+	"time"
+)
+
+// Simple test for appsody build command. A future enhancement would be to verify the image that gets built.
+func TestBuild(projectDir string) error {
+
+	// appsody build
+
+	imageName := "testbuildimage"
```
kylegc: I think this should be something like `dev.local/nodejs` where `nodejs` is the stack name (current directory for now, but eventually should be pulled from stack.yaml)

cmd/stack_validate_build.go
```diff
@@ -0,0 +1,76 @@
...
+	Info.Log("Running appsody build")
+	Info.Log("******************************************")
+	_, err := RunAppsodyCmdExec([]string{"build", "--tag", imageName}, projectDir)
+	if err != nil {
+		Error.Log(err)
+		return err
+	}
+
+	// It will take a while for the image to build, so lets use docker image ls to wait for it
+	fmt.Println("calling docker image ls to wait for the image")
```
kylegc: Looks to me like the `appsody build` command runs synchronously, so do we really need this code to "wait" for the build to finish?

A `docker image ls` is still good to make sure the image was created. What other tests could we do on the image? Maybe a docker run? This can be a future PR.

cmd/stack_validate_init.go
```diff
@@ -0,0 +1,29 @@
...
+
+// Simple test for appsody build command. A future enhancement would be to verify the image that gets built.
+func TestInit(stack string, projectDir string) error {
+
+	Info.Log("******************************************")
+	Info.Log("Running appsody init for stack: " + stack)
+	Info.Log("******************************************")
+	_, err := RunAppsodyCmdExec([]string{"init", stack}, projectDir)
+	if err != nil {
+		Error.Log(err)
```
kylegc: You can probably just return err here and let the caller log it. In fact, since this is all this file is doing, you probably don't need this func in a separate file.

cmd/stack_validate_run.go
```diff
@@ -0,0 +1,91 @@
...
+
+	if !isHealthy {
+		Error.Log("appsody ps never found the correct container")
+		return errors.New("appsody ps never found the correct container")
+	}
+
+	Info.Log("Appsody run did not fail")
+
+	// stop and clean up after the run
+	func() {
```
kylegc: This does not need to be surrounded by a func()

cmd/stack_validate_testcmd.go
```diff
@@ -0,0 +1,30 @@
...
+
+// Simple test for appsody build command. A future enhancement would be to verify the image that gets built.
+func TestTest(projectDir string) error {
+
+	Info.Log("******************************************")
+	Info.Log("Running appsody test")
+	Info.Log("******************************************")
+	_, err := RunAppsodyCmdExec([]string{"test"}, projectDir)
+	if err != nil {
+		Error.Log(err)
```
kylegc: Same comment as init. just return the error and this probably doesn't justify a separate file.

Issue #45, opened 2019-07-01
Enable stack development using CLI


kylegc:
- create stack
- test stack

Re-use tests and invoke them using CLI
- linting
- test platform with all templates

### Ideas from @henrynash in #262 
When development stacks, the local build process (i.e. using the ci scripts), while reasonable, requires the developer to understand a bit too much...and is somewhat cumbersome. It might be better if the appsody CLI itself supported this process directly, e.g.:

appsody stack cmd [stack-name] [local-repo-name]

cmd - can be build, lint, test or package

stack-name, if not specified would default to the current directory (e.g. experimental/my-stack).

local-repo-name, if not specified, would default to "local-test"

We might also add a flag to 'appsody init' to use local repos (i.e. the equivalent of setting 'export APPSODY_PULL_POLICY=IFNOTPRESENT')

nastacio:
It will be great to see this working cross-platform, including Windows. Thanks for moving this forward.

PR #389, opened 2019-10-02
Windowsfileurlfix

tnixa:
Related to #45


henrynash:
Is there an issue/feature request # that this links to? It would be useful to reference this in the description.

Issue #368, opened 2019-09-25
Implement the `appsody stack create` command


kylegc:
The first pass of an `appsody stack create` command will create a new stack as a copy of an existing stack.

FUTURE: in the future, we may add an `--extends` option to create a new stack that inherits from an existing stack. This is out of scope right now.

Usage: `appsody stack create <name> --copy <stack>`
* --copy is optional, without it, default to "samples/sample-stack"
Example: `appsody stack create java-liberty --copy incubator/java-microprofile`

## Steps of the command --copy
1. Create a directory called <name> under the current directory (maybe add a --target option for the parent dir)
2. Download "--copy stack" from git (using http github rest api)
3. (If not already done by step 2) Copy stack contents to directory created in step 1

neeraj-laad:
@appsody/cli just wondering if we will always create a stack based on appsody/stacks repo. OR should we allow users to specify their own git repo. I'm thinking users or enterprises who might be maintaining a fork of appsody/stacks.

kylegc:
Good question Neeraj. I think for now assuming the appsody/stacks repo is okay, but I agree the enterprise scenario needs to be addressed. 

We could have an option on the command to specify the git repo, but I'm thinking we can do better. Eventually we should be able to link the stacks back to their github repo source via either the stack image labels or the repo index. We could use that information to then download the source for the `stack create` command.

PR #392, opened 2019-10-03
Stack create command

Bazif-Khan:
fixes: https://github.com/appsody/appsody/issues/368

### DIFF
--- a/cmd/stack_create.go
+++ b/cmd/stack_create.go
@@ -57,7 +57,7 @@ The stack name must start with a lowercase letter, and can contain only lowercas
 
 			stack := args[0]
 
-			match, err := isValidContainerName(stack)
+			match, err := IsValidProjectName(stack)
 
 			if !match {
 				return err
--- a/cmd/stack_create.go
+++ b/cmd/stack_create.go
@@ -57,7 +57,7 @@ The stack name must start with a lowercase letter, and can contain only lowercas
 
 			stack := args[0]
 
-			match, err := validateKubernetesResourceName(stack)
+			match, err := isValidContainerName(stack)
 
 			if !match {
 				return err
--- a/cmd/stack_create.go
+++ b/cmd/stack_create.go
@@ -43,7 +43,7 @@ func newStackCreateCmd(rootConfig *RootCommandConfig) *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 
 			if len(args) < 1 {
-				return errors.New("Required parameter missing. You must specify a stack name.")
+				return errors.New("Required parameter missing. You must specify a stack name")
 			}
 
 			stack := args[0]
@@ -60,7 +60,7 @@ func newStackCreateCmd(rootConfig *RootCommandConfig) *cobra.Command {
 			}
 
 			if exists {
-				return errors.New("A stack named" + stack + "already exists in your directory. Specify a unique stack name.")
+				return errors.New("A stack named" + stack + "already exists in your directory. Specify a unique stack name")
 			}
 
 			err = downloadFileToDisk("https://github.com/appsody/stacks/archive/master.zip", filepath.Join(getHome(rootConfig), "extract", "repo.zip"), config.Dryrun)
--- a/cmd/stack_create.go
+++ b/cmd/stack_create.go
@@ -125,10 +125,12 @@ func unzip(src string, dest string, copy string) (bool, error) {
 
 		fileName := strings.Replace(f.Name, string(os.PathSeparator)+"stacks-master", "", -1)
 		Debug.log(fileName)
-		Info.log(fileName)
+
+		Info.log(filepath.Join("stacks-master", string(os.PathSeparator), copy) + string(os.PathSeparator))
 		if !strings.HasPrefix(fileName, filepath.Join("stacks-master", string(os.PathSeparator), copy)+string(os.PathSeparator)) {
 			continue
 		} else {
+			Info.log("Inside Else----------------------")
 			valid = true
 		}
 
--- a/cmd/stack_create.go
+++ b/cmd/stack_create.go
@@ -20,6 +20,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -58,7 +59,7 @@ func newStackCreateCmd(rootConfig *RootCommandConfig) *cobra.Command {
 				return errors.New("This stack named " + stack + " already exists")
 			}
 
-			err = downloadFileToDisk("https://github.com/appsody/stacks/archive/master.zip", filepath.Join(getHome(rootConfig), "/extract/repo.zip"), config.Dryrun)
+			err = downloadFileToDisk("https://github.com/appsody/stacks/archive/master.zip", filepath.Join(getHome(rootConfig), "extract", "repo.zip"), config.Dryrun)
 			if err != nil {
 				return err
 			}
@@ -67,7 +68,7 @@ func newStackCreateCmd(rootConfig *RootCommandConfig) *cobra.Command {
 				return err
 			}
 
-			valid, unzipErr := unzip(filepath.Join(getHome(rootConfig), "/extract/repo.zip"), stack, config.copy)
+			valid, unzipErr := unzip(filepath.Join(getHome(rootConfig), "extract", "repo.zip"), stack, config.copy)
 
 			if unzipErr != nil {
 				return unzipErr
@@ -78,10 +79,10 @@ func newStackCreateCmd(rootConfig *RootCommandConfig) *cobra.Command {
 			}
 
 			//deleting the stacks repo zip
-			os.Remove(filepath.Join(getHome(rootConfig), "/extract/repo.zip"))
+			os.Remove(filepath.Join(getHome(rootConfig), "extract", "repo.zip"))
 
 			//moving out the stack which we need
-			err = os.Rename(filepath.Join(stack, "/stacks-master/", config.copy), projectType)
+			err = os.Rename(filepath.Join(stack, "stacks-master", config.copy), projectType)
 			if err != nil {
 				return err
 			}
@@ -123,6 +124,11 @@ func unzip(src string, dest string, copy string) (bool, error) {
 			return valid, fmt.Errorf("%s: illegal file path", fpath)
 		}
 
+		if runtime.GOOS == "windows" {
+			if strings.Contains(copy, "/") {
+				copy = strings.Replace(copy, "/", "\\", -1)
+			}
+		}
 		fileName := strings.Replace(f.Name, string(os.PathSeparator)+"stacks-master", "", -1)
 		if !strings.HasPrefix(fileName, filepath.Join("stacks-master", string(os.PathSeparator), copy)+string(os.PathSeparator)) {
 			continue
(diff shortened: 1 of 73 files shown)


cmd/utils.go
```diff
@@ -1060,3 +1060,63 @@ func IsEmptyDir(name string) bool {
...
+
+	defer resp.Body.Close()
+	if resp.StatusCode != 200 {
+		buf, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			Debug.log("Could not read contents of response body: ", err)
+		} else {
+			Debug.logf("Contents http response:\n%s", buf)
+		}
+		resp.Body.Close()
```
EnriqueL8: Since we are moving this function, we could fix this bug. This is already performed by the `defer`.

cmd/utils.go
```diff
@@ -1060,3 +1060,63 @@ func IsEmptyDir(name string) bool {
...
+		return fmt.Errorf("Could not download %s: %s", href, resp.Status)
+	}
+
+	_, err = io.Copy(writer, resp.Body)
+	if err != nil {
+		return fmt.Errorf("Could not copy http response body to writer: %s", err)
+	}
+	resp.Body.Close()
```
EnriqueL8: Same as above.

cmd/stack_create.go
```diff
@@ -0,0 +1,190 @@
...
+type stackCreateCommandConfig struct {
+	*RootCommandConfig
+	copy string
+}
+
+func newStackCreateCmd(rootConfig *RootCommandConfig) *cobra.Command {
+	config := &stackCreateCommandConfig{RootCommandConfig: rootConfig}
+
+	var stackCmd = &cobra.Command{
+		Use:   "create",
```
kylegc: Since this command requires a name argument, the Use: should be `"create <name>"`

cmd/stack_create.go
```diff
@@ -0,0 +1,190 @@
...
+		If a copy flag is specified, stack create command will create a new stack as a copy of that existing stack.`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+
+			if len(args) < 1 {
+				return errors.New("Stack create command should have a stack name. Run `appsody stack create <name>` to create one")
+			}
+
+			stack := args[0]
+
+			exists, _ := Exists(stack)
```
kylegc: You should handle the error instead of ignoring it.

cmd/stack_create.go
```diff
@@ -0,0 +1,190 @@
...
+				err = os.Rename("sample-stack", stack)
+				if err != nil {
+					return err
+				}
+			}
+
+			return nil
+		},
+	}
+	stackCmd.PersistentFlags().StringVar(&config.copy, "copy", "", "Copy existing stack")
```
kylegc: Set the default here to `"sample-stack"`, and then you don't need the big `else` block above to handle config.copy == "".

cmd/stack_create.go
```diff
@@ -0,0 +1,190 @@
...
+
+			stack := args[0]
+
+			exists, _ := Exists(stack)
+
+			if exists {
+				return errors.New("This stack named " + stack + " already exists")
+			}
+
+			err := downloadFolderToDisk("https://github.com/appsody/stacks/archive/master.zip", getHome(rootConfig)+"/extract/repo.zip")
```
kylegc: Were you not able to download just a single stack from the repo?

cmd/stack_create.go
```diff
@@ -0,0 +1,190 @@
...
+			if err != nil {
+				return err
+			}
+
+			if config.copy != "" {
+				repoIndex := strings.Index(config.copy, "/")
```
kylegc: Don't assume there's a "/" here. You should use the same logic as init.go's parseProjectParm() to get the stack.

cmd/stack_create.go
```diff
@@ -0,0 +1,190 @@
...
+
+				copiedStack := config.copy[repoIndex+1:]
+
+				valid, unzipErr := unzip(getHome(rootConfig)+"/extract/repo.zip", stack, config.copy)
```
kylegc: All of the places you construct a filesystem path, you need to use `filepath.Join()` so that it works on windows.

cmd/stack_create.go
```diff
@@ -0,0 +1,190 @@
...
+					return errors.Errorf("This is not a valid stack. Please specify any existing stack as <repo>/<stack>")
+				}
+				os.Remove(getHome(rootConfig) + "/extract/repo.zip")
+
+				err := os.Rename(stack+"/stacks-master/"+config.copy, copiedStack)
+				if err != nil {
+					return err
+				}
+
+				os.RemoveAll(stack)
```
kylegc: Lots of renaming and removing stuff. Can you make this more efficient with less `os` calls? Additionally, some code comments to explain what is going on would be nice.

cmd/utils.go
```diff
@@ -1345,6 +1345,64 @@ func IsEmptyDir(name string) bool {
...
+	}
+
+	_, err = io.Copy(writer, resp.Body)
+	if err != nil {
+		return fmt.Errorf("Could not copy http response body to writer: %s", err)
+	}
+	return nil
+}
+
+func downloadFolderToDisk(url string, destFile string) error {
```
kylegc: There's a downloadFileToDisk method in init.go. Can you move that here and use that instead?

cmd/stack_create.go
Bazif-Khan: When I unzip a specific folder, e.g. nodejs stack, it's inside stacks-master/incubator/nodejs, and I want just nodejs folder and want to rename it. So, I am moving it out and rename it and delete the empty folder stacks-master/incubator.

cmd/stack_create.go
Bazif-Khan: GitHub API allows you to download single file only, so for downloading the whole folder I'll have to do multiple HTTP requests recursively and GitHub API only allows you 60 unauthorized requests per hour, so I thought it's better to download the whole repo with one request and then extract only the contents which are required.

kylegc: Ok thanks for the explanation. We'll keep what you have for now, but we may want to investigate alternative options as the stacks repo grows larger.

cmd/stack_create.go
```diff
@@ -0,0 +1,195 @@
...
+				return err
+			}
+			exists, err := Exists(stack)
+
+			if err != nil {
+				return err
+			}
+
+			if exists {
+				return errors.New("This stack named " + stack + " already exists")
```
helenmasters: A stack named ” + stack + ” already exists in your repository. Specify a unique stack name.

cmd/stack_create.go
```diff
@@ -0,0 +1,195 @@
...
+			}
+
+			valid, unzipErr := unzip(filepath.Join(getHome(rootConfig), "extract", "repo.zip"), stack, config.copy)
+
+			if unzipErr != nil {
+				return unzipErr
+			}
+
+			if !valid {
+				return errors.Errorf("This is not a valid stack. Please specify any existing stack as <repo>/<stack>")
```
helenmasters: "Invalid stack name:” + stack + ”. Stack name must be in the format <repository>/<stack>."
The above message should be used if this is only about the stack name format, if it refers to the stack not being found or potentially not existing the message should be more like...
"The stack that you specified:” + stack + ” is not valid. The stack must be specified as <repo>/<stack>."

cmd/stack_create.go
```diff
@@ -0,0 +1,182 @@
...
+
+		if runtime.GOOS == "windows" {
+			fileName := strings.Replace(f.Name, "/stacks-master", "", -1)
+			if !strings.HasPrefix(fileName, "stacks-master/"+copy+"/") {
+				continue
+			} else {
+				valid = true
+			}
+		} else {
+			fileName := strings.Replace(f.Name, string(os.PathSeparator)+"stacks-master", "", -1)
```
EnriqueL8: Just a bit confused about this logic: we are replacing  `"/stack-master"` with `""` , and then checking if the file path has a prefix of `"stack-master/<copy-arg>/"`. From printing these variables, `f.Name` never starts with `/stacks-master/` but just `stacks-master/`, maybe I am missing something? Is it because we were previously using `fpath` with starts with `<directory-name>/stacks-master/`?

cmd/stack_create.go
```diff
@@ -0,0 +1,182 @@
...
+		if f.FileInfo().IsDir() {
+			// Make Folder
+			err := os.MkdirAll(fpath, os.ModePerm)
+			if err != nil {
+				return valid, err
+			}
+			continue
+		}
+
+		// Make File
```
EnriqueL8: Just a suggestion, we create a separate function in the utils for this: such as`copyFile(src string, des, string)`.

cmd/stack_create.go
```diff
@@ -0,0 +1,182 @@
...
+			}
+
+			valid, unzipErr := unzip(filepath.Join(getHome(rootConfig), "extract", "repo.zip"), stack, config.copy)
+
+			if unzipErr != nil {
+				return unzipErr
+			}
+
+			if !valid {
+				return errors.Errorf("Invalid stack name:" + stack + ". Stack name must be in the format <repo>/<stack>")
```
EnriqueL8: Instead of `stack`, it should be `config.copy`. 
I get the following error:
image

Bazif-Khan: Yesterday I updated the error messages which @helenmasters recommended, and I accidentally wrote stack instead of copy.config. Thanks for testing

cmd/stack_create.go
Bazif-Khan: This line is not being used and I forgot to remove it. So I have deleted it. Thanks

cmd/stack_create.go
```diff
@@ -0,0 +1,180 @@
...
+	*RootCommandConfig
+	copy string
+}
+
+func newStackCreateCmd(rootConfig *RootCommandConfig) *cobra.Command {
+	config := &stackCreateCommandConfig{RootCommandConfig: rootConfig}
+
+	var stackCmd = &cobra.Command{
+		Use:   "create <name>",
+		Short: "Create a new stack as a copy of an existing stack",
```
helenmasters: Short: "Create a new Appsody stack."

cmd/stack_create.go
```diff
@@ -0,0 +1,180 @@
...
+		Long: `This command will create a new stack as a copy of an existing sample stack in the current directory that has the structure of an Appsody stack.
```
helenmasters: Create a new Appsody stack, called <name>, in the current directory. You can use this stack as a starting point for developing your own Appsody stack.

By default, the new stack is based on the example stack: samples/sample-stack. If you want to use a different stack as the basis for your new stack, use the copy flag to specify the stack you want to use as the starting point. You can use `appsody list` to see the available stacks.

Examples:
   `appsody stack create my-stack`
    Creates a stack called my-stack, based on the example stack “samples/sample-stack”.

   `appsody stack create my-stack --copy incubator/nodejs-express`
    Creates a stack called my-stack, based on the Node.js.Express stack.

The stack name must start with a letter or number, and can contain only lowercase letters, numbers, or dashes, and cannot end with a dash. The stack name cannot exceed 128 characters.

cmd/stack_create.go
```diff
@@ -0,0 +1,180 @@
...
+			//rename the stack to the name which user want
+			err = os.Rename(projectType, stack)
+			if err != nil {
+				return err
+			}
+
+			return nil
+		},
+	}
+	stackCmd.PersistentFlags().StringVar(&config.copy, "copy", "samples/sample-stack", "Copy existing stack")
```
helenmasters: "Copy the specified stack. The format is <repository>/<stack>"

PR #394, opened 2019-10-03
Adding machine output for list and repo list commands

### LINKED ISSUE #53: Machine readable output for list/repo list
I want to script installation of a repo, and verification of the stacks contained within.
Today, this involves passing the human formatted output from appsody list/repo list through grep/sed/cut etc to extract & identify the info by line/column/regex etc. 
This approach is obviously fragile, and less than ideal.

kubectl offers `-o json` .. I'd like to see something like this for appsody, to enable scripts to parse appsody query command output

An alternative could be to allow selection of the fields to be output, and if headers should be included etc. This approach is used for other unix commands, eg, `ps` supports `-h` to suppress headers, and other args to allow selection of columns.. This makes the output from ps easier to handle within scripts etc, at the cost of making the flags for the interface kind of complex.

**Additional context**
Without any current flags to produce this kind of output, it should be considered that the current text format is likely to be parsed by scripts, and thus changes to such should be very carefully considered due to potential impact to tooling.

EnriqueL8:
Inspired from #230 
Fixes #53 

I think the tests can be better... Let me know what you think @kylegc

### DIFF
--- a/cmd/list_test.go
+++ b/cmd/list_test.go
@@ -17,6 +17,7 @@ import (
 	"strings"
 	"testing"
 
+	cmd "github.com/appsody/appsody/cmd"
 	"github.com/appsody/appsody/cmd/cmdtest"
 )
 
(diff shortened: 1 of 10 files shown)


cmd/repo_list.go
```diff
@@ -15,10 +15,18 @@
...
 
 import (
+	"encoding/json"
+
 	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v2"
 )
 
+type repoListCommandConfig struct {
+	output string
```
kylegc: Same comment to add `*RootCommandConfig`

cmd/list_test.go
```diff
@@ -74,12 +75,6 @@ func TestListV2(t *testing.T) {
 		t.Error("list command should contain id 'java-microprofile'")
 	}
 
-	output, _ = cmdtest.RunAppsodyCmdExec([]string{"list", "appsodyhub"}, ".")
-
-	// we expect 2 instances
-	if !(strings.Count(output, "java-microprofile") == 1) {
```
kylegc: I had some issues with this test as well, and I think it needs to be reworked. However this part is not failing for me, I was getting issues with the `strings.Count(output, "nodejs ") == 2` a couple lines down. What was the reason for removing this?

cmd/repo.go
```diff
@@ -302,7 +302,7 @@ func downloadIndex(url string) (*RepoIndex, error) {
 }
 
 func (index *RepoIndex) listProjects(repoName string, config *RootCommandConfig) (string, error) {
```
kylegc: While you are changing this code, can you come up with better names for the two `listProjects` methods?

cmd/list_test.go
EnriqueL8: I had also errors with the `"nodejs "` line, but it gets fixed when you run the whole test again because the previous test adds a new repository with a `nodejs` stack.

I removed this because the same check is performed when doing the following, just above.
```go
if !strings.Contains(output, "java-microprofile") {
	t.Error("list command should contain id 'java-microprofile'")
}
```

EnriqueL8:
Thanks for the comments @kylegc! 
- I have changed the key names of JSON to be consistent with the YAML one.
- I have changed it to an array of Templates, and have added a function to convert the array to a string for printing. There is an inconsistency between JSON and YAML when outputting Boolean values using the `omitempty` flag. If a boolean is set to `False` YAML does not output it, but JSON does. This is related to adding the `default` field for the template.
- For the inconsistency between the two commands, I believe that the `repo list` command is related to precise information about the repos being used that is why the APIVersion might be of use. For `list` we just want the list of stacks available, I wonder what would be the use case to have the APIVersion there. I will change the code to add the APIVersion into the list command.

cmd/repo.go
EnriqueL8: After looking again at these function, I believe this needs more work than just renaming. They should just be extract into a common function, since they share a lot of code.

kylegc: Fair enough. Please create a separate issue to fix this :)

kylegc:
For the default field, what if you do not specify `omitempty`? Will it be displayed as false then?

EnriqueL8:
> For the default field, what if you do not specify `omitempty`? Will it be displayed as false then?

Yes

cmd/repo.go
EnriqueL8: Just seen this! Yes, I'll create an issue to maybe refactor `repo.go`?

PR #414, opened 2019-10-07
Git utilities

kewegner:
the function GetGitInfo returns a struct with the following:

repo URL
branch
Boolean whether changes have been made

Last commit Info:
 commit long SHA
 commit author
 commit datetime
 commit URL

This code has only been tested in a local workspace environment, and would need to be tested in the pipeline environments, and with a wider set of git project environments.

### DIFF
--- a/cmd/git_utils.go
+++ b/cmd/git_utils.go
@@ -151,7 +151,7 @@ func GetGitInfo(dryrun bool) (GitInfo, error) {
 	gitInfo.Commit, gitErr = RunGitGetLastCommit(gitInfo.RemoteURL, dryrun)
 	if gitErr != nil {
 		Info.log("Received error getting current commit: ", gitErr)
-		//return gitInfo, gitErr
+
 	}
 
 	return gitInfo, nil
--- a/cmd/git_utils.go
+++ b/cmd/git_utils.go
@@ -177,7 +177,10 @@ func RunGitGetLastCommit(URL string, dryrun bool) (CommitInfo, error) {
 	if gitErr != nil {
 		return commitInfo, gitErr
 	}
-	json.Unmarshal([]byte(strings.Trim(commitStringInfo, trimChars)), &commitInfo)
+	err := json.Unmarshal([]byte(strings.Trim(commitStringInfo, trimChars)), &commitInfo)
+	if err != nil {
+		return commitInfo, errors.Errorf("JSON Unmarshall error: %v", err)
+	}
 	if URL != "" {
 		commitInfo.URL = stringBefore(URL, ".git") + "/commit/" + commitInfo.SHA
 	}
--- a/cmd/git_utils.go
+++ b/cmd/git_utils.go
@@ -30,10 +30,10 @@ type CommitInfo struct {
 }
 
 type GitInfo struct {
-	Branch      string
-	Upstream    string
-	RemoteURL   string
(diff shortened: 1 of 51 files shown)


cmd/git_utils.go
```diff
@@ -0,0 +1,215 @@
...
+		}
+
+	} else {
+		Info.log("Unable to determine origin to compute repository URL")
+	}
+
+	gitInfo.Commit, gitErr = RunGitGetLastCommit(gitInfo.RemoteURL, dryrun)
+	if gitErr != nil {
+		Info.log("Received error getting current commit: ", gitErr)
+		//return gitInfo, gitErr
```
EnriqueL8: Should we return an error, or return the `gitInfo` struct without the current commit? I am not sure

cmd/git_utils.go
```diff
@@ -0,0 +1,215 @@
...
+	lineSeparator := "\n"
+	if runtime.GOOS == "windows" {
+		lineSeparator = "\r\n"
+	}
+	outputLines := strings.Split(output, lineSeparator)
+
+	var changeLines []string
+	const noCommits = "## No commits yet on "
+	const branchPrefix = "## "
+	const branchSeparatorString = "..."
```
EnriqueL8: I tested this command locally, and did not get a `branchSeparatorString` after my branch name. 
Instead I get just 
image
So the branch field in GitInfo is not set correctly. Is the separator needed when you have multiple branches?

(Note: git.go is the local version of this file)

cmd/git_utils.go
kewegner: The thought here was not to error out, if we can't find the commit.

cmd/git_utils.go
```diff
@@ -0,0 +1,215 @@
...
+	for _, value := range outputLines {
```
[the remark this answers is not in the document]
kewegner: Kyle and Neeraj had said we only needed a boolean whether there were changes, so I am going to get rid of the changes list for the moment.

cmd/git_utils.go
```diff
@@ -0,0 +1,215 @@
...
+
+			}
+			// this must be a change line
+			if !branchFound {
+				changeLines = append(changeLines, value)
+			}
+		}
+
+	}
+	gitInfo.Changes = append(gitInfo.Changes, changeLines...)
```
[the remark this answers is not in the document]
kewegner: See above.

cmd/git_utils.go
```diff
@@ -0,0 +1,215 @@
...
+	gitInfo.Commit, gitErr = RunGitGetLastCommit(gitInfo.RemoteURL, dryrun)
+	if gitErr != nil {
+		Info.log("Received error getting current commit: ", gitErr)
+		//return gitInfo, gitErr
+	}
+	changesMade := false
+	if len(changeLines) != 0 {
+		changesMade = true
+	}
+	gitInfo.ChangesMade = changesMade
```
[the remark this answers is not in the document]
kewegner: logic will be change per above.

cmd/git_utils.go
kewegner: what do you get when you do git remote -v out of curiosity?  What about git config?

kewegner: It's ok I can reproduce it.

PR #440, opened 2019-10-11
Add wait for test test to complete

kylegc:
Previously this test was not waiting for the command to finish and checking for errors, it was always immediately successfully passing the test.

### DIFF
--- a/functest/test_test.go
+++ b/functest/test_test.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/appsody/appsody/cmd/cmdtest"
 )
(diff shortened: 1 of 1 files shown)


functest/test_test.go
```diff
@@ -61,26 +62,46 @@ func TestTestSimple(t *testing.T) {
...
 
-		// stop and cleanup
-		func() {
-			_, err = cmdtest.RunAppsodyCmdExec([]string{"stop"}, projectDir)
-			if err != nil {
-				fmt.Printf("Ignoring error running appsody stop: %s", err)
+		waitForError := 20 // in seconds
+		stillWaiting := true
+		log.Println("Waiting to see if test will fail...")
+		for stillWaiting {
```
EnriqueL8: Just as a suggestion, we could remove this bool, and just have `for`, then we could use `break` instead of assigning the bool to `false` to exit the "for" loop.

kylegc: I tried your suggestion, but the` break`s applied to the `select` block instead of the `for` loop.

Issue #284, opened 2019-09-09
proposal: appsody describe


gireeshpunathil:
**Is your feature request related to a problem? Please describe.**

This is an RFE to implement a feature `appsody describe` as part of appsody `cmd` that dumps the state of the project, as an integrated view. More specifically:
**Describe the solution you'd like**

Local view:
 - project configuration
   - project assets
   - stack image name
 - cluster configuration
   - cluster name
   - api end point name
   - namespace
   - cluster user

Image view:
 - mount bindings
 - project env
 - working directory
 - unix user and group name
 - server entrypoint

Remote view (if deployed):
 - pods view
 - deployments view
 - routes view
 - service endpoint url
 - scale view
 - resource consumption view

**Describe alternatives you've considered**
None

**Additional context**
Myself and / or @pushkarnk plan to work on this, but please let me know WDYT.

seabaylea:
This aligns with some work that we're starting to look at for increased visibility of what is deployed into Kubernetes via the AppsodyOperator - and we'll absolutely want to support visualising that data from the Appsody CLI.

It also makes sense, as proposed, to give as similar a view as possible for local projects/apps (ie, extending what you get today from `appsody ps`.

pushkarnk:
I am giving this a shot. 

Assuming that Jane would issue the `appsody describe` command from within an appsody application project directory (no other location makes sense), it is first essential to infer two things: whether the app has been deployed and, if it is deployed, what is the name of the corresponding `AppsodyApplication`. AFAIK, there is no formal way to get this done.

It would be good to have a unique identifier for each appsody application project, which is passed on the the image and container as a label. This will help us map an `AppsodyApplication` instance to a project directory.

Currently, the name of the image/AppsodyApplication/pod is derived from the name of the project directory by prefixing the latter . However this means` /root/projects/foo/ms1` and `/root/projects/bar/ms1` will have the same prefix of `ms1` and it is not possible to do a project -> pod mapping.

Solutions I can think of:
1. Generate a unique identifier for each appsody project and include it as a label in the image (addition to https://github.com/appsody/appsody/pull/453)
2. Add a label with the path of the project directory: `path=/root/projects/foo/ms1`

gireeshpunathil:
> Generate a unique identifier for each appsody project and include it as a label in the image 

I guess this warrants that we also store this locally so that the CLI can read it

So I support option 2, in which the CLI can easily obtain the path without additional recording.

kylegc:
[...]

PR #453, opened 2019-10-14
Adding Labels to final 

EnriqueL8:
Adding labels to final docker image

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -78,8 +78,9 @@ func build(config *buildCommandConfig) error {
 
 	projectName, perr := getProjectName(config.RootCommandConfig)
 	if perr != nil {
-		return errors.Errorf("%v", perr)
+		return perr
 	}
+
 	extractDir := filepath.Join(getHome(config.RootCommandConfig), "extract", projectName)
 	dockerfile := filepath.Join(extractDir, "Dockerfile")
 	buildImage := projectName //Lowercased
@@ -103,8 +104,10 @@ func build(config *buildCommandConfig) error {
 
 	labels, err := getLabels(config)
 	if err != nil {
-		return errors.Errorf("%v", err)
+		return err
 	}
+
+	// It would be nicer to only call the --label flag once. Could also use the --label-file flag.
 	for _, label := range labels {
 		cmdArgs = append(cmdArgs, "--label", label)
 	}
@@ -125,7 +128,7 @@ func build(config *buildCommandConfig) error {
 func getLabels(config *buildCommandConfig) ([]string, error) {
 	var labels []string
 
-	dockerLabels, err := getDockerLabels(config.RootCommandConfig)
+	dockerLabels, err := getStackLabels(config.RootCommandConfig)
 	if err != nil {
 		return labels, err
 	}
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -128,7 +128,7 @@ func build(config *buildCommandConfig) error {
 func getLabels(config *buildCommandConfig) ([]string, error) {
 	var labels []string
 
-	dockerLabels, err := getStackLabels(config.RootCommandConfig)
+	stackLabels, err := getStackLabels(config.RootCommandConfig)
 	if err != nil {
 		return labels, err
 	}
@@ -143,8 +143,15 @@ func getLabels(config *buildCommandConfig) ([]string, error) {
 		Warning.log(err)
 	}
 
-	for key, value := range dockerLabels {
+	for key, value := range stackLabels {
 		delete(configLabels, key)
+
+		key = strings.Replace(key, "org.opencontainers.image", "dev.appsody.stack", -1)
+
+		// This is temporarily until we update the labels in stack dockerfile
+		if key == "appsody.stack" {
+			key = "dev.appsody.stack.id"
+		}
 		labelString := fmt.Sprintf("%s=%s", key, value)
 		labels = append(labels, labelString)
 	}
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -15,6 +15,7 @@
 package cmd
 
 import (
+	"fmt"
 	"path/filepath"
 	"regexp"
 	"strings"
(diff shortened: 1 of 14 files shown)


cmd/build.go
```diff
@@ -98,8 +99,16 @@ func build(config *buildCommandConfig) error {
 			return err
 		}
 		cmdArgs = append(cmdArgs, options...)
+	}
 
+	labels, err := getLabels(config)
+	if err != nil {
+		return errors.Errorf("%v", err)
```
kylegc: Is there a reason why you are wrapping the `err` in another error without changing the message?

cmd/root.go
```diff
@@ -48,9 +48,10 @@ type RootCommandConfig struct {
 	UnsupportedRepos []string
 
 	// package scoped, these are mostly for caching
-	setupConfigRun bool
-	imagePulled    map[string]bool
-	cachedEnvVars  map[string]string
+	setupConfigRun     bool
+	imagePulled        map[string]bool
+	cachedEnvVars      map[string]string
+	cachedDockerLabels map[string]string
```
kylegc: Maybe we should call these `cachedStackLabels`? The term Docker is too ambiguous

cmd/utils.go
```diff
@@ -437,50 +437,161 @@ func UserHomeDir() string {
...
+	}
+
+	if commitInfo.URL != "" {
+		labels[commitString+"url"] = commitInfo.URL
+	}
+
+	return labels, nil
+}
+
+func getDockerLabels(config *RootCommandConfig) (map[string]string, error) {
```
kylegc: Similar to above, I'm thinking a better name would be `getStackLabels`

cmd/utils.go
```diff
@@ -437,50 +437,161 @@ func UserHomeDir() string {
...
+
+func getGitLabels(config *RootCommandConfig) (map[string]string, error) {
+	gitInfo, err := GetGitInfo(config.Dryrun)
+	if err != nil {
+		return nil, err
+	}
+
+	var labels = make(map[string]string)
+
+	gitString := "git."
```
kylegc: These aren't going to all align, but lets use the OpenContainer annotations as much as possible.
https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys

EnriqueL8: Do you mean using OpenContainer annotations for the git information? I haven't seen anything on that... Or do you mean adding OCI annotations for the final image being built?

pushkarnk:
@EnriqueL8 Thanks for this PR. It is going to help us for a plausible implementation of #284. 

I had a question in that context: does you PR address [this](https://github.com/appsody/appsody/issues/284#issuecomment-542580724) issue? I took a glance at your changes and I guess the answer is no?

EnriqueL8:
I don't think it addresses the issue, but it might help you for your implementation.

cmd/utils.go
```diff
@@ -284,19 +295,24 @@ func getProjectConfig(config *RootCommandConfig) (ProjectConfig, error) {
...
+		Debug.log("Project stack from config file: ", projectConfig.Stack)
 		imageRepo := config.CliConfig.GetString("images")
 		Debug.log("Image repository set to: ", imageRepo)
 		if imageRepo != "index.docker.io" {
-			stack = imageRepo + "/" + stack
+			projectConfig.Stack = imageRepo + "/" + projectConfig.Stack
 		}
-		Debug.log("Pulling stack image as: ", stack)
-		config.ProjectConfig = &ProjectConfig{stack}
+		Debug.log("Pulling stack image as: ", projectConfig.Stack)
```
kylegc: I see the message was there before, but it doesn't look like we are pulling the stack image yet. I think we sould delete this log message as we print out the stack in several other debug logs.

cmd/utils.go
```diff
@@ -445,54 +461,69 @@ func getConfigLabels(config *RootCommandConfig) (map[string]string, error) {
...
-}
+	labels[ociKeyPrefix+"created"] = t.Format(time.RFC3339)
 
-func getGitLabels(config *RootCommandConfig) (map[string]string, error) {
-	gitInfo, err := GetGitInfo(config.Dryrun)
-	if err != nil {
-		return nil, err
+	var maintainersString string
+	for index, maintainer := range projectConfig.Maintainers {
+		maintainersString += maintainer.Name + ", " + maintainer.Email + ", " + maintainer.GithubID
```
kylegc: lets leave off the GithubID for now. just the maintainer name and email should be enough

kylegc: Also I think we should put the email in parenthesis and use commas only to separate multiple maintainers. 
For example:
`Neeraj Laad (neeraj.laad@gmail.com)`
and for multiple:
`Neeraj Laad (neeraj.laad@gmail.com), Kyle Christianson (christik@us.ibm.com)`

cmd/utils.go
```diff
@@ -445,54 +461,69 @@ func getConfigLabels(config *RootCommandConfig) (map[string]string, error) {
...
 	}
 
-	if commitInfo.URL != "" {
-		labels[commitString+"url"] = commitInfo.URL
+	var commitInfo = gitInfo.Commit
+	revisionKey := appsodyKeyPrefix + "revision"
+	if commitInfo.SHA != "" {
+		labels[revisionKey] = commitInfo.SHA
+		if gitInfo.ChangesMade {
+			labels[revisionKey] += "-changesMade:true"
```
kylegc: I'm not necessarily opposed to this, but I'm wondering your reason for using "-changesMade:true" instead of "-modified" as discussed?

cmd/utils.go
```diff
@@ -545,6 +576,12 @@ func getStackLabels(config *RootCommandConfig) (map[string]string, error) {
 			labelsMap := containerConfig["Labels"].(map[string]interface{})
 
 			for key, value := range labelsMap {
+				key = strings.Replace(key, "org.opencontainers.image", "dev.appsody.stack", -1)
```
kylegc: I think this transformation should happen when you are applying the `appsody build` labels rather than in the `getStackLabels()` func. Because other users of this func may expect the "org.opencontainers...." labels on the stack.

cmd/utils.go
EnriqueL8: Yeah, I prefer that - Good idea!

cmd/utils.go
EnriqueL8: I agree

cmd/utils.go
EnriqueL8: True - have changed it

Issue #447, opened 2019-10-11
Appsody on K8S should minimize the wait time for pulling logs from the running pods


chilanti:
**Is your feature request related to a problem? Please describe.**
Appsody on K8S needs to stream the console/log messages from the running apps to the pod where Appsody is running. However, the target pod may be slow to come up, causing the `kubectl logs` command to fail. The command should be retried often (active polling) and should start showing the logs as soon as practically possible.

**Describe the solution you'd like**
After `appsody run` has started the target deployment, the logs should appear on the Appsody container as soon as possible (rather than after a pre-determined long wait).

PR #457, opened 2019-10-15
Codewind logging phase1

kewegner:
Partial fix of #447
This will diagnose kubectl or wait error resulting from kubectl logs command.


cmd/dev_common.go
```diff
@@ -353,41 +354,47 @@ func commonCmd(config *devCommonConfig, mode string) error {
 			}
 		}
 		Info.log("Waiting 30 seconds for the container to start")
```
chilanti: Let's remember to remove this

PR #477, opened 2019-10-18
Refactor project name

### LINKED ISSUE #287: Add --project-name option to init, add the name to .appsody-config.yaml, and validate it against k8s rules

kylegc:
Fixes #287 

Removed projectName from RootCommandConfig and use ProjectConfig.ProjectName instead
Removed setProjectName func and instead created a simpler saveProjectNameToConfig func
Simplify getProjectName

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -265,7 +265,10 @@ func install(config *initCommandConfig) error {
 	if configErr != nil {
 		return configErr
 	}
-	if projectConfig.ProjectName == "" {
+
+	// save the project name to .appsody-config.yaml only if it doesn't already exist there
+	// or if the user specified --project-name on the command line
+	if projectConfig.ProjectName == "" || config.projectName != defaultProjectName(config.RootCommandConfig) {
 		err := saveProjectNameToConfig(config.projectName, config.RootCommandConfig)
 		if err != nil {
 			return err
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -85,6 +85,10 @@ func initAppsody(stack string, template string, config *initCommandConfig) error
 	if noTemplate {
 		Warning.log("The --no-template flag has been deprecated.  Please specify a template value of \"none\" instead.")
 	}
+	valid, err := IsValidProjectName(config.projectName)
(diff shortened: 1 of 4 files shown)


Bazif-Khan:
Overall PR looks good to me; however, when you initialize the project with wrong project-name flag value e.g. H. Then project-name will not be saved in the .appsody-config.yaml. Then if you run appsody deploy then project will be deployed successfully but it's printing project name many times on CLI because it's going in saveProjectNameToConfig() function every time.

------

Screen Shot 2019-10-18 at 14 14 51

PR #490, opened 2019-10-21
changes for setting pull policy for local stacks

tnixa:


cmd/repo_add.go
```diff
@@ -42,9 +42,9 @@ func newRepoAddCmd(config *RootCommandConfig) *cobra.Command {
 				return errors.Errorf("Invalid repository name. The <name> must be less than 50 characters")
 
 			}
-			match, _ := regexp.MatchString("^[a-zA-Z0-9\\-_]{1,50}$", repoName)
+			match, _ := regexp.MatchString("^[a-zA-Z0-9\\-_\\.]{1,50}$", repoName)
 			if !match {
-				return errors.Errorf("Invalid repository name. The <name> may only contain digits, numbers, dashes '-', and underscores '_'.")
+				return errors.Errorf("Invalid repository name. The <name> may only contain digits, numbers, dashes '-', underscores '_' and dots '.'.")
```
helenmasters: The word 'may' expresses possibility, so the error message would be better as:

"The name can contain only letters (lowercase or uppercase), numbers, dashes '-', underscores '_', and periods '.'"

Assuming that you meant "letters, numbers" and not "digits, numbers"?

helenmasters:
@tnixa apologies, I didn't make it clear. Please can you put back the first sentence in the error message? I only meant you to fix the second sentence.  It should end up as...

"Invalid repository name. The name can contain only letters (lowercase or uppercase), numbers, dashes '-', underscores '_', and periods '.'"

Issue #462, opened 2019-10-15
kubectl apply command fails but no error message given


tnixa:
**Describe the bug**
`appsody deploy` fails during the `kubectl apply` command but only gives the exit status. We should try to provide addition error messages.

**To Reproduce**
1. `mkdir nodejs`
2. `cd nodejs`
3. `appsody init nodejs --project-name nodejs_project-test`
4. `appsody deploy`

**Expected behavior**
In `utils.go` if we change the `execCmd.Output()` to `execCmd.CombinedOutput()` we get the exit status and the error message...
```
Info] Running command: kubectl apply -f app-deploy.yaml --namespace default
[Error] kubectl apply failed: exit status 1 The AppsodyApplication "nodejs_test-project" is invalid: metadata.name: Invalid value: "nodejs_test-project": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')

[Error] Failed to deploy to your Kubernetes cluster: exit status 1
```

**Actual behaviour**
As is we only get the exit status:
```
[Info] Running command: kubectl apply -f app-deploy.yaml --namespace default
[Error] kubectl apply failed: exit status 1
[Error] Failed to deploy to your Kubernetes cluster: exit status 1

Failed to deploy to your Kubernetes cluster: exit status 1
github.com/appsody/appsody/cmd.newDeployCmd.func1
        /Users/tnixa/go/src/github.com/appsody/appsody/cmd/deploy.go:216
github.com/appsody/appsody/vendor/github.com/spf13/cobra.(*Command).execute
        /Users/tnixa/go/src/github.com/appsody/appsody/vendor/github.com/spf13/cobra/command.go:762
github.com/appsody/appsody/vendor/github.com/spf13/cobra.(*Command).ExecuteC
        /Users/tnixa/go/src/github.com/appsody/appsody/vendor/github.com/spf13/cobra/command.go:852
github.com/appsody/appsody/vendor/github.com/spf13/cobra.(*Command).Execute
        /Users/tnixa/go/src/github.com/appsody/appsody/vendor/github.com/spf13/cobra/command.go:800
github.com/appsody/appsody/cmd.ExecuteE
        /Users/tnixa/go/src/github.com/appsody/appsody/cmd/root.go:197
[...]

PR #507, opened 2019-10-23
Standardised most of the error propagation in utils.go

upLukeWeston:
Corrected instances where only stdErr was being propagated as an exit message to instead return stdOut along with stdErr.

Fixes: #462

### DIFF
--- a/cmd/docker_commands.go
+++ b/cmd/docker_commands.go
@@ -6,6 +6,8 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+
+	"github.com/pkg/errors"
 )
 
 func DockerRunAndListen(args []string, logger appsodylogger, interactive bool, verbose bool, dryrun bool) (*exec.Cmd, error) {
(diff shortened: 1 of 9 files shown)


kylegc:
Take a look at the code snippet in issue #462. The documentation says stderr should be part of the error, but we must have something wrong with that code because it didn't seem to work. Maybe you can figure that out?

kylegc:
Hi Luke, you are on the right track with your test, however, the commands you chose are not available on the travis test system so it failed. For this unit test, I suggest you run commands like `ls` and/or `cp`.

You could also write some functests to call the `docker` commands. (docker should be available when the functests run). I'd suggest a new docker_commands_test.go file in functest to call some of the methods directly like `RunDockerInspect` and check the error output.

cmd/docker_commands.go
```diff
@@ -98,10 +100,10 @@ func RunCommandAndListen(commandValue string, args []string, logger appsodylogge
 			}
 		}()
 
-		err = execCmd.Start()
+		out, err := SeperateOutput(execCmd)
```
kylegc: Hey Luke. The new tests look great. This change however, I don't think should be made. The `cmd.Start()` func behaves different from the `cmd.Output()` which you are replacing. Lets leave `Start()`'s alone for now. This may be the reason your travis build is failing.

upLukeWeston: I did think this would be the case but didn't want to write it off immediately! I'll remove the change & the corresponding test and leave just the DockerInspect tests.

kylegc: Yay travis passed! :-) So were you able to test this with any of the kubectl functions? Reason I ask is because this issue was originally reported against a kubectl apply command. It would be good to verify the original issue is fixed.

upLukeWeston: Yes I tried manually with all of the Kube functions I could. I'll add an automated test to verify this, though!

Issue #428, opened 2019-10-10
`appsody init <stack> none` downloads templates


knolleary:
**Describe the bug**
Running `appsody init <stack> none` still appears to download and unpack templates even though they won't be used.

The first time I saw this in the log, I thought the `none` option had been ignored and a template had been applied. It left me uncertain as to what had happened. 

**To Reproduce**

1. Run `appsody init nodejs-express none`
```
$ appsody init nodejs-express none
Running appsody init...
Downloading nodejs-express template project from https://github.com/appsody/stacks/releases/download/nodejs-express-v0.2.7/incubator.nodejs-express.v0.2.7.templates.simple.tar.gz
Download complete. Extracting files from nodejs-express.tar.gz
...
Successfully initialized Appsody project
```

**Expected behavior**
I would not expect the tool to be downloading and extracting templates when I have explicitly said I don't want a template to be used.

The success message at the end should also reaffirm what template, if any, has been applied.

**Environment Details (please complete the following information):**
 - OS: OSX
 - appsody 0.4.6

kylegc:
We do need to download and extract the template just to get the `.appsody-config.yaml` file out of it. 

We will use this issue to improve on the log messages. Specifically stating that `none` was not ignored, that we are only extracting the config file, and that the successful message did or did not apply a template.

knolleary:
I asked in slack a couple days ago and was told the template didn't need to provide `.appsody-config.yaml`.

https://appsody.slack.com/archives/CKZ50HJET/p1570636324003600

> The .appsody-config.yaml is not part of the the source structure, but will be generated as part of the stack building process, and will be placed in the user directory by the appsody init, command. This file specifies the stack image that will be used, and can be overridden for testing purposes to a locally built stack.

kylegc:
Yes, that's all correct. The `.appsody-config.yaml` file is not part of the stack source code, it is generated at stack build time and baked into the template tar files. Then the CLI downloads the tar file and extracts `.appsody-config.yaml` at init time.

I'm not saying this is the best way to do things, that's just how it is today. There may be changes to this process in the future.

PR #514, opened 2019-10-25
init: improve log message when template is none

skoh7645:
Improve log message when user specifies not to download a stack template on `appsody init`

Fixes: #428


kylegc:
Thanks for the fix Sandy. The original issue also says 
"The success message at the end should also reaffirm what template, if any, has been applied."
I think this is a good idea. Can you add an alternative success message too?
Thanks

cmd/init.go
```diff
@@ -241,7 +246,8 @@ func initAppsody(stack string, template string, config *initCommandConfig) error
 			Info.log("If you wish to proceed and overwrite files in the current directory, try again with the --overwrite option.")
 			// this leave the tar file in the dir
 			return errors.Errorf("Error extracting project template: %v", errUntar)
-
+		} else if inputTemplateName != "none" {
```
kylegc: Hi Sandy, sorry for the delay in reviews here.

I don't think this is what the issue intended. I think they wish to alter the existing "Successfully initialized Appsody project" message a few lines down. Change that to be something like:
"Successfully initialized Appsody project with the nodejs-express stack and the default template" or
"Successfully initialized Appsody project with the nodejs-express stack and no template"

PR #521, opened 2019-10-25
Change repo name appsodyhub to incubator

kylegc:


EnriqueL8:
Would the `repository.yaml` files under `testdata/` also need updating?

kylegc:
I've added automatic user migration, and updated the testdata (thanks Enrique).

PR #528, opened 2019-10-28
Added Go templating for accessing metadata values in stacks

CameronMcWilliam:
# Appsody templating - **NOW OBSOLETE METHOD**
This is a first draft PR, we may decide this is not the way to go ***or*** that there are better alternative methods.  Also I have not written in Go before so any feedback on general code practices would be much obliged too.

## Method

* Takes place in `appsody stack package` command
* Use built in Go `text/template` package
* Make copy of stack directory
* Pull values (***currently only `name` and `version` will be pulled***) from `stack.yaml` file into a `Metadata` object
* `filepath.Walk` through copied directory
* Check if file is read only, if read only, set to write only temporarily
* Create new template from file by using `ParseFiles` on file
* Open target copied file and `ExecuteTemplate` using `metadata` object
* Close file and if originally read only, set back to read only
* After docker build, delete copied folder leading non templated original folder behind

## Further details

* Need to discuss exactly what metadata we want.
* If a user wants to use templating they can wrap their variable as so:
```
{{"{{.UserVariableName}}"}}
```
> This could be avoided by using custom Delimiters

* Current method applies template to entire project, could make stack creator declare a pre defined array of files that need templating as to avoid any unwanted templating.  This however, means more work for stack creator.
* Unsure on behaviour on Windows currently, there are Windows only workarounds in place but they may need to be fixed.


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "vendor/github.com/hectane/go-acl"]
```
kylegc: Can this file be removed now?

kylegc:
@CameronMcWilliam Can you talk a little more about why you check for readonly and modify the permissions? I do believe we need to preserve the stack's original file permissions, but if you are initially copying the entire dir, doesn't that mean you already have write permission?

.gitmodules
CameronMcWilliam: Yes, I will remove now

CameronMcWilliam:
> @CameronMcWilliam Can you talk a little more about why you check for readonly and modify the permissions? I do believe we need to preserve the stack's original file permissions, but if you are initially copying the entire dir, doesn't that mean you already have write permission?

I was under the impression that the copydir folder preserved file permissions, It appears I was wrong, there's an easy fix to this that I'll get to working on now.

kylegc:
The `copyDir()` func uses the `cp` command. Maybe we need to add the `cp -p` flag to that command?

CameronMcWilliam:
> The `copyDir()` func uses the `cp` command. Maybe we need to add the `cp -p` flag to that command?

We could... but it's not necessary as far as I'm aware, because I check the file permission of the old non-templated path to the new templated path and apply it there.  It actually makes it easier with just the cp command as it means all files are writeable so I can apply templates to it and ***then*** it changes it back at the end to its old file permissions.  Unless the -p flag does more than just permissions and I'm missing something...

CameronMcWilliam:
# Appsody templating

**UPDATED VERSION**

## Method

* Takes place in `appsody stack package` command
* Use built in Go `text/template` package
* Make copy of stack directory
* Pull these values from `stack.yaml` file into a `Metadata` object
* `filepath.Walk` through copied directory
* Check if file is read only, if read only, set to write only temporarily
* Create new template from file by using `ParseFiles` on file
* Open target copied file and `ExecuteTemplate` using `metadata` object
* Close file and if originally read only, set back to read only
* After docker build, delete copied folder leading non templated original folder behind

## Use cases 

* To use a variable within the project wrap the Variable with `{{ }}` and put a period before the variable. For example, if we wanted to get the Stack Name we would use:
```
This is the {{.StackName}} stack.
```
* To use the custom variables, the stack creator will create (if it does not already exist) a `custom-variables` map in their `stack.yaml` and create as many variables as they feel necessary.  Example `stack.yaml` file:
```
name: Starter Sample
version: 0.1.1
description: Runnable starter stack, copy to create a new stack
license: Apache-2.0
language: bash
maintainers: 
  - name: Henry Nash
    email: henry.nash@uk.ibm.com
    github-id: henrynash
default-template: simple
custom-variables:
  Variable1: value1
  Variable2: value2
  Variable3: value3
```
To then use these custom variables:
```
This is {{.StackCustomTemplate.Variable1}}, this is {{.StackCustomTemplate.Variable2}} and this is {{.StackCustomTemplate.Variable3}}
```
* The variable names used are as follows:
    - `StackID` (the stack name from the stack path)
    - `StackName` (`name` value from `stack.yaml`)
    - `StackMajorVersion` (`name` major value from `stack.yaml`)
    - `StackMinorVersion` (`version` minor value from `stack.yaml`)
    - `StackPatchVersion` (`name` patch value from `stack.yaml`)
    - `StackImageNamespace` (`image-namespace` from user defined image-namespace flag, default is `dev.local`)
    - `StackCustomTemplate` (`custom variables` map values from `stack.yaml`)
* If a user wants to use templating they can wrap their variable as so:
```
{{"{{.UserVariableName}}"}}
```
* Current method applies template to entire project, image and template.
* To preserve file permissions on Windows the stack creator must use RUN `chmod` from the Dockerfile-stack

CameronMcWilliam:
# Appsody templating

**UPDATED VERSION**

## Method

* Takes place in `appsody stack package` command
* Use built in Go `text/template` package
* Make copy of stack directory
* Pull these values from `stack.yaml` file into a `templateMetada` map
* `filepath.Walk` through copied directory
* Check if file is read only, if read only, set to write only temporarily
* Create new template from file by using `ParseFiles` on file
* Open target copied file and `ExecuteTemplate` using `metadata` object
* Close file and if originally read only, set back to read only
* After docker build, delete copied folder leading non templated original folder behind

## Use cases 

* To use a variable within the project wrap the key with `{{  }}`. For example, if we wanted to get the stack name we would use:
```
This is the {{.stack.name}} stack.
```
* To use the custom variables, the stack creator will create (if it does not already exist) a `templating-data` map in their `stack.yaml` and create as many key and value pairs as they feel necessary.  Example `stack.yaml` file:
```
name: Starter Sample
version: 0.1.1
description: Runnable starter stack, copy to create a new stack
license: Apache-2.0
language: bash
maintainers: 
  - name: Henry Nash
    email: henry.nash@uk.ibm.com
    github-id: henrynash
default-template: simple
templating-data:
  key1: value1
  key2: value2
  key3: value3
```
To then use these custom variables:
```
This is {{.stack.key1}}, this is {{.stack.key2}} and this is {{.stack.key3}}
```
* The key names used are as follows:
    - `.stack.id` (the stack name from the stack path)
    - `.stack.name` (`name` value from `stack.yaml`)
    - `.stack.description` (`description` value from `stack.yaml`)
    - `.stack.created` (`timestamp` of when stack package command generated stack)
    - `.stack.version.major` (`version` major value from `stack.yaml`)
    - `.stack.version.minor` (`version` minor value from `stack.yaml`)
    - `.stack.version.patch` (`version` patch value from `stack.yaml`)
    - `.stack.version.full` (`version` value from `stack.yaml`)
    - `.stack.image.namespace` (`image-namespace` from user defined image-namespace flag, default is `dev.local`)

* If a user doesn't want to use this templating method they can wrap their variable as so:
```
{{"{{ }}"}}
```
* Current method applies template to entire project, image and template.
* To preserve file permissions on Windows the stack creator must use RUN `chmod` from the Dockerfile-stack on files that need certain permissions

cmd/stack_package.go
```diff
@@ -192,8 +304,13 @@ func newStackPackageCmd(rootConfig *RootCommandConfig) *cobra.Command {
 			if err != nil {
 				return err
 			}
+
+			// overriding time label with stack package currentTime generated earlier
+			labels["org.opencontainers.image.created"] = currentTime
```
kylegc: You should use `labels[ociKeyPrefix+"created"]` here.

There's actually quite a lot of overlap in the labels and the template variables. I'd like you to think about how to reuse more of the code. Perhaps you can get the labels first and then get the template variables from the labels? Or maybe encapsulate all the stack metadata into a single object/type and have methods to get label maps or template maps.

cmd/stack_package.go
```diff
@@ -88,9 +92,134 @@ func newStackPackageCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
 			Info.Log("******************************************")
 
-			stackPath := rootConfig.ProjectDir
+			// get current time
+			currentTime := time.Now().Format(time.RFC3339)
+
+			projectPath := rootConfig.ProjectDir
+
+			// make a copy of the folder to apply template to
+			err := copyDir(projectPath, projectPath+"packagecopy")
```
kylegc: Had you considered putting the templating code in one or more functions? This function is very big already, I'd like to see it broken down.

cmd/stack_package.go
```diff
@@ -88,9 +91,43 @@ func newStackPackageCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
+
+			// remove copied folder locally, no matter the output
+			defer os.RemoveAll(projectPath + "packagecopy")
+
+			// sets stack path to be the copied folder
+			stackPath := projectPath + "packagecopy"
 			Debug.Log("stackPath is: ", stackPath)
 
+			// get the stack name from the stack path (removes copy suffix)
+			stackID := filepath.Base(strings.Replace(stackPath, "packagecopy", "", 1))
```
kylegc: You have the `projectPath` variable above, use that here instead of the strings.Replace()

cmd/stack_package.go
```diff
@@ -88,9 +91,43 @@ func newStackPackageCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
+			if err != nil {
+				os.RemoveAll(projectPath + "packagecopy")
+				return errors.Errorf("Error trying to copy directory: %v", err)
+			}
+
+			// remove copied folder locally, no matter the output
+			defer os.RemoveAll(projectPath + "packagecopy")
+
+			// sets stack path to be the copied folder
+			stackPath := projectPath + "packagecopy"
```
kylegc: This is okay, but there are a lot of other places in the code which hard code `projectPath + "packagecopy"`. Those should all use this variable instead.

cmd/stack_package.go
kylegc: Thanks for doing this!

kylegc: Thanks for splitting up the functions! I think it looks much better :)

cmd/stack_package.go
kylegc: Also I really like this better. Thanks!

CameronMcWilliam:
> A few more comments.
> Also, I see you added testdata, but I don't see any automated tests that use it. Is this intentional?

Sorry haven't got round to testing yet.  Will be looking at that after completing docs and changes to this.

cmd/stack_package.go
```diff
@@ -399,9 +429,121 @@ func getLabelsForStackImage(stackID string, buildImage string, stackYaml StackYa
...
+	return templateMetadata
+
+}
+
+// applyTemplating walks through the copied folder directory and applies a template using the
+// previously created templateMetada to all files in the target directory
+func applyTemplating(projectPath string, templateMetadata interface{}) error {
+
+	// sets stack path to be the copied folder
+	stackPath := projectPath + "stackPath"
```
kylegc: Good job getting rid of the `projectPath + "stackPath"` everywhere else, but you are still left with this one. I think it would be best to pass this in as a func parameter. The point is, if we ever change the location, it only has to be changed once in the code.

cmd/stack_package.go
```diff
@@ -88,9 +91,43 @@ func newStackPackageCmd(rootConfig *RootCommandConfig) *cobra.Command {
 			Info.Log("Running appsody stack package")
 			Info.Log("******************************************")
 
-			stackPath := rootConfig.ProjectDir
+			projectPath := rootConfig.ProjectDir
+
+			// sets stack path to be the copied folder
+			stackPath := projectPath + "stackPath"
```
kylegc: This is a temporary directory to apply the templating and run the package. I don't think we should be creating it in the user's projectPath. First, it could conflict with user's real files. Second, if something goes wrong and files are left there, the user may not like this and it may show up in git if they track this dir.

A better location would either be in `/tmp` or `~/.appsody`. We've used the appsody dir for other temp artifacts like for extract and logs, so I think this makes most sense. Here's what I'm thinking:
`stackPath := filepath.Join(getHome(rootConfig), "stacks", "packaging-" + stackID)`

CameronMcWilliam: Agreed, this is a much better approach, @neeraj-laad and I discussed doing this last Friday, I'll implement this today

cmd/stack_package.go
```diff
@@ -93,12 +93,23 @@ func newStackPackageCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
+			stackID := filepath.Base(projectPath)
+			Debug.Log("stackID is: ", stackID)
+
 			// sets stack path to be the copied folder
-			stackPath := projectPath + "stackPath"
+			stackPath := filepath.Join(getHome(rootConfig), "stacks", "packaging-"+stackID)
 			Debug.Log("stackPath is: ", stackPath)
 
+			// creates stackPath dir if it doesn't exist
+			err := os.MkdirAll(strings.Replace(stackPath, stackID, "", 1), 0777)
```
kylegc: What is the strings.Replace for?

cmd/stack_package_test.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"testing"
+
+	"github.com/appsody/appsody/cmd/cmdtest"
+)
+
+func TestStackPackageTest(t *testing.T) {
+
+	args := []string{"stack", "package"}
+	_, err := cmdtest.RunAppsodyCmdExec(args, "./testdata/starter")
```
kylegc: Did you try the preferred `RunAppsodyCmd` func (without the `Exec`)?

Issue #502, opened 2019-10-22
Add more git commit details to the image labels


kylegc:
When deployed to openshift, the appsody-operator resources could provide io.openshift labels. The appsody-operator may determine when running on openshift and add these labels dynamically, but the currently the appsody build command doesn't provide all the information required.

For example, see https://github.com/openshift/source-to-image/blob/master/pkg/scm/git/types.go for a list of the io.openshift git labels. We are missing the following:
- Commit date
- Commit message
- Commit author (name and email)
- Committer (name and email)
- ContextDir - subpath within the repository where the application code resides

These labels should be added under the `dev.appsody.image.` prefix

PR #532, opened 2019-10-28
Adding more git labels for Openshift

EnriqueL8:
#502 

FYI @kylegc I had to revert the cleaning of the functions (removing the config) because I need to use it to get the `projectDir`. I was thinking of passing only the `projectDir` and `dryrun`, but I think in the future we might use the `config` again.

Tested locally using a `nodejs-express` stack.

### DIFF
--- a/cmd/git_utils.go
+++ b/cmd/git_utils.go
@@ -23,10 +23,15 @@ import (
 )
 
 type CommitInfo struct {
-	Author string
-	SHA    string
-	Date   string
-	URL    string
+	Author         string
+	AuthorEmail    string
+	Committer      string
+	CommitterEmail string
+	SHA            string
+	Date           string
+	URL            string
+	Message        string
+	ContextDir     string "json:,omitempty"
 }
 
 type GitInfo struct {
@@ -81,7 +86,7 @@ func stringBetween(value string, pre string, post string) string {
 }
 
 //RunGitFindBranc issues git status
-func GetGitInfo(dryrun bool) (GitInfo, error) {
+func GetGitInfo(config *RootCommandConfig) (GitInfo, error) {
 	var gitInfo GitInfo
 	version, vErr := RunGitVersion(false)
 	if vErr != nil {
@@ -95,7 +100,7 @@ func GetGitInfo(dryrun bool) (GitInfo, error) {
 
 	kargs := []string{"status", "-sb"}
 
-	output, gitErr := RunGit(kargs, dryrun)
+	output, gitErr := RunGit(kargs, config.Dryrun)
 	if gitErr != nil {
 		return gitInfo, gitErr
 	}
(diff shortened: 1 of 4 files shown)


cmd/utils.go
```diff
@@ -636,6 +638,26 @@ func getGitLabels(dryrun bool) (map[string]string, error) {
 		}
 	}
 
+	if commitInfo.Author != "" && commitInfo.AuthorEmail != "" {
```
kylegc: I'm not sure if git allows one of these without the other, but I think we should still add the label if it is just the name or the email. If both are empty strings, skip the label. (same goes for committer)

cmd/utils.go
```diff
@@ -636,6 +638,26 @@ func getGitLabels(dryrun bool) (map[string]string, error) {
...
+
+	if commitInfo.Date != "" {
+		labels[appsodyImageCommitKeyPrefix+"date"] = commitInfo.Date
+	}
+
+	if commitInfo.Message != "" {
+		labels[appsodyImageCommitKeyPrefix+"message"] = commitInfo.Message
+	}
+
+	if commitInfo.contextDir != "" {
```
kylegc: Did you figure out the default value from S2I? Do they just not include this label if the project is at the root?

cmd/git_utils.go
```diff
@@ -191,8 +196,20 @@ func RunGitGetLastCommit(URL string, dryrun bool) (CommitInfo, error) {
 	if URL != "" {
 		commitInfo.URL = stringBefore(URL, ".git") + "/commit/" + commitInfo.SHA
 	}
-	return commitInfo, nil
 
+	gitLocation, gitErr := exec.Command("git", "rev-parse", "--show-toplevel").Output()
+	if gitErr != nil {
+		return commitInfo, gitErr
+	}
+	gitLocationString := strings.Replace(string(gitLocation), "\n", "", 1)
```
kylegc: Not a big deal, but could use the Trim() functions instead.

cmd/utils.go
EnriqueL8: These both suggest that we are overriding the default behaviour:
https://github.com/openshift/source-to-image/blob/master/docs/cli.md#context-directory
https://docs.openshift.com/container-platform/3.3/dev_guide/builds.html#source-code
So my intuition says that it should be blank?

cmd/utils.go
EnriqueL8: Apparently the name is always there: https://stackoverflow.com/questions/11579311/git-commit-as-different-user-without-email-or-only-email. So will change to allow for only the name, and then append the email if it exists.

cmd/utils.go
kylegc: Ok, we will not include the contextDir label if the appsody project is at the root of the git repo

cmd/utils.go
kylegc: That sounds good.
Looks like you forgot to remove the `&& commitInfo.AuthorEmail != ""` from this if check

PR #533, opened 2019-10-28
Build and deploy flags

### LINKED ISSUE #450: Appsody build must support pushing to an image registry
Currently `appsody build` only stores the built image in the local docker registry cache, and requires the user to issue a`docker push` command if the image must be uploaded to a registry.

Appsody build should include: 
1) a `--push` flag that triggers the push to the default image registry
2) a `--push-url` flag that allows the user to specify the registry URL for the image

This would enable the user to run:
`appsody build -t <my-repo>/<my-image> --push`
which would cause the image to be uploaded to the "default registry" (i.e. the registry to which the `docker` CLI is currently logged in).
Or:
`appsody build -t <my-repo>/<my-image> --push-url <registry-url>`
which causes the image to be pushed to `<registry-url>/<my-repo>/<my-image>`.

Related to #410

### LINKED ISSUE #451: Appsody deploy must allow specifying push and pull image registry URLs
This issue is related to #410: when the image registry is hosted on a private cloud or on a shared K8S cluster, it may be accessible through different URLs: 
* an externally addressable URL, for developers who need to push their image to the registry
* an internally addressable URL, which is visibile only within the K8S cluster, for pods that need to pull the image at container creation time.

Currently `appsody deploy` doesn't support specifying these two URLs, causing either a docker push failure as the command is run, or a docker pull failure when the contained in being created after deployment.

Provide the ability to specify a "push registry URL" and a "pull registry URL" in `appsody deploy`.  The push URL is used to executed the `docker push` command. The pull URL is stored in the `app-deploy.yaml` manifest.
To recap - users will be able to run: 
* `appsody deploy -t <my-repo>/<my-image> --push`
just as they do today. No changes here - the image will be pushed to the default registry and it is assumed to be pulled from that same registry by the K8S cluster (this is the docker.io use case).
* `appsody deploy -t <my-repo>/<my-image> --push-url <external registry URL> --pull-url <internal registry URL>`
which causes the image to be pushed to `<external registry URL>/<my-repo>/<my-image>`, and to be pulled from `<internal registry URL>/<my-repo>/<my-image>`
* `appsody/deploy` will still call the image `dev.local/<image name>` - no changes.

**Additional context**

### LINKED ISSUE #454: `appsody deploy --dryrun` fails
**Describe the bug**
`appsody deploy --dryrun` fails and returns nonzero error code

**Expected behavior**
dryrun should gracefully run, print what the command would do without doing it, and exit with a zero status

**Actual behaviour**
Here is the last part of the command output
```
Dry run skipped construction of file app-deploy.yaml
Created deployment manifest: app-deploy.yaml
Found existing deployment manifest app-deploy.yaml
[Error] Could not read app-deploy.yaml file: open app-deploy.yaml: no such file or directory
```

 - OS: macOS

 - CLI version: 0.4.6
 - Stack you are using: any

**Screenshots**

**Additional context**

Once this is fixed, we need to update the deploy_test.go TestDeploySimple so it does not ignore the error.

kewegner:
Fixes #451 
Also fixes dry run issue
And fixes build flags
Fixes #450 
Fixes #454

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -28,6 +28,8 @@ type buildCommandConfig struct {
 	*RootCommandConfig
 	tag                string
 	dockerBuildOptions string
+	pushURL            string
+	push               bool
 }
 
 func checkDockerBuildOptions(options []string) error {
@@ -59,7 +61,8 @@ func newBuildCmd(rootConfig *RootCommandConfig) *cobra.Command {
 
 	buildCmd.PersistentFlags().StringVarP(&config.tag, "tag", "t", "", "Docker image name and optionally a tag in the 'name:tag' format")
 	buildCmd.PersistentFlags().StringVar(&config.dockerBuildOptions, "docker-options", "", "Specify the docker build options to use.  Value must be in \"\".")
-
+	buildCmd.PersistentFlags().BoolVar(&config.push, "push", false, "Push the Docker image to the image repository.")
+	buildCmd.PersistentFlags().StringVar(&config.pushURL, "push-url", "", "The registry URL for the image repository.")
 	buildCmd.AddCommand(newBuildDeleteCmd(config))
 	buildCmd.AddCommand(newSetupCmd(config))
 	return buildCmd
@@ -118,6 +121,16 @@ func build(config *buildCommandConfig) error {
 
 	if execError != nil {
 		return execError
+	}
+	if config.push {
+		pushImage := buildImage
+		if config.pushURL != "" {
+			pushImage = config.pushURL + "/" + buildImage
+		}
+		err := DockerPush(pushImage, config.Dryrun)
+		if err != nil {
+			return errors.Errorf("Could not push the docker image - exiting. Error: %v", err)
+		}
 	}
 	if !config.Dryrun {
 		Info.log("Built docker image ", buildImage)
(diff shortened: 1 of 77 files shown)


cmd/build.go
```diff
@@ -119,6 +122,16 @@ func build(config *buildCommandConfig) error {
 	if execError != nil {
 		return execError
 	}
+	if config.push {
+		pushImage := buildImage
```
skoh7645: Should `pushImage := config.tag` rather than `buildImage` here? Since we would need their username as part of the tag, for `docker push` to know where to push the image

kewegner: Isn't buildImage set to the value of tag above if one is present?

kewegner:
Two comments:

1) the pull-url is not taken into account for --generate-only.  Is this needed?
2) I did not add the pull-url parameter to build.go in this PR.

cmd/build.go
skoh7645: But initially you set `buildImage := projectName`, and we only want to run `docker push` if `-t` has been specified and is not `config.tag != ""`, because doing `docker push` on `projectName` won't work

cmd/deploy.go
```diff
@@ -50,6 +50,28 @@ type Spec struct {
 	ApplicationImage string `yaml:"applicationImage"`
 }
 
+func findNamespaceRepositoryAndTag(image string) string {
+	var nameSpaceRepositoryAndTag string
+	if strings.Count(image, "/") > 1 {
```
chilanti: Could this have been done with `strings.Split(image, "/")`?

kewegner: yes please re-review

PR #535, opened 2019-10-29
cmd: check for stack minimum requirements and enforce if the user does not meet them

Kamran64:
Related to: https://github.com/appsody/stacks/issues/423

As per the discussions in the issue, this PR is a proof of concept to show how we could block a stacks usage if the user has out of date versions of certain technologies restricted by the stack. 

**Reasons to implement this:**
- Prevents users from beginning to write their application and then seeing that they cannot use the stack later down the `run/test/debug` line.
- Provides a more meaningful error message than a stream of Docker/k8s etc. errors.

For now the PR only enforces versions of Docker, the Appsody CLI, and Buildah before the init phase. We can add more technologies to this check somewhat easily e.g. k8s client/server before the deploy phase as a warning.

**Improvements (Tick off as added)**
- [x] I'd like to move some of the code in the `init` file into the `CheckStackRequirements` function so that we don't have duplicate code every time we want to do a check.
- [x] If possible, avoid using the `reflect` package as it potentially doesn't align with best practices(?)
- [x] Write tests
- [x] Don't enforce Docker if the user is using Buildah and vice versa

**Notes**
- This will be an optional feature for stack maintainers to use. Currently we would put this in for stacks that use:
    - `chown` as docker versions above 17.09.0 are required
    - `APPSODY_PREP` as older versions of the CLI do not support this env var.

For testing: I've been using an index with the following field defined under a stack:
```
requirements:
  docker-version: ">= 17.09.0"
  appsody-version: ">= 0.2.7"
```
If we decide to implement this, we will have to update the indexes first with this data and then put the PR in.

**Syntax for requirements**
To set a version constraint, add a new field to the `stack.yaml` titled `requirements` with nested fields e.g. `appsody-version`, `docker-version` etc. Version constraints follow the [semver package](https://github.com/Masterminds/semver) to allow stack maintainers to specify ranges and exact version matches among other things.


cmd/utils.go
```diff
@@ -2016,6 +2017,54 @@ func Targz(source, target string) error {
...
 }
 
+//Compares the minimum requirements of a stack against the user to determine whether they can use the stack or not.
+func CheckStackRequirements(requirementArray map[string]string, buildah bool) error {
+	versionRegex := regexp.MustCompile(`(\d)+\.(\d)+\.(\d)+`)
+	upgradesRequired := 0
+
+	Info.log("Checking stack requirements...")
+
+	for technology, minVersion := range requirementArray {
```
chilanti: As @kylegc mentioned, in some development scenarios, `appsody version` may return `vlatest`.
There are also situations where we return `0.0.0` (because unfortunately the debian and RPM packages can't be labelled `vlatest`).
 I would suggest checking these specific conditions and warning the user that there are no guarantees that the stack will work with this development version of Appsody.

Kamran64: Thanks for the review @chilanti - I've added a check so that if a valid version specified by the regex can't be retrieved from the cmd output (or if the version is 0.0.0), the user will be warned.

PR #541, opened 2019-10-30
Make stack github urls labels drill down to specific stack

### LINKED ISSUE #516: Stack image labels should point to stack specific readme or folder in github
The following oci labels point to github URLs:
```
"org.opencontainers.image.url": "https://github.com/appsody/stacks", 
"org.opencontainers.image.documentation": "https://github.com/appsody/stacks",
"org.opencontainers.image.source": "https://github.com/appsody/stacks/tree/master",
```

These URLs should drill down into the repo to point to the specific stack.
"documentation" should point specifically to the readme file.

kylegc:
Fixes #516

### DIFF
--- a/cmd/stack_package.go
+++ b/cmd/stack_package.go
@@ -348,6 +348,13 @@ func getLabelsForStackImage(stackID string, buildImage string, stackYaml StackYa
 		Info.log(err)
 	}
 
+	if contextDir, ok := gitLabels[appsodyImageCommitKeyPrefix+"contextDir"]; ok {
+		stackUrl := gitLabels[ociKeyPrefix+"source"] + contextDir
+		gitLabels[ociKeyPrefix+"url"] = stackUrl
+		gitLabels[ociKeyPrefix+"documentation"] = stackUrl + "/README.md"
+		gitLabels[ociKeyPrefix+"source"] = stackUrl + "/image"
+	}
+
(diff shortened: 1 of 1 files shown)


cmd/stack_package.go
```diff
@@ -348,6 +348,14 @@ func getLabelsForStackImage(stackID string, buildImage string, stackYaml StackYa
 		Info.log(err)
 	}
 
+	if contextDir, ok := gitLabels[appsodyImageCommitKeyPrefix+"contextDir"]; ok {
+		stackURL := gitLabels[ociKeyPrefix+"source"] + contextDir
+		gitLabels[ociKeyPrefix+"url"] = stackURL
+		// These are enforced by the stack lint so they should exist
+		gitLabels[ociKeyPrefix+"documentation"] = stackURL + "/README.md"
+		gitLabels[ociKeyPrefix+"source"] = stackURL + "/image"
```
EnriqueL8: What happens when the `contextDir` is empty? Can we point to the `README.md` in that case?

kylegc: If contextDir is empty, the key shouldn't be in the map so this code is skipped by the if check.

kylegc: Ok I see what you mean. If the user is building a stack off a different git repository, they may use the root of the repository. I've modified the code slightly to handle this case.

PR #544, opened 2019-10-30
Add labels to deploy

EnriqueL8:
Adding labels and annotations to `app-deploy.yaml`


cmd/build.go
```diff
@@ -135,18 +137,20 @@ func build(config *buildCommandConfig) error {
 	if !config.Dryrun {
 		Info.log("Built docker image ", buildImage)
 	}
+
+	Info.log(convertLabelsToKubeFormat(labels))
```
kylegc: Should this be here?

cmd/deploy.go
```diff
@@ -38,16 +41,24 @@ type deployCommandConfig struct {
...
 
 type AppsodyApplication struct {
-	APIVersion string   `yaml:"apiVersion"`
-	Kind       string   `yaml:"kind"`
-	Metadata   Metadata `yaml:"metadata"`
-	Spec       Spec     `yaml:"spec"`
+	APIVersion string                         `yaml:"apiVersion" json:"apiVersion"`
+	Kind       string                         `yaml:"kind" json:"kind"`
+	Metadata   Metadata                       `yaml:"metadata" json:"metadata"`
+	Spec       v1beta1.AppsodyApplicationSpec `yaml:"spec" json:"spec"`
```
chilanti: Explore using `v1beta1.AppsodyApplication` - (unless there's a good reason not to) - for future releases.

cmd/deploy.go
```diff
@@ -38,16 +41,24 @@ type deployCommandConfig struct {
...
 type Metadata struct {
-	Name string `yaml:"name"`
+	Name        string            `yaml:"name" json:"name"`
+	Labels      map[string]string `yaml:"labels" json:"labels"`
+	Annotations map[string]string `yaml:"annotations" json:"annotations"`
 }
-type Spec struct {
-	ApplicationImage string `yaml:"applicationImage"`
+
+var supportedLabels = []string{
```
kylegc: From the code I see that these will become k8s resource "labels" and everything else will be "annotations". But I didn't know what this meant until I dug into the code.  A better variable name or comment would be nice here.

cmd/deploy.go
EnriqueL8: Yes, I had not seen that `v1beta1.AppsodyApplicaiton` has a field for [ObjectType](https://github.com/kubernetes/apimachinery/blob/bb31b70367b776cca4a73412138411e77a36a104/pkg/apis/meta/v1/types.go#L110) which has [labels](https://github.com/kubernetes/apimachinery/blob/bb31b70367b776cca4a73412138411e77a36a104/pkg/apis/meta/v1/types.go#L231). Would need to investigate if our local `AppsodyApplication` is used somewhere else in the codebase, maybe when deploying to Knative?

PR #545, opened 2019-10-30
Use upstream branch name for git urls if it is different than local branch name

kylegc:


cmd/utils.go
```diff
@@ -626,6 +626,11 @@ func getGitLabels(config *RootCommandConfig) (map[string]string, error) {
 		labels[ociKeyPrefix+"url"] = gitInfo.RemoteURL
 		labels[ociKeyPrefix+"documentation"] = gitInfo.RemoteURL
 		labels[ociKeyPrefix+"source"] = gitInfo.RemoteURL + "/tree/" + gitInfo.Branch
+		upstreamSplit := strings.Split(strings.Split(gitInfo.Upstream, " ")[0], "/")
```
EnriqueL8: Just wondering why are we spliting the upstream by whitespaces? From https://github.com/appsody/appsody/blob/master/cmd/git_utils.go#L165, we just split on `/` since the format should be `upstream/branch`.

kylegc: Because the gitInfo.Upstream may contain a space and then more information. For example:
```
$ git status -sb
## useUpstreamBranchInGitUrls...kylegc/useUpstreamBranchInGitUrls [ahead 1]
```
I need to first strip off the `" [ahead 1]"` Then split on the `/`.

Perhaps this should have been done before gitInfo.Upstream was set?

kylegc: issue created https://github.com/appsody/appsody/issues/554

PR #551, opened 2019-10-31
Add new image-namespace argument and use value when building

groeges:
Add new image-namespace argument and use value when building

Signed-off-by: Steve Groeger GROEGES@uk.ibm.com

### DIFF
--- a/cmd/stack_package.go
+++ b/cmd/stack_package.go
@@ -71,6 +71,8 @@ func newStackPackageCmd(rootConfig *RootCommandConfig) *cobra.Command {
 	// 3. build a docker image
 	// 4. create/update an appsody repo for the stack
 
+	var imageNamespace string
+
 	var stackPackageCmd = &cobra.Command{
 		Use:   "package",
 		Short: "Package a stack in the local Appsody environment",
@@ -102,7 +104,7 @@ func newStackPackageCmd(rootConfig *RootCommandConfig) *cobra.Command {
 			appsodyHome := getHome(rootConfig)
 			Debug.Log("appsodyHome is:", appsodyHome)
 
-			devLocal := filepath.Join(appsodyHome, "stacks", "dev.local")
+			devLocal := filepath.Join(appsodyHome, "stacks", imageNamespace)
 			Debug.Log("devLocal is: ", devLocal)
 
 			// create the devLocal directory in appsody home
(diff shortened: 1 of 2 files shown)


cmd/stack_package.go
```diff
@@ -230,6 +232,9 @@ func newStackPackageCmd(rootConfig *RootCommandConfig) *cobra.Command {
 				return errors.Errorf("Error reading directories: %v", err)
 			}
 
+			repoDirName := filepath.Base(filepath.Dir(stackPath))
```
kylegc: We can't count on the repo name to be in the path. A user can run this command from any path in their filesystem.

cmd/stack_package.go
```diff
@@ -242,7 +247,7 @@ func newStackPackageCmd(rootConfig *RootCommandConfig) *cobra.Command {
 				Debug.Log("sourceDir is: ", sourceDir)
 
 				// create name for the tar files
-				versionedArchive := filepath.Join(devLocal, stackID+".v"+stackYaml.Version+".templates.")
+				versionedArchive := filepath.Join(devLocal, repoDirName+"."+stackID+".v"+stackYaml.Version+".templates.")
```
kylegc: I think we discussed that the CI scripts would prepend the repo name with the `add-to-repo --release-url` option.
Does that sound familiar?

cmd/stack_package.go
```diff
@@ -102,7 +104,7 @@ func newStackPackageCmd(rootConfig *RootCommandConfig) *cobra.Command {
 			appsodyHome := getHome(rootConfig)
 			Debug.Log("appsodyHome is:", appsodyHome)
 
-			devLocal := filepath.Join(appsodyHome, "stacks", "dev.local")
+			devLocal := filepath.Join(appsodyHome, "stacks", imageNamespace)
```
kylegc: I'm concerned about this. I know we had a discussion about whether or not the archives should always be stored in `~/.appsody/stacks/dev.local` or in separate dirs, but I don't think we came to a conclusion. 

With your code here, I believe we would be putting the `dev.local-index.yaml` in `~/.appsody/stacks/imageNamespace`. Therefore there could be multiple `dev.local-index.yaml`s around, and they won't get merged properly.

In my opinion, we should leave this hard coded to "dev.local" _for now._ This seems safest as we continue to develop the new `add-to-repo` command functionality.

groeges:
@kylegc Have made the changes you suggested and pushed the latest code. 
Could you please review again and see that everything is OK.

PR #557, opened 2019-10-31
Check kubernetes label value rules

kylegc:


cmd/utils.go
```diff
@@ -283,27 +283,48 @@ func getProjectDir(config *RootCommandConfig) (string, error) {
...
+	}
+	return false, errors.Errorf("Invalid project-name \"%s\". The name must start with a lowercase letter, contain only lowercase letters, numbers, or dashes, and cannot end in a dash.", name)
+}
+
+func IsValidKubernetesLabelValue(value string) (bool, error) {
+	if value == "" {
+		return true, nil
+	}
+	if len(value) > 68 {
+		return false, errors.New("The label must be less than 128 characters")
```
EnriqueL8: Error message should be 68

cmd/utils.go
```diff
@@ -283,27 +283,48 @@ func getProjectDir(config *RootCommandConfig) (string, error) {
...
 	}
 
-	return match, errors.Errorf("Invalid project-name \"%s\". The name must start with a lowercase letter, contain only lowercase letters, numbers, or dashes, and cannot end in a dash.", name)
+	match, err := regexp.MatchString("^[a-z0-9A-Z]([a-z0-9A-Z-_\\.]*[a-z0-9A-Z])?$", value)
```
EnriqueL8: If I am reading this correctly, we are letting the value have `\` and `_` - is this allowed my kubernetes?

kylegc: `_` is allowed but not `\`

The `\\` is to escape a single `\` in go, so that it can escape the `.` in regex.

However, I believe escaping the `.` in regex is optional when in `[ ]` so I've removed the slashes.

Issue #558, opened 2019-11-01
Stack linter should check for valid license values


neeraj-laad:
**Describe the bug**
Currently, the linter does not check if the license value is a valid SPDX license ID (https://spdx.org/licenses/). We must check that users specify an accurate value for this field as part of linting the stack - `appsody stack lint`

kylegc:
Does this issue also apply to the `license` field in `.appsody-config.yaml`? We should either handle that here or in another issue.

We should also check that the `license` field is a valid kubernetes label value. I've already done this for the `.appsody-config.yaml` but not the `stack.yaml`.
See utils.go `IsValidKubernetesLabelValue()` func.

skoh7645:
@kylegc thanks, I have made the relevant changes to the PR

PR #562, opened 2019-11-01
stack lint: check for valid license values

skoh7645:
Check that users specify an accurate value for `license` field in `stack.yaml` as part of linting the stack - `appsody stack lint`

Fixes: #558


Kamran64:
LGTM - I'll allow someone from the CLI team to take a look before approving :)

cmd/utils.go
```diff
@@ -617,7 +617,7 @@ func getConfigLabels(projectConfig ProjectConfig) (map[string]string, error) {
 	}
 
 	if projectConfig.License != "" {
-		if valid, err := IsValidKubernetesLabelValue(projectConfig.License); !valid {
+		if valid, err := IsValidKubernetesLabelValue(projectConfig.License); !valid || !checkValidLicense(projectConfig.License) {
 			return labels, errors.Errorf("%s license value is invalid. %v", ConfigFile, err)
```
kylegc: We need to have a better error message here if the license doesn't match an SPDX, otherwise the user will not know how to fix the error. I think we should link to a URL with all the valid values.

kylegc: Sandy, this is almost perfect, but there's a small issue with the error message. The getConfigLabels() method is called from both the stack package (using stack.yaml license) and from the build command (using .appsody-config.yaml). Thus, the error message cannot say "stack must have". Ideally we should tell the user exactly what file has the bad license.

cmd/utils.go
```diff
@@ -619,6 +620,8 @@ func getConfigLabels(projectConfig ProjectConfig) (map[string]string, error) {
 	if projectConfig.License != "" {
 		if valid, err := IsValidKubernetesLabelValue(projectConfig.License); !valid {
 			return labels, errors.Errorf("%s license value is invalid. %v", ConfigFile, err)
+		} else if err := checkValidLicense(projectConfig.License); err != nil {
+			return labels, errors.Errorf("The .appsody-config.yaml SPDX license ID is invalid: %v.", err)
```
kylegc: I'm pretty sure this can also be the `stack.yaml` - when called from stack package.

kylegc: Sorry @skoh7645 I'm not being clear. This is a utility method that can be called with information from either `.appsody-config.yaml` or `stack.yaml`. Therefore you cannot hardcode the filename here, you'll have to figure out a different way to print it in the message.

skoh7645: Sorry @kylegc for not understanding! I didn't realise we allowed users to add to the `.appsody-config.yaml` file to add their own labels. I have addressed this in the most recent commit

PR #566, opened 2019-11-04
Improve test output to only show detail on failures

### LINKED ISSUE #493: Revise logging strategy for CLI unit tests
Currently the CLI unit tests output a huge volume of text which can make it hard to find where failures are occurring. 

We should look at how we can best capture and organize the test output to make it more consumable.

kylegc:
Fixes #493 

This isn't pretty, but it was something I had created previously and used locally for myself.
It prints live updates of the current test being run and its PASS/FAIL status, then at the end prints the detailed output only for the failed tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 /package
 .vscode/
 cover.out
+test.out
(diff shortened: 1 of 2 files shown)


djones6:
@kylegc This is actually pretty cool 😄 But I agree it's a bit of a hack.

@upLukeWeston and I will investigate ways to do this via go test, but in the meantime, I don't see the harm in merging this as-is.  What do you think?

kylegc:
Haha yeah... the problem I had with `go test` is that it either outputs everything or nothing. My goal with this was to at least print the test names as they ran. That is important because if travis detects no output it will fail after a certain amount of time.

@chilanti what do you think about this. Should we merge it?

chilanti:
I think we should go ahead and merge it for the time being.

PR #567, opened 2019-11-04
Fix git labels from stack package

EnriqueL8:


cmd/stack_package.go
```diff
@@ -361,20 +361,21 @@ func getLabelsForStackImage(stackID string, buildImage string, stackYaml StackYa
...
+				branchURL += contextDir
+				gitLabels[ociKeyPrefix+"url"] = branchURL
+			}
+			// These are enforced by the stack lint so they should exist
+			gitLabels[ociKeyPrefix+"documentation"] = branchURL + "/README.md"
+			gitLabels[ociKeyPrefix+"source"] = branchURL + "/image"
+		}
+
+		for key, value := range gitLabels {
+			labelString := fmt.Sprintf("%s=\"%s\"", key, value)
```
kylegc: Why did you add the quotes here? Do quotes also need to be added here? https://github.com/appsody/appsody/blob/master/cmd/build.go#L255

EnriqueL8: I will revert this change - I was wondering if we could have an issue with docker build command if we are not wrapping the labels around quotes... Need to investigate further on that.

Issue #572, opened 2019-11-05
appsody stack create: errors if directory with stack name already exists


skoh7645:
**Describe the bug**
The `appsody stack create --copy` flag does not work if you have a project with the same stack name as the stack you want to copy

**To Reproduce**
1. Have a project named `nodejs-express` in your current working directory
2. Run appsody stack create my-stack --copy incubator/nodejs-express
3. See error

**Expected behavior**
The command should work as this should not be a limitation

**Actual behaviour**
Command gives an error

**Environment Details (please complete the following information):**

**If applicable please specify:**
 - CLI version: 0.4.9
 - Stack you are using:

**Screenshots**
<img width="819" alt="Screenshot 2019-11-05 at 11 52 01" src="https://user-images.githubusercontent.com/52743301/68205812-19ea4580-ffc3-11e9-84a5-6a961a7f1b91.png">

PR #596, opened 2019-11-06
appsody stack create: errors if directory with stack name already exists bug fix

Bazif-Khan:
fixes: https://github.com/appsody/appsody/issues/572

### DIFF
--- a/cmd/stack_create.go
+++ b/cmd/stack_create.go
@@ -50,6 +50,7 @@ Examples:
 
 The stack name must start with a lowercase letter, and can contain only lowercase letters, numbers, or dashes, and cannot end with a dash. The stack name cannot exceed 128 characters.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			ExistingStackFolderExist := false
 
 			if len(args) < 1 {
 				return errors.New("Required parameter missing. You must specify a stack name")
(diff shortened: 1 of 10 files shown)


cmd/stack_create.go
```diff
@@ -111,23 +110,43 @@ The stack name must start with a lowercase letter, and can contain only lowercas
 				Info.logf("Dry Run -Skipping moving out of stack: %s from %s", projectType, filepath.Join(stack, "stacks-master", config.copy))
 
 			} else {
-				err = os.Rename(filepath.Join(stack, "stacks-master", config.copy), projectType)
+				ExistingStackFolderExist, err = Exists(projectType)
```
kylegc: Is this `projectType` variable used for anything other than naming this temporary directory? 
A couple suggestions:
1. Give `projectType` a better name. If it is supposed to refer to the stack, call it `stackID`. If it is just used for a temporary directory name, call it something like `stackTempDir`
2. Can you always set the directory name with the timestamp? That way you don't have to use `ExistingStackFolderExist` and have multiple code branches everywhere.
Maybe even start it with a dot (to make it hidden) and include `temp` in the name. That way if it doesn't get cleaned up the it will be clear to the user that it is temporary.
like this `.temp-nodejs-express-20060102150405`

Bazif-Khan: I have made changes. Thanks for the useful comments.

cmd/stack_create.go
```diff
@@ -108,10 +106,12 @@ The stack name must start with a lowercase letter, and can contain only lowercas
 
 			//moving out the stack which we need
 			if config.Dryrun {
-				Info.logf("Dry Run -Skipping moving out of stack: %s from %s", projectType, filepath.Join(stack, "stacks-master", config.copy))
+				Info.logf("Dry Run -Skipping moving out of stack: %s from %s", stackTempDir, filepath.Join(stack, "stacks-master", config.copy))
 
 			} else {
-				err = os.Rename(filepath.Join(stack, "stacks-master", config.copy), projectType)
+				stackTempDir = ".temp-" + stackTempDir + "-" + currentTime
```
kylegc: Just a small suggestion, this line should be moved up to where you first define the `stackTempDir` variable (line 90)

Bazif-Khan: I'm setting the value of `stackTempDir` variable in else case because we want to set the value of `stackTempDir` if the user has not selected dry run mode.

If the user is running in the dry run mode, then we want the value of `stackTempDir` to be the existing stack name which the user has given in copy flag as I'm using that in (line 109).

kylegc: I don't understand why you want to print something different in dryrun mode. Shouldn't you print what will actually happen?

Bazif-Khan: When we unzip a particular stack e.g. nodejs which user wants, it is in the structure `stacks-master/incubator/nodejs`. In (line 109) I'm printing that I'm moving out the stack which user want so when the user is running dry run mode, I want to print properly the name of the stack which I'm moving out, not the name of the temporary directory.

kylegc: ok thanks. I'll approve and merge

PR #601, opened 2019-11-06
Volume mount Appsody Controller

### LINKED ISSUE #399: Appsody on K8S should avoid separate mount for the controller
Appsody on K8S should minimize the number of volume mounts. Currently, it requires a separate PVC for the appsody controller, which could be copied in the workspace mount.

Remove the PVC for controller, and copy the appsody controller into the workspace instead. The deployment yaml that gets generated by `appsody run` needs to reflect the new mount strategy.

Leave things as they are today and have Che only use a single volume for all users where the controller is stored.

### LINKED ISSUE #467: Copy controller into container instead of mounting

In an upcoming release of Codewind, the existing `/codewind-workspace` docker bind mount will be removed.  This will mean that it's no longer possible to bind mount the controller located at `/codewind-workspace/.extensions/codewind-appsody-extension/bin/appsody-controller` into the Appsody container (since it won't exist anywhere on the host system anymore)

Can we have an option to copy the controller into the Appsody container instead?

**Additional context**

### LINKED ISSUE #600: Appsody should avoid bind-mounting the controller
Some of the scenarios in which Appsody has to operate (Codewind, K8S, Che) would be simpler to support if Appsody didn't rely on bind mounts for the controller executable.

Switch to a volume-mount approach. The controller is packaged separately - in a docker image, which can populate the volume. 

Copying the controller into the target app container. This solution, however, would be hard to support in the K8S scenario.

**Additional context**
See #399 (#467).

chilanti:
Fixes #600 
Fixes #399 
Fixes #467


Makefile
```diff
@@ -22,10 +22,10 @@ BINARY_EXT_linux :=
 BINARY_EXT_darwin :=
 BINARY_EXT_windows := .exe
 DOCKER_IMAGE_RPM := alectolytic/rpmbuilder
-DOCKER_IMAGE_DEB := appsody/debian-builder
+DOCKER_IMAGE_DEB := appsody/debian-builder:0.1.0
 GH_ORG ?= appsody
-CONTROLLER_VERSION ?=0.2.6
-CONTROLLER_BASE_URL := https://github.com/${GH_ORG}/controller/releases/download/$(CONTROLLER_VERSION)
+CONTROLLER_VERSION ?=0.3.0
```
kylegc: I think you should export this and make it APPSODY_CONTROLLER_VERSION.
That way `make test` will use it instead of `latest`

cmd/dev_common.go
```diff
@@ -142,75 +155,45 @@ func commonCmd(config *devCommonConfig, mode string) error {
...
-		executable, _ := os.Executable()
-		binaryLocation, err := filepath.Abs(filepath.Dir(executable))
-		Debug.log("Binary location ", binaryLocation)
+	//destController := os.Getenv("APPSODY_MOUNT_CONTROLLER")
+	controllerImageName := overrideControllerImage
+	//if destController != "" {
+	//	Debug.log("Overriding appsody-controller mount with APPSODY_MOUNT_CONTROLLER env variable: ", destController)
+	//} else {
+	//destController := "appsody-controller"
+	controllerVolumeName := fmt.Sprintf("%s-%s", "appsody-controller", CONTROLLERVERSION)
```
kylegc: I'm concerned when the CONTROLLERVERSION is `latest`, the volume named `appsody-controller-latest` will never be updated. 
This should only happen in development scenarios, a.k.a. someone is developing a new controller and wants to test a dev image with the CLI. However, even in that case, I think we need to run the init-controller image to always update the volume.

.travis.yml
```diff
@@ -18,14 +18,15 @@ stages:
   - name: deploy
     if: tag IS present
 
-install-controller: &install-controller
-  install:
-      - if [ "$TRAVIS_OS_NAME" = "windows" ]; then choco install make; fi
-      - make install-controller
+#install-controller: &install-controller
```
kylegc: I'd vote to remove commented code. Keeps things cleaner.

cmd/dev_common.go
chilanti: This has been addressed

PR #603, opened 2019-11-06
Output version update message to Warning

james-wallis:
### Summary 
* Print the version update message to Stderr (this means that scripts can read only from stdout when parsing JSON from the `-o json` flag)
* Reformat the `doVersionCheck` function so that minimal upkeep is needed if the message is changed in common parts and added `GetUpdateString` function.

### Testing
Manually tested.
Written unit tests for `GetUpdateString` function to test different output depending on OS.

### Note
The only catch with using `Warning.logf` instead of `Info.logf` is the `[Warning]` that is printed before the version message so the output for MacOS is now:
```
[Warning] 
*
*
*

A new CLI update is available.
Please run `brew upgrade appsody` to upgrade from vlatest --> 0.4.10.

*
*
*
```
@kylegc FYI.

Signed-off-by: James Wallis <james.wallis1@ibm.com>

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -1805,14 +1805,9 @@ func getLatestVersion() string {
 func doVersionCheck(config *RootCommandConfig) {
 	var latest = getLatestVersion()
 	var currentTime = time.Now().Format("2006-01-02 15:04:05 -0700 MST")
-
-	if latest != "" && VERSION != "vlatest" && VERSION != latest {
-		switch os := runtime.GOOS; os {
-		case "darwin":
-			Info.logf("\n*\n*\n*\n\nA new CLI update is available.\nPlease run `brew upgrade appsody` to upgrade from %s --> %s.\n\n*\n*\n*", VERSION, latest)
-		default:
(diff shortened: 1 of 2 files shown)


cmd/utils.go
```diff
@@ -1805,14 +1805,9 @@ func getLatestVersion() string {
...
 	var currentTime = time.Now().Format("2006-01-02 15:04:05 -0700 MST")
-
-	if latest != "" && VERSION != "vlatest" && VERSION != latest {
-		switch os := runtime.GOOS; os {
-		case "darwin":
-			Info.logf("\n*\n*\n*\n\nA new CLI update is available.\nPlease run `brew upgrade appsody` to upgrade from %s --> %s.\n\n*\n*\n*", VERSION, latest)
-		default:
-			Info.logf("\n*\n*\n*\n\nA new CLI update is available.\nPlease go to https://appsody.dev/docs/getting-started/installation#upgrading-appsody and upgrade from %s --> %s.\n\n*\n*\n*", VERSION, latest)
-		}
+	if true || latest != "" && VERSION != "vlatest" && VERSION != latest {
```
kylegc: Why `true ||`? Doesn't this make the `if` always evaluate to `true`?

james-wallis: Oops. Missed that when I was cleaning up the code. Thanks for pointing it out!

Issue #177, opened 2019-08-07
Improve cli development process documentation


kewegner:
We should highlight the following for contributing new code to appsody:
- Enhance the documentation for the various tags in the makefile
- Enhance the documentation for .travis.yaml
- New developers should use make test locally to test their code prior to doing PRs
-New developers should use make lint locally to lint their code prior to doing PRs
- Clarify how to set up the binaries (appsody and appsody-controller) when developing new function for appsody locally

henrynash:
I strongly agree with this (having just had to go through working out nearly all of the above from first principles for this first time!). If this is already being worked on, then great, otherwise, I'd be happy to take a crack updating the documentation, if that would help.

chilanti:
@henrynash - are you still available to work on this? If you are - feel free to assign this to yourself. Thanks.

henrynash:
HI, yes, I'll pick this one up...

PR #604, opened 2019-11-07
Better description and additional makefile support

henrynash:
Improve the docs on building the CLI locally (and enhance the make file to make to easier to get a copy of your build CLI).

Partially Fixes: #177

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,10 +128,10 @@ build-darwin: ## Build the OSX binary
 .PHONY: localbin-darwin
 localbin-darwin: ## copy the darwin binary to local bin
 
-.PHONY: localbin-darwin
+.PHONY: localbin-linux
 localbin-linux: ## copy the linux binary to local bin
 
-.PHONY: localbin-darwin
+.PHONY: localbin-windows
 localbin-windows: ## copy the windows binary to local bin
 
 localbin-darwin localbin-linux localbin-windows:
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ GO_TEST_COVER_VIEWER := go tool cover -func=cover.out && go tool cover -html=cov
 VERSION ?= 0.0.0
 COMMAND := appsody
 BUILD_PATH := $(PWD)/build
+LOCAL_BIN_PATH := $(PWD)/bin
 PACKAGE_PATH := $(PWD)/package
 DOCS_PATH := $(PWD)/my-project
 GO_PATH := $(shell go env GOPATH)
@@ -100,6 +101,7 @@ ensure: $(DEP_BINARY) ## Runs `dep ensure` to make sure the Gopkg.lock and /vend
 .PHONY: clean
 clean: ## Removes existing build artifacts in order to get a fresh build
 	rm -rf $(BUILD_PATH)
+	rm -rf $(LOCAL_BIN_PATH)
 	rm -rf $(PACKAGE_PATH)
 	rm -f $(GOLANGCI_LINT_BINARY)
 	rm -f $(DEP_BINARY)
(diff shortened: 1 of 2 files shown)


build.md
```diff
@@ -1,52 +1,80 @@
...
 * go version 1.12 or higher is installed and in the path
 * docker is installed and running
 * wget is installed
 
-After setting the `GOPATH` env var correctly, just run `make <action...>` from the command line, within the same directory where `Makefile` resides. For example `make package clean` will run the `package` and then the `clean` actions.
+After setting the `GOPATH` env var correctly, just run `make <action...>` from the command line, within the same directory where `Makefile` resides.
 
+## What is supported by the makefile to build?
 
-Some of the scripts have conditional paths, because certain Linux commands behave differently on OS/X and elsewhere (fun).
```
helenmasters: Henry, whilst you're editing this file, please can you change the OS/X references to macOS as we shouldn't be referring to OS/X in our docs? (I do know that you didn't put those references in!)

henrynash: will do

build.md
```diff
@@ -1,52 +1,80 @@
...
 
-* Next, you need to build the `appsody-controller` (sorry). It's here: https://github.com/appsody/controller. Just build the binary, and move it to the /usr/local/Cellar/appsody/`<release_tag>`/appsody-controller. Call it `appsody-controller` (mandatory!) and make sure that it has +x permissions (also mandatory).
+```bash
+make build-darwin localbin-darwin
+```
 
-* If you are replacing an old installation, you may want to delete the appsody home directory (`$HOME/.appsody`).
+You can simply execute the newly build CLI by referencing it directly (`./bin/appsody`) or add this `bin` directory to your $PATH (or %PATH% if on Windows) and then just use the `appsody` command.
+
+If your CLI changes are not dependent on changes to the appsody controller, then there is nothing else you need to do to test locally. If your changes do need a matching appsody controller, then you need to ensure that the respective controller is installed in your `~/.appsody` directory (since it is from there that it is then mounted into the stack image at docker run time).
```
kylegc: As of 0.5.0, the controller is no longer mounted from `~/.appsody`, and `make install-controller` is gone. We now get the controller from pulling a docker image. Developers can customize the image by specifying either environment variable:
`APPSODY_CONTROLLER_VERSION` - the version tag of the `appsody/init-controller` image to use
`APPSODY_CONTROLLER_IMAGE` - a full image override to init the controller

Makefile
```diff
@@ -134,6 +136,19 @@ build-linux build-windows: ## Build the binary of the respective operating syste
 build-darwin: ## Build the OSX binary
 	GOOS=$(os) GOARCH=amd64 go build -o $(BUILD_PATH)/$(build_binary) -ldflags "-X main.VERSION=$(VERSION)"
 
+.PHONY: localbin-darwin
+localbin-darwin: ## copy the darwin binary to local bin
+
+.PHONY: localbin-darwin
```
kylegc: This should be `localbin-linux`

Makefile
```diff
@@ -134,6 +136,19 @@ build-linux build-windows: ## Build the binary of the respective operating syste
...
+localbin-linux: ## copy the linux binary to local bin
+
+.PHONY: localbin-darwin
```
kylegc: This should be `localbin-windows`

Makefile
henrynash: oops, will fix

Makefile
henrynash: oops, will fix

build.md
henrynash: got it, will amend accordingly

neeraj-laad:
@henrynash Would you be able to make these changes in the next couple of days?

henrynash:
@neeraj-laad Yep, sorry for the delay....update pushed to resolve the issues

Issue #576, opened 2019-11-05
Standardize 'appsody debug' help text format 


skoh7645:
Apply standard created for CLI commands' help text under issue #463, to the `appsody debug` command

PR #607, opened 2019-11-07
appsody debug: reformat help text

skoh7645:
Changed help text formatting for `appsody debug` command

Fixes: #576

### DIFF
--- a/cmd/debug.go
+++ b/cmd/debug.go
@@ -26,7 +26,7 @@ func newDebugCmd(rootConfig *RootCommandConfig) *cobra.Command {
   Starts the debugging environment, passing the "--privileged" option to the "docker run" command as a flag.
   
   appsody debug --name my-project-dev2 -p 3001:3000
-  Starts the debugging environment, names the development container "my-project-dev2", binds the container port 3000 to the host port 3001.`,
+  Starts the debugging environment, names the development container "my-project-dev2", and binds the container port 3000 to the host port 3001.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 
 			Info.log("Running debug environment")
--- a/cmd/debug.go
+++ b/cmd/debug.go
@@ -20,8 +20,13 @@ func newDebugCmd(rootConfig *RootCommandConfig) *cobra.Command {
 	// debug Cmd represents the debug command
 	var debugCmd = &cobra.Command{
 		Use:   "debug",
-		Short: "Run the local Appsody environment in debug mode",
-		Long:  `This starts a docker based continuous build environment for your project with debugging enabled.`,
+		Short: "Debug your Appsody project",
+		Long:  `Start a container-based continuous build environment for your Appsody project, with debugging enabled.`,
+		Example: `  appsody debug --docker-options "--privileged"
+  Starts debugging environment and passes the "--privileged" option to the docker run as a flag.
+  
+  appsody debug --name my-project-dev2 -p 3001:3000
+  Starts debugging environment, names development container "my-project-dev2" and binds container port 3000 to host port 3001.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 
 			Info.log("Running debug environment")
--- a/cmd/debug.go
+++ b/cmd/debug.go
@@ -20,13 +20,13 @@ func newDebugCmd(rootConfig *RootCommandConfig) *cobra.Command {
 	// debug Cmd represents the debug command
 	var debugCmd = &cobra.Command{
 		Use:   "debug",
-		Short: "Debug your Appsody project",
+		Short: "Debug your Appsody project.",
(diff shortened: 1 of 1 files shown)


cmd/debug.go
```diff
@@ -20,8 +20,13 @@ func newDebugCmd(rootConfig *RootCommandConfig) *cobra.Command {
 	// debug Cmd represents the debug command
 	var debugCmd = &cobra.Command{
 		Use:   "debug",
-		Short: "Run the local Appsody environment in debug mode",
-		Long:  `This starts a docker based continuous build environment for your project with debugging enabled.`,
+		Short: "Debug your Appsody project",
```
doveye: Please add a full-stop to the end: "Run the local Appsody environment in debug mode."

cmd/debug.go
```diff
@@ -20,8 +20,13 @@ func newDebugCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
+		Long:  `Start a container-based continuous build environment for your Appsody project, with debugging enabled.`,
```
helenmasters: I'm going to ask @SueChaplain, or @doveye to review the text of these changes.

@skoh7645 as part of this change set please can you clarify the descriptions for the following flags:
--name
--deps-volume
--docker-options
--network
--no-watcher
--publish
--publish-all
as it looks like all of these optional flags apply to the run, test and debug modes on the development container and I don't think that the current descriptions are clear enough. Specifically I found --deps-volume confusing, what if APPSODY_DEPS dir is already set?

cmd/debug.go
```diff
@@ -20,8 +20,13 @@ func newDebugCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
+		Example: `  appsody debug --docker-options "--privileged"
+  Starts debugging environment and passes the "--privileged" option to the docker run as a flag.
```
doveye: Need more articles, ie "the debugging environment". What do you mean by "docker run"? Do you mean "the Docker "run" command", or "the "docker run" command", or something else?

cmd/debug.go
```diff
@@ -20,8 +20,13 @@ func newDebugCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
+  
+  appsody debug --name my-project-dev2 -p 3001:3000
+  Starts debugging environment, names development container "my-project-dev2" and binds container port 3000 to host port 3001.`,
```
doveye: - Articles again :)  "The debugging environment", "the development container", "the container port". 
- Comma before the "and" at the end of a list, ie "Starts xxx, names xxx, and binds xxx".
- Do we need to say a bit more about the development container, or will it be obvious to the user? eg "Starts the debugging environment with a development container named ...". As a non-expert, I'm wondering where the development container comes from?

skoh7645: I think it should be obvious enough to the user? The `appsody <run|test|debug>` commands all run a `docker run` command which starts the development container

cmd/debug.go
skoh7645: I kept leaving out articles to make it fit better on the CLI - I will push these changes but might need help to make these example explanations shorter!

cmd/debug.go
skoh7645: I tried it out and the command creates a docker volume using the name of the argument passed to `--deps-volume` and mounts it to the APPSODY_DEPS dir.

cmd/debug.go
doveye: Ah ok, didn't realise there was a length issue! You could maybe have "Starts the debugging environment, passing the "--privileged" option to the docker run as a flag.", though that's only slightly shorter.

cmd/debug.go
doveye: Fair enough :)

cmd/debug.go
skoh7645: Thanks! I have pushed the chnages

cmd/debug.go
```diff
@@ -20,13 +20,13 @@ func newDebugCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
-		Short: "Debug your Appsody project",
+		Short: "Debug your Appsody project.",
 		Long:  `Start a container-based continuous build environment for your Appsody project, with debugging enabled.`,
 		Example: `  appsody debug --docker-options "--privileged"
-  Starts debugging environment and passes the "--privileged" option to the docker run as a flag.
+  Starts the debugging environment and passes the "--privileged" option to the "docker run" command as a flag.
   
   appsody debug --name my-project-dev2 -p 3001:3000
-  Starts debugging environment, names development container "my-project-dev2" and binds container port 3000 to host port 3001.`,
+  Starts the debugging environment, names the development container "my-project-dev2" and, binds the container port 3000 to the host port 3001.`,
```
doveye: Comma should be before the add rather than after:
 "Starts the debugging environment, names the development container "my-project-dev2", and binds the container port 3000 to the host port 3001."

Issue #580, opened 2019-11-05
Standardize 'appsody list' help text format


skoh7645:
Apply standard created for CLI commands' help text under issue #463, to the `appsody list` command

PR #613, opened 2019-11-08
appsody list: reformat help text

skoh7645:
Changed help text formatting for `appsody list` command

Fixes: #580

### DIFF
--- a/cmd/list.go
+++ b/cmd/list.go
@@ -35,7 +35,7 @@ func newListCmd(rootConfig *RootCommandConfig) *cobra.Command {
 		Short: "List the available Appsody stacks.",
 		Long: `List all the Appsody stacks available in your repositories. 
 
-An asterisk in the repository column denotes the default repository. An asterisk in the template column denotes the default template used, when you initialise an Appsody project.`,
+An asterisk in the repository column denotes the default repository. An asterisk in the template column denotes the default template that is used when you initialise an Appsody project.`,
 		Example: `  appsody list
   Lists all available stacks for each of your repositories.
   
--- a/cmd/list.go
+++ b/cmd/list.go
@@ -32,10 +32,10 @@ func newListCmd(rootConfig *RootCommandConfig) *cobra.Command {
 	// listCmd represents the list command
 	var listCmd = &cobra.Command{
 		Use:   "list [repository]",
-		Short: "Lists the available Appsody stacks.",
-		Long: `Lists all the Appsody stacks available in your repositories. 
+		Short: "List the available Appsody stacks.",
+		Long: `List all the Appsody stacks available in your repositories. 
 
-An asterisk under the repository column denotes the default repository. An asterisk under the template column denotes the default stack template you will get when you initialise an Appsody project.`,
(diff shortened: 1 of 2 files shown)


cmd/list.go
```diff
@@ -32,8 +32,15 @@ func newListCmd(rootConfig *RootCommandConfig) *cobra.Command {
 	// listCmd represents the list command
 	var listCmd = &cobra.Command{
 		Use:   "list [repository]",
-		Short: "List the Appsody stacks available to init",
-		Long:  `This command lists all the stacks available in your repositories. If you omit the  optional [repository] parameter, the stacks for all the repositories are listed. If you specify the repository name [repository], only the stacks in that repository will be listed.`,
+		Short: "Lists the available Appsody stacks.",
+		Long: `Lists all the Appsody stacks available in your repositories.
```
doveye: Should be "List" rather than "Lists", to be consistent with the create command, and others.

cmd/list.go
```diff
@@ -32,8 +32,15 @@ func newListCmd(rootConfig *RootCommandConfig) *cobra.Command {
 	// listCmd represents the list command
 	var listCmd = &cobra.Command{
 		Use:   "list [repository]",
-		Short: "List the Appsody stacks available to init",
-		Long:  `This command lists all the stacks available in your repositories. If you omit the  optional [repository] parameter, the stacks for all the repositories are listed. If you specify the repository name [repository], only the stacks in that repository will be listed.`,
+		Short: "Lists the available Appsody stacks.",
+		Long: `Lists all the Appsody stacks available in your repositories. 
+
+An asterisk under the repository column denotes the default repository. An asterisk under the template column denotes the default stack template you will get when you initialise an Appsody project.`,
```
doveye: Do we mean "under" the column, or "in"?
Not a fan of "you will get when" - how about " that is used when"?

skoh7645: I have changed this to "in the column"

cmd/list.go
```diff
@@ -32,10 +32,10 @@ func newListCmd(rootConfig *RootCommandConfig) *cobra.Command {
 	// listCmd represents the list command
 	var listCmd = &cobra.Command{
 		Use:   "list [repository]",
-		Short: "Lists the available Appsody stacks.",
-		Long: `Lists all the Appsody stacks available in your repositories. 
+		Short: "List the available Appsody stacks.",
+		Long: `List all the Appsody stacks available in your repositories. 
 
-An asterisk under the repository column denotes the default repository. An asterisk under the template column denotes the default stack template you will get when you initialise an Appsody project.`,
+An asterisk in the repository column denotes the default repository. An asterisk in the template column denotes the default template used, when you initialise an Appsody project.`,
```
doveye: You don't need a comma after "used":
"... denotes the default template used when you initialise an Appsody project."
(Being picky, as a writer I'd say "... denotes the default template that is used when you initialise an Appsody project.", but since this needs to be short and won't be translated, it's not really necessary :) )

skoh7645: I made the changes you requested as it's only the example explanations that we need to try keep short :)

Issue #579, opened 2019-11-05
Standardize 'appsody init' help text format


skoh7645:
Apply standard created for CLI commands' help text under issue #463, to the `appsody init` command

PR #615, opened 2019-11-08
init: reformat help text

skoh7645:
Changed help text formatting for the `appsody init` command

Fixes: #579


cmd/init.go
```diff
@@ -48,18 +48,24 @@ func newInitCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
-With the [stack], [repository]/[stack], [stack] [template] or [repository]/[stack] [template] arguments, this command will setup a new Appsody project. It will create an Appsody stack config file, unzip a template app, and run the stack init script to setup the local dev environment. It is typically run on an empty directory and may fail
-if files already exist. See the --overwrite and --no-template options for more details.
-Use 'appsody list' to see the available stack options.
-
-If keyword "none" is specified instead of a [template], the project will be initialized to use Appsody, and no template will be provided.
-
-Without the [stack] argument, this command must be run on an existing Appsody project and will only run the stack init script to
-setup the local dev environment.`,
+		Short: "Initialize an Appsody project.",
+		Long: `Sets up the local Appsody development environment around an existing project or, optionally provides a template project. In all cases, this command unzips a template application, and runs the stack init script to setup the local development environment.
```
neeraj-laad: `In all cases, this command unzips a template application, and runs the stack init script to setup the local development environment.`

I don't believe this is accurate.

neeraj-laad: How about:

Sets up the local Appsody development environment. You can do this for an existing project or create a new template application.

cmd/init.go
```diff
@@ -48,18 +48,24 @@ func newInitCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
-If keyword "none" is specified instead of a [template], the project will be initialized to use Appsody, and no template will be provided.
-
-Without the [stack] argument, this command must be run on an existing Appsody project and will only run the stack init script to
-setup the local dev environment.`,
+		Short: "Initialize an Appsody project.",
+		Long: `Sets up the local Appsody development environment around an existing project or, optionally provides a template project. In all cases, this command unzips a template application, and runs the stack init script to setup the local development environment. 
+
+By default, the command creates an Appsody stack configuration file and provides a simple default application. You can also choose to initialize a project with a different template application, or no template at all. 
+
+This command will fail if you are not in an empty directory, unless you specify you want to overwrite these files.
```
neeraj-laad: you say these files but not describe or list them.

cmd/init.go
```diff
@@ -48,18 +48,24 @@ func newInitCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
+Use 'appsody list' to see the available stack options.`,
```
neeraj-laad: Use 'appsody list' to see the available stacks and templates.

cmd/init.go
```diff
@@ -48,18 +48,24 @@ func newInitCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
+		Example: `  appsody init experimental/quarkus 
+  Initializes a project with the "quarkus" default stack template from your "experimental" repository.
```
neeraj-laad: Initializes a project with the default template for "quarkus" stack in the "experimental" repository.

neeraj-laad: Perhaps the first example could avoid the repository use?

cmd/init.go
```diff
@@ -48,18 +48,24 @@ func newInitCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
+
+  appsody init nodejs-express scaffold
+  Initializes a project with the "nodejs-express" "scaffold" stack template from your default repository.
```
neeraj-laad: Initializes a project with the "scaffold" template from "nodejs-express" stack in the default repository.

cmd/init.go
```diff
@@ -48,18 +48,24 @@ func newInitCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
+  
+  appsody init nodejs-express none
+  Initializes a project with the "nodejs-express" stack from your default repository, with no initial template application provided.
```
neeraj-laad: should we say that it will create the .appsody_config file and runs the stack init script to setup the local development environment?

cmd/init.go
doveye: How about:
"If you run this command in a directory that is not empty, the command fails unless you specify the "overwrite" option[or flag, whichever term we use]." (Assuming that option is described).

cmd/init.go
doveye: Should be "Set up...", to be consistent with the "create" command, and others. 
Think the other sentence needs a bit of clarification - "Set up the local Appsody development environment for an existing project" makes sense, but "Set up the local Appsody development environment create a new template application" doesn't.

cmd/init.go
skoh7645: @neeraj-laad "these files" are referring to the files that are on their local machine that they want to overwrite with an `init`

cmd/init.go
skoh7645: @neeraj-laad `appsody init` won't create a config file, it only runs the init script

cmd/init.go
```diff
@@ -48,18 +48,24 @@ func newInitCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
 
-If the [repository] is not specified the default repository will be used. If no [template] is specified, the default template will be used.
-With the [stack], [repository]/[stack], [stack] [template] or [repository]/[stack] [template] arguments, this command will setup a new Appsody project. It will create an Appsody stack config file, unzip a template app, and run the stack init script to setup the local dev environment. It is typically run on an empty directory and may fail
-if files already exist. See the --overwrite and --no-template options for more details.
-Use 'appsody list' to see the available stack options.
+By default, the command creates an Appsody stack configuration file and provides a simple default application. You can also choose to initialize a project with a different template application, or no template at all. 
 
-If keyword "none" is specified instead of a [template], the project will be initialized to use Appsody, and no template will be provided.
+If you run this command in a directory that is not empty, and you want to initialise a project with a template application, the command fails unless you specify the "overwrite" option [--overwrite].
```
neeraj-laad: shouldn't this be in the Example section?

cmd/init.go
```diff
@@ -48,18 +48,24 @@ func newInitCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
+		Example: `  appsody init nodejs-express
+  Initializes a project with the default template from the "nodejs-express" stack in the default repository.
+  
+  appsody init nodejs-express scaffold
+  Initializes a project with the "scaffold" template from "nodejs-express" stack in the default repository.
 
-Without the [stack] argument, this command must be run on an existing Appsody project and will only run the stack init script to
-setup the local dev environment.`,
+  appsody init experimental/quarkus none
+  Initializes a project with the default template for the "quarkus" stack in the "experimental" repository.
```
neeraj-laad: This is incorrect. When the user specifies 'none' we do not provide any template application.

cmd/init.go
```diff
@@ -48,18 +48,27 @@ func newInitCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
-If keyword "none" is specified instead of a [template], the project will be initialized to use Appsody, and no template will be provided.
-
-Without the [stack] argument, this command must be run on an existing Appsody project and will only run the stack init script to
-setup the local dev environment.`,
+		Short: "Initialize an Appsody project.",
+		Long: `Set up the local Appsody development environment. You can do this for an existing project or use the template application provided by the stack. 
+
+By default, the command creates an Appsody stack configuration file and provides a simple default application. You can also choose to initialize a project with a different template application, or no template at all. 
+
+If you run this command in a directory that is not empty, and you want to initialise a project with a template application, the command fails unless you specify the "overwrite" option [--overwrite].
```
helenmasters: initialize (US spelling)

cmd/init.go
```diff
@@ -48,18 +48,27 @@ func newInitCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
-Use 'appsody list' to see the available stack options.
-
-If keyword "none" is specified instead of a [template], the project will be initialized to use Appsody, and no template will be provided.
-
-Without the [stack] argument, this command must be run on an existing Appsody project and will only run the stack init script to
-setup the local dev environment.`,
+		Short: "Initialize an Appsody project.",
+		Long: `Set up the local Appsody development environment. You can do this for an existing project or use the template application provided by the stack. 
+
+By default, the command creates an Appsody stack configuration file and provides a simple default application. You can also choose to initialize a project with a different template application, or no template at all.
```
helenmasters: the words "at all" are not needed after "or no template"

cmd/init.go
helenmasters: This sentence is back to front, we need something like...
"To initialize a project with a template application, in a directory that is not empty, you need to specify the overwrite option."

PR #620, opened 2019-11-12
Updates to readme, release, and build MD files

### LINKED ISSUE #619: Update references to the controller in the readme file
Appsody has a new approach to packaging and provisioning the controller. The readme and associated documents need to reflect this new approach.

Ensure that the readme and related documents accurately describe the new architecture.

chilanti:
References to the controller were updated, and the related environment variables also documented.
Fixes #619


build.md
```diff
@@ -7,15 +7,24 @@ Prerequisites:
...
-After setting the `GOPATH` env var correctly, just run `make <action...>` from the command line, within the same directory where `Makefile` resides. For example `make package clean` will run the `package` and then the `clean` actions.
+After setting the `GOPATH` env var correctly, just run `make <action...>` from the command line, within the same directory where `Makefile` resides. 
+
+For example - if you want to obtain a fully built Appsody CLI for all the supported platform run:
+```
+make clean package
+```
+which will run the `clean` action, and then the `package` action. The latter produces the built binaries (as described in the next section).
+
+If you inspect the `Makefile`, you'll notice that it invokes a number of scripts - and you'll notice a few conditional paths, because certain Linux commands behave differently on OS/X and elsewhere.
```
helenmasters: Minor comment - wherever we refer to OS/X in our docs/READMEs we should update it to be macOS

build.md
```diff
@@ -33,22 +42,31 @@ Here's a description of the various artifacts as you would see them in a release
...
-
 * If you are replacing an old installation, you may want to delete the appsody home directory (`$HOME/.appsody`).
 
+### The Appsody Controller
+In order to enable rapid app development, the Appsody CLI relies on a component called the `Appsody controller`, which is built separately, and which is meant to be run within the container that hosts the Appsody app itself (as opposed to being executed on the developers' system). However, the good news is that the Appsody CLI automatically downloads the Appsody controller when necessary, in the form of a Docker image.
+
+When the Appsody CLI is built using the `Makefile`, the CLI "knows" which version of the controller needs to be obtained. If you just compile the Appsody CLI binary using `go build`, the Appsody CLI will pull the `latest` version of the controller image (which may or may not be a good match - so, be aware of that).
+
+As a developer, you can also override which version of the Appsody controller the CLI uses, in two different ways:
+1) By setting the `APPSODY_CONTROLLER_VERSION` env var prior to launching the Appsody CLI. For example, `export APPSODY_CONTROLLER_VERSION=0.3.0`
```
helenmasters: Are we deliberately not documenting these within the env vars on the website...?

build.md
chilanti: Done.

build.md
chilanti: I assume you refer to these: https://appsody.dev/docs/stacks/environment-variables
Well... those are different env vars, in that they are consumed by the stacks. The env vars described here (and there are others, which we haven't documented since they cater to very specific audiences like Codewind) are consumed by the the CLI. I'd keep them separate - definitely we don't want to document in the same place as the stack env vars.
We had some discussions on how, where, and if we should document the CLI env vars...

PR #621, opened 2019-11-12
Moving app-deploy generation to build and dropping knative support

EnriqueL8:
This PR:
- Drops the generation of a sample knative template
- Adds `pull-url` flag to the `build`command
- Moves the generation of the `app-deploy.yaml` to the build command
- Marks `--force` and `--generate-only` as deprecated on the deploy command
- Uses the `AppsodyApplication` struct from the operator, instead of creating our own
- Adds basic tests for the `app-deploy.yaml`and flags


kylegc:
I don't think we need `--force` anymore. But you should leave it on the deploy command for now, and mark it as DEPRECATED. That way existing users of this flag won't get an error.

Also, the deploy `--generate-only` flag should be marked as deprecated.

EnriqueL8:
@kylegc  I'll make those chan
@kewegner I agree extensive testing should be carried out. Good catch on the tag argument, I'll pass the value through `findNamespaceRepositoryAndTag`?

Issue #626, opened 2019-11-13
Use one struct for holding stack.yaml data


Kamran64:
**Is your feature request related to a problem? Please describe.**
Across the repo, we have several different structs holding `stack.yaml` data - E.g. Within `stack_yaml_lint.go` and `stack_package.go` and possibly other places.

**Describe the solution you'd like**
Would it be possible to reuse one struct across the entire repo? cc @appsody/cli - Is this change easy enough/worth it?

Kamran64:
I think only the `stack_yaml_lint.go` and `stack_package.go` use it. I'll update the Linter code to use the struct from `stack_package`

Issue #625, opened 2019-11-13
Update the linter to check for valid templating values


Kamran64:
**Is your feature request related to a problem? Please describe.**
When the work done by: https://github.com/appsody/appsody/pull/528 goes in, we should consider updating the `stack.yaml` linter to check for valid key:value pairs. 

**Describe the solution you'd like**
Add a check to the linter, that templating is used. If not, warn the user (but don't error), to encourage more stack creators/contributors to utilise this capability within their stack. If they have added it here, we can validate that the keys and values specified are valid for the stack (@CameronMcWilliam - Is there a strict naming requirement for what keys should be set as?)

CameronMcWilliam:
Keys and values must be strings (for now) and keys must begin with an alphanumeric character.

PR #627, opened 2019-11-13
Update the linter to check for custom template variables and other misc fixes

Kamran64:
Fixes: https://github.com/appsody/appsody/issues/625
Fixes: https://github.com/appsody/appsody/issues/626

- Checks that custom templating variables have been used in the stack. If not, the user will be warned. If they have, and the the keys or values do not match an alphanumeric format, the Linter will throw an error.

- Removes the struct defined in the `stack_yaml_lint.go` to use the `stack_package.go` struct

- Removes the function to check for the default template. This has now been achieved by adding a field to the struct (which for some reason wasn't working previously!)

- Adds a function to `utils.go` to check the input meets the format for semver (And removes the existing version check function in the Linter)


kylegc:
@Kamran64 Please resolve the merge conflicts in this PR. Thanks

Kamran64:
@kylegc - Resolved :)

Kamran64:
Sorry @kylegc - I've just had to fix a few more merge conflicts since there's been a few changes to these files in the past week and my PR rewrites a couple of stuff. Should be ready for review now thanks.

BarDweller:
why are custom template values limited to alphanumeric regex? won't this break the planned version range value I'm using over in 517/529 (stacks repo prs)?

Kamran64:
Thanks @BarDweller - You're right. I just fixed this so that the check is only enforced on keys and not values.

PR #634, opened 2019-11-14
Unit tests for templating

CameronMcWilliam:
2 simple tests that use the new templating features of `stack package`.  I planl to add more to these in the future, but it would be good if we can get a couple tests in for this release just to demonstrate the new feature.


cmd/stack_package_test.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+
+	// read the whole file at once
+	b, err := ioutil.ReadFile(stackPath + "/templating.txt")
+	if err != nil {
+		panic(err)
+	}
+	s := string(b)
+	if !strings.Contains(s, "id: starter") {
+		t.Fatalf("Templating text did not match expected values: %v", err)
```
kylegc: Where's `err` coming from here?

cmd/stack_package_test.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+
+	// read the whole file at once
+	b, err := ioutil.ReadFile(stackPath + "/templating.txt")
+	if err != nil {
+		panic(err)
+	}
+	s := string(b)
+	if !strings.Contains(s, "id: starter, name: Starter Sample, version: 0.1.1, description: Runnable starter stack, copy to create a new stack, tag: dev.local/starter:SNAPSHOT, maintainers: Henry Nash <henry.nash@uk.ibm.com>, semver.major: 0, semver.minor: 1, semver.patch: 1, semver.majorminor: 0.1, image.namespace: dev.local, customvariable1: value1, customvariable2: value2") {
+		t.Fatal("Templating text did not match expected values")
```
kylegc: You should `t.Log()` the `s` variable so when we are debugging the error we know what the string contained.

cmd/stack_package_test.go
```diff
@@ -0,0 +1,300 @@
...
+	// apply templating to stack
+	err = cmd.ApplyTemplating(projectPath, stackPath, templateMetadata)
+	if err != nil {
+		t.Fatalf("Error applying template: %v", err)
+	}
+
+	// read the whole file at once
+	b, err := ioutil.ReadFile(stackPath + "/templating.txt")
+	if err != nil {
+		panic(err)
```
kylegc: We should avoid panics

cmd/stack_package_test.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+	s := string(b)
+	if !strings.Contains(s, "id: <no value>") {
+		t.Fatalf("Templating text did not match expected values: %v", err)
```
kylegc: Where's `err` coming from here?

Issue #577, opened 2019-11-05
Standardize 'appsody deploy' help text format 


skoh7645:
Apply standard created for CLI commands' help text under issue #463, to the commands:
- [x] `appsody deploy`
- [x] `appsody deploy delete`

PR #635, opened 2019-11-14
deploy: refactor help text

skoh7645:
Reformatted help text for the following commands:
-  `appsody deploy`
-  `appsody deploy delete`

Fixes: #577


cmd/deploy.go
```diff
@@ -99,9 +99,16 @@ func newDeployCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
 	var deployCmd = &cobra.Command{
 		Use:   "deploy",
-		Short: "Build and deploy your Appsody project to your Kubernetes cluster",
-		Long: `This command extracts the code from your project, builds a local Docker image for deployment,
-generates a deployment manifest (yaml) file if one is not present, and uses it to deploy your image to a Kubernetes cluster, either via the Appsody operator or as a Knative service.`,
+		Short: "Build and deploy your Appsody project to Kubernetes.",
+		Long: `Build and deploy a local container image of your Appsody project to your Kubernetes cluster. 
+		
+The stack, along with your Appsody project, is extracted to a local directory before the container image of your Appsody project is built for deployment.
+The command generates a deployment manifest file, "app-deploy.yaml", if one is not present. The CLI uses this file to deploy your image to your Kubernetes cluster, either via the Appsody operator or as a Knative service.`,
```
kewegner: I wonder if we should change the wording about Knative Service here, as of milestone 10 we will always use the appsody operator.   The --knative option determines if the appsody-operator deploys the application as a Knative Service.

cmd/deploy_delete.go
```diff
@@ -23,8 +23,15 @@ func newDeleteDeploymentCmd(deployConfig *deployCommandConfig) *cobra.Command {
...
 	var deleteDeploymentCmd = &cobra.Command{
 		Use:   "delete",
-		Short: "Delete your deployed Appsody project from a Kubernetes cluster",
-		Long:  `This command deletes your deployed Appsody project from the configured Kubernetes cluster using your existing deployment manifest.`,
+		Short: "Delete your deployed Appsody project from a Kubernetes cluster.",
+		Long: `Delete your deployed Appsody project from the configured Kubernetes cluster using your existing deployment manifest.
+
+By default, the command looks for the deployed project in the "default" namespace, and uses the generated "app-deploy.yaml" deployment manifest, unless you specify otherwise.`,
+		Example: `  appsody deploy delete -f my-deploy.yaml
+  Deletes the pod using the type and name specified in the "my-deploy.yaml" deployment manifest, in the "default" namespace.
```
kewegner: Should we say deployment rather than deletes the pod?

cmd/deploy.go
skoh7645: @kewegner How about "The CLI uses this file to deploy your image to your Kubernetes cluster via the Appsody operator, and optionally as a Knative service if you specify the '--knative' flag."?

kewegner: Yes I think that wording is OK.

cmd/deploy.go
```diff
@@ -49,9 +49,16 @@ func newDeployCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
 	var deployCmd = &cobra.Command{
 		Use:   "deploy",
-		Short: "Build and deploy your Appsody project to your Kubernetes cluster",
-		Long: `This command extracts the code from your project, builds a local Docker image for deployment,
-generates a deployment manifest (yaml) file if one is not present, and uses it to deploy your image to a Kubernetes cluster, either via the Appsody operator or as a Knative service.`,
+		Short: "Build and deploy your Appsody project to Kubernetes.",
+		Long: `Build and deploy a local container image of your Appsody project to your Kubernetes cluster. 
+		
+The stack, along with your Appsody project, is extracted to a local directory before the container image of your Appsody project is built for deployment.
+The command generates a deployment manifest file, "app-deploy.yaml", if one is not present. The CLI uses this file to deploy your image to your Kubernetes cluster via the Appsody operator, and optionally as a Knative service if you specify the "--knative" flag.`,
```
Kamran64: This part is now done as part of `appsody build` - I suppose since deploy does a build under the covers this is still technically correct though. However, does the new help text for `appsody build` reference this?

kylegc: Correct. I think all of this is fine to keep here. Somewhere we may want to say that this runs an `appsody build` first, then applies the `app-deploy.yaml` to the cluster. 

The `appsody build` help text probably needs review.

cmd/deploy.go
```diff
@@ -49,9 +49,16 @@ func newDeployCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
+		Example: `  appsody deploy --namespace my-namespace
+  Builds and deploys your project to the "my-namespace" namespace in your local Kubernetes cluster.
+  
+  appsody deploy -t my-repo/nodejs-express --push-url external-registry-url --pull-url internal-registry-url
+  Builds the container image, sets the namespace to "my-repo", sets the name to "nodejs-express", pushes image to "external-registry-url/my-repo/nodejs-express", and creates a deployment manifest that tells the K8S cluster to pull the image from "internal-registry-url/my-repo/nodejs-express".`,
```
kylegc: I think it reads better if you replace
`Builds the container image, sets the namespace to "my-repo", sets the name to "nodejs-express",`
with the following
`Builds and tags the image as "my-repo/nodejs-express",`

cmd/deploy.go
kylegc: We should probably also add something about the appsody operator. The operator is installed and used for the deployment.

skoh7645: @kylegc I have made the requested changes to this command. Do we need to mention more about the Appsody Operator in this command past "The CLI uses this file to deploy your image to your Kubernetes cluster via the Appsody Operator...."?

cmd/deploy_delete.go
```diff
@@ -23,8 +23,15 @@ func newDeleteDeploymentCmd(deployConfig *deployCommandConfig) *cobra.Command {
...
-		Long:  `This command deletes your deployed Appsody project from the configured Kubernetes cluster using your existing deployment manifest.`,
+		Short: "Delete your deployed Appsody project from a Kubernetes cluster.",
+		Long: `Delete your deployed Appsody project from the configured Kubernetes cluster using your existing deployment manifest.
+
+By default, the command looks for the deployed project in the "default" namespace, and uses the generated "app-deploy.yaml" deployment manifest, unless you specify otherwise.`,
+		Example: `  appsody deploy delete -f my-deploy.yaml
+  Deletes the AppsodyApplication using the name and type specified in the "my-deploy.yaml" deployment manifest, in the "default" namespace.
+  
+  appsody deploy delete --namespace my-namespace
+  Deletes the AppsodyApplication using the name and type specified in the "app-deploy.yaml" deployment manifest, in the "my-namespace" namespace.`,
```
kylegc: `Deletes the AppsodyApplication from the "my-namespace" namespace using the name and type specified in the "app-deploy.yaml" deployment manifest.

cmd/deploy_delete.go
```diff
@@ -45,6 +52,6 @@ func newDeleteDeploymentCmd(deployConfig *deployCommandConfig) *cobra.Command {
 		},
 	}
 
-	deleteDeploymentCmd.PersistentFlags().StringVarP(&deployConfigFile, "file", "f", "app-deploy.yaml", "The file name to use for the deployment configuration.")
+	deleteDeploymentCmd.PersistentFlags().StringVarP(&deployConfigFile, "file", "f", "app-deploy.yaml", "Name of the deployment configuration you want to use.")
```
kylegc: I liked having "file" in here because it must be a valid filename. So maybe we use
`Filename of the deployment configuration you want to use`

cmd/deploy.go
```diff
@@ -49,9 +49,16 @@ func newDeployCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
 	var deployCmd = &cobra.Command{
 		Use:   "deploy",
-		Short: "Build and deploy your Appsody project to your Kubernetes cluster",
-		Long: `This command extracts the code from your project, builds a local Docker image for deployment,
-generates a deployment manifest (yaml) file if one is not present, and uses it to deploy your image to a Kubernetes cluster, either via the Appsody operator or as a Knative service.`,
+		Short: "Build and deploy your Appsody project to Kubernetes.",
+		Long: `Build and deploy a local container image of your Appsody project to your Kubernetes cluster. 
+		
+The command extracts the stack, along with your Appsody project to a local directory. The extracted project is used to build the container image for deployment using 'appsody build'.
+A deployment manifest file, "app-deploy.yaml", is generated if one is not present, and applied to the cluster. The CLI uses this file to deploy your image to your Kubernetes cluster via the Appsody Operator, and optionally as a Knative service if you specify the "--knative" flag.`,
```
kylegc: After this line, I would add something like:
`If the Appsody Operator cannot be found, it will be installed on your cluster.`

cmd/deploy.go
kylegc: As I said in the other comments, I think we should include a simple statement that says if the operator doesn't exist, it will be installed.

cmd/deploy_delete.go
```diff
@@ -23,8 +23,15 @@ func newDeleteDeploymentCmd(deployConfig *deployCommandConfig) *cobra.Command {
 	var deployConfigFile string
 	var deleteDeploymentCmd = &cobra.Command{
 		Use:   "delete",
-		Short: "Delete your deployed Appsody project from a Kubernetes cluster",
-		Long:  `This command deletes your deployed Appsody project from the configured Kubernetes cluster using your existing deployment manifest.`,
+		Short: "Delete your deployed Appsody project from a Kubernetes cluster.",
+		Long: `Delete your deployed Appsody project from the configured Kubernetes cluster using your existing deployment manifest.
```
doveye: Needs a comma after "cluster", otherwise could be read as "the configured Kubernetes cluster that uses your existing deployment manifest":
"Delete your deployed Appsody project from the configured Kubernetes cluster, using your existing deployment manifest."

cmd/deploy_delete.go
```diff
@@ -23,8 +23,15 @@ func newDeleteDeploymentCmd(deployConfig *deployCommandConfig) *cobra.Command {
...
+
+By default, the command looks for the deployed project in the "default" namespace, and uses the generated "app-deploy.yaml" deployment manifest, unless you specify otherwise.`,
```
doveye: I'd remove the comma in the middle, to clarify that "unless you specify otherwise" applies to both the namespace and the deployment manifest:
"By default, the command looks for the deployed project in the "default" namespace and uses the generated "app-deploy.yaml" deployment manifest, unless you specify otherwise."

cmd/deploy_delete.go
```diff
@@ -23,8 +23,15 @@ func newDeleteDeploymentCmd(deployConfig *deployCommandConfig) *cobra.Command {
...
+		Example: `  appsody deploy delete -f my-deploy.yaml
+  Deletes the AppsodyApplication from the "default" namespace using the name and type specified in the "my-deploy.yaml" deployment manifest.
```
doveye: Needs a comma before "using" again, for same reason as previously. Also, presumably we're talking about an "Appsody application" rather than an "AppsodyApplication" (or is the latter valid?):
"Deletes the Appsody application from the "default" namespace, using the name and type specified in the "my-deploy.yaml" deployment manifest."

cmd/deploy_delete.go
```diff
@@ -23,8 +23,15 @@ func newDeleteDeploymentCmd(deployConfig *deployCommandConfig) *cobra.Command {
...
+  
+  appsody deploy delete --namespace my-namespace
+  Deletes the AppsodyApplication from the "my-namespace" namespace using the name and type specified in the "app-deploy.yaml" deployment manifest`,
```
doveye: Same comments as earlier, plus needs a full stop at the end:
"Deletes the Appsody application from the "my-namespace" namespace, using the name and type specified in the "app-deploy.yaml" deployment manifest."

cmd/deploy_delete.go
```diff
@@ -45,6 +52,6 @@ func newDeleteDeploymentCmd(deployConfig *deployCommandConfig) *cobra.Command {
 		},
 	}
 
-	deleteDeploymentCmd.PersistentFlags().StringVarP(&deployConfigFile, "file", "f", "app-deploy.yaml", "The file name to use for the deployment configuration.")
+	deleteDeploymentCmd.PersistentFlags().StringVarP(&deployConfigFile, "file", "f", "app-deploy.yaml", "Filename of the deployment configuration you want to use.")
```
doveye: We've used "deployment manifest" before, so should this be the same? We don't need "to use", because that's implied, but it might be worth explaining a bit about what it's used for, eg
"The name of the deployment manifest file for your application." or "Your application's deployment manifest." (if either of those make technical sense!)

cmd/deploy_delete.go
doveye: According to the docs, you have to run this command from your Appsody project directory, so that would be worth adding to the help?

skoh7645: I have added this now, thanks

cmd/deploy_delete.go
skoh7645: In this case the latter is valid as the resource is of kind `AppsodyApplication`

skoh7645: I have added the comma :)

cmd/deploy.go
```diff
@@ -49,9 +49,19 @@ func newDeployCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
 	var deployCmd = &cobra.Command{
 		Use:   "deploy",
-		Short: "Build and deploy your Appsody project to your Kubernetes cluster",
-		Long: `This command extracts the code from your project, builds a local Docker image for deployment,
-generates a deployment manifest (yaml) file if one is not present, and uses it to deploy your image to a Kubernetes cluster, either via the Appsody operator or as a Knative service.`,
+		Short: "Build and deploy your Appsody project to Kubernetes.",
+		Long: `Build and deploy a local container image of your Appsody project to your Kubernetes cluster. 
+		
+The command extracts the stack, along with your Appsody project to a local directory. The extracted project is used to build the container image for deployment using 'appsody build'.
+A deployment manifest file, "app-deploy.yaml", is generated if one is not present, and applied to the cluster. The CLI uses this file to deploy your image to your Kubernetes cluster via the Appsody Operator, and optionally as a Knative service if you specify the "--knative" flag.
```
doveye: Can you use steps in Go? I think they would be clearer, eg:

> The command performs the following steps:
> 1. Extracts the stack, along with your Appsody project, to a local directory.
> 2. Runs the `appsody build` command to build the container image for deployment.
> 3. Generates a deployment manifest file, "app-deploy.yaml", if one is not present, then applies it to the cluster. 
> 4. Deploys your image to your Kubernetes cluster via the Appsody operator, and optionally as a Knative service if you specify the "--knative" flag.

Just to check a couple of things:
- Does it makes sense to say a deployment manifest is "applied to a cluster"?
- If you specify --knative, do you get the Knative service **in addition** to the normal deployment? (That's how I interpret the above content.)

cmd/deploy.go
```diff
@@ -49,9 +49,19 @@ func newDeployCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
+
+
+If the Appsody Operator cannot be found, it will be installed on your cluster.`,
+		Example: `  appsody deploy --namespace my-namespace
+  Builds and deploys your project to the "my-namespace" namespace in your local Kubernetes cluster.
+  
+  appsody deploy -t my-repo/nodejs-express --push-url external-registry-url --pull-url internal-registry-url
+  Builds and tags the image as "my-repo/nodejs-express", pushes image to "external-registry-url/my-repo/nodejs-express", and creates a deployment manifest that tells the Kubernetes cluster to pull the image from "internal-registry-url/my-repo/nodejs-express".`,
```
doveye: Needs a "the" in "pushes image"

cmd/deploy.go
```diff
@@ -49,9 +49,19 @@ func newDeployCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
-		Long: `This command extracts the code from your project, builds a local Docker image for deployment,
-generates a deployment manifest (yaml) file if one is not present, and uses it to deploy your image to a Kubernetes cluster, either via the Appsody operator or as a Knative service.`,
+		Short: "Build and deploy your Appsody project to Kubernetes.",
+		Long: `Build and deploy a local container image of your Appsody project to your Kubernetes cluster. 
+		
+The command extracts the stack, along with your Appsody project to a local directory. The extracted project is used to build the container image for deployment using 'appsody build'.
+A deployment manifest file, "app-deploy.yaml", is generated if one is not present, and applied to the cluster. The CLI uses this file to deploy your image to your Kubernetes cluster via the Appsody Operator, and optionally as a Knative service if you specify the "--knative" flag.
+
+
+If the Appsody Operator cannot be found, it will be installed on your cluster.`,
```
doveye: You could add this to step 4, above:
> 4. Deploys your image to your Kubernetes cluster via the Appsody operator, and optionally as a Knative service if you specify the "--knative" flag. If an Appsody operator cannot be found, one is installed on your cluster.

cmd/deploy.go
skoh7645: Steps in Go look good (I've tested and it translates well to the website as well). I would say maybe I could make "then applies it to the cluster" by changing it to "then applies it to your Kubernetes cluster", but it does make sense to use the word "apply" in this case. I will chase up the answer to your second question

skoh7645: @doveye it is knative *instead of* normal

cmd/deploy.go
```diff
@@ -52,16 +52,17 @@ func newDeployCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
-The command extracts the stack, along with your Appsody project to a local directory. The extracted project is used to build the container image for deployment using 'appsody build'.
-A deployment manifest file, "app-deploy.yaml", is generated if one is not present, and applied to the cluster. The CLI uses this file to deploy your image to your Kubernetes cluster via the Appsody Operator, and optionally as a Knative service if you specify the "--knative" flag.
+The command performs the following steps:
 
-
-If the Appsody Operator cannot be found, it will be installed on your cluster.`,
+1. Extracts the stack, along with your Appsody project, to a local directory.
+2. Runs the appsody build command to build the container image for deployment.
+3. Generates a deployment manifest file, "app-deploy.yaml", if one is not present, then applies it to your Kubernetes cluster.
+4. Deploys your image to your Kubernetes cluster via the Appsody operator, or optionally as a Knative service if you specify the "--knative" flag. If the Appsody Operator cannot be found, it will be installed on your cluster.`,
```
doveye: Need lowercase "o" for operator. I don't think we need "optionally" now that we know it's a mutually exclusive thing. One more minor thing - is it "the" operator, or "an" operator? The docs suggest you can have multiple operators?
"4. Deploys your image to your Kubernetes cluster via the Appsody operator, or as a Knative service if you specify the "--knative" flag. If the Appsody operator cannot be found, it will be installed on your cluster."

skoh7645: You can have multiple operators. I've changed this now

cmd/deploy.go
```diff
@@ -49,9 +49,20 @@ func newDeployCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
 		Use:   "deploy",
-		Short: "Build and deploy your Appsody project to your Kubernetes cluster",
-		Long: `This command extracts the code from your project, builds a local Docker image for deployment,
-generates a deployment manifest (yaml) file if one is not present, and uses it to deploy your image to a Kubernetes cluster, either via the Appsody operator or as a Knative service.`,
+		Short: "Build and deploy your Appsody project to Kubernetes.",
+		Long: `Build and deploy a local container image of your Appsody project to your Kubernetes cluster. 
+		
+The command performs the following steps:
+
+1. Extracts the stack, along with your Appsody project, to a local directory.
```
EnriqueL8: Do we need to specify that the `extract` step happens? Maybe this should instead be specified in the build help text

skoh7645: I'll remove this as we already mention the `extract` in the `build` help text. Thanks

PR #639, opened 2019-11-15
Rework log output

djones6:
Appsody CLI currently has a global logging configuration.  This is incompatible with running multiple commands concurrently during testing, which is something we'd like to do in the future (parallel testing).

This PR moves the logging configuration into a separate `LoggingConfig` struct, that is nested within `RootCommandConfig`.  Every log statement now uses the `config.<Level>.Log()` function rather than a global function.  Functions that did not already receive a `Config` of some type are now passed a `LoggingConfig` explicitly.

~~I have also replaced all uses of `fmt.Print` with `t.Log` in the test code.  Similarly, `t` has been propagated to all test utility functions that need to print.  This ensures that test output appears in chronological order.~~
Moved to https://github.com/appsody/appsody/pull/657

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -179,25 +179,25 @@ func build(config *buildCommandConfig) error {
 	}
 
 	cmdArgs = append(cmdArgs, "-f", dockerfile, extractDir)
-	Debug.log("final cmd args", cmdArgs)
+	config.Debug.log("final cmd args", cmdArgs)
 	var execError error
 	if !config.Buildah {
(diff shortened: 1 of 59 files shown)


kylegc:
I've split this into two PRs: one for refactoring logging in the cmd package, and one for refactoring tests to use `t.log`. 
The reason behind this is to confirm the changes to cmd don't break the existing test suite. This PR will contain the cmd changes. I'll create a follow-up PR to integrate the test changes.

Issue #586, opened 2019-11-05
Standardize 'appsody stop' help text format


skoh7645:
Apply standard created for CLI commands' help text under issue #463, to the `appsody stop` command

PR #641, opened 2019-11-15
appsody stop: refactor help text

skoh7645:
Refactor help text for the `appsody stop` command

Fixes: #586


cmd/stop.go
```diff
@@ -23,12 +23,16 @@ func newStopCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
-		Short: "Stops the local Appsody docker container for your project",
-		Long: `Stop the local Appsody docker container for your project.
+		Short: "Stop the local running Appsody container.",
+		Long: `Stop the local running Appsody container for your project.
 
-Stops the docker container specified by the --name flag. 
-If --name is not specified, the container name is determined from the current working directory (see default below).
-To see a list of all your running docker containers, run the command "docker ps". The name is in the last column.`,
+By default, the command stops the Appsody container that was launched from the project in your current working directory. 
+To see a list of all your running Appsody containers, run the command "appsody ps". The names of you running Appsody projects are in the second column.`,
```
Kamran64: nit: you --> your

cmd/stop.go
```diff
@@ -23,12 +23,16 @@ func newStopCmd(rootConfig *RootCommandConfig) *cobra.Command {
 	// stopCmd represents the stop command
 	var stopCmd = &cobra.Command{
 		Use:   "stop",
-		Short: "Stops the local Appsody docker container for your project",
-		Long: `Stop the local Appsody docker container for your project.
+		Short: "Stop the local running Appsody container.",
+		Long: `Stop the local running Appsody container for your project.
```
doveye: Think some clarity is needed here, but not sure what! Either:
"Stop the local, running Appsody container for your project."  (ie, it's local and it's running)
or
"Stop the local-running Appsody container for your project."  (ie, it's running locally, as opposed to something running remotely)

cmd/stop.go
```diff
@@ -23,12 +23,16 @@ func newStopCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
 
-Stops the docker container specified by the --name flag. 
-If --name is not specified, the container name is determined from the current working directory (see default below).
-To see a list of all your running docker containers, run the command "docker ps". The name is in the last column.`,
+By default, the command stops the Appsody container that was launched from the project in your current working directory. 
+To see a list of all your running Appsody containers, run the command "appsody ps". The names of your running Appsody projects are in the second column.`,
```
doveye: Is it not obvious which column contains the names? Could be risky saying which column they're in, in case it changes (as it already has, judging by the original text!) If it's not obvious, does the column have a heading that you could use instead? Also, "containers" and "projects" are both used to mean the same thing in one sentence - suggest changing "projects" to "containers". eg:
"The container names are shown in the "blah" column."

skoh7645: It is obvious which column contains the names, I have removed this sentence

Issue #578, opened 2019-11-05
Standardize 'appsody extract' help text format


skoh7645:
Apply standard created for CLI commands' help text under issue #463, to the `appsody extract` command

PR #642, opened 2019-11-15
extract: refactor help text

skoh7645:
Refactor help text for the appsody extract` command

Fixes: #578


cmd/extract.go
```diff
@@ -37,9 +37,15 @@ func newExtractCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
-		Long: `This copies the full project, stack plus app, into a local directory
-in preparation to build the final container image.`,
+		Short: "Extract your Appsody project to a local directory.",
+		Long: `Extract the full application (the stack and your Appsody project), into a local directory, in preparation to build the final container image.
+		
+By default, your project gets extracted into your local '$HOME/.appsody/extract' directory, unless specified otherwise.`,
+		Example: `  appsody extract --target-dir $HOME/extract/directory
+  Extracts your project from the container to the local '$HOME/extract/directory' on your local system.
+  
+  appsody extract --name my-container
```
kylegc: I'd say remove this example. The `--name` option will very rarely be used so it will probably just confuse people.

cmd/extract.go
```diff
@@ -37,9 +37,12 @@ func newExtractCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
 		Use:   "extract",
-		Short: "Extract the stack and your Appsody project to a local directory",
-		Long: `This copies the full project, stack plus app, into a local directory
-in preparation to build the final container image.`,
+		Short: "Extract your Appsody project to a local directory.",
+		Long: `Extract the full application (the stack and your Appsody project), into a local directory, in preparation to build the final container image.
+		
+By default, your project gets extracted into your local '$HOME/.appsody/extract' directory, unless specified otherwise.`,
+		Example: `  appsody extract --target-dir $HOME/my-extract/directory
+  Extracts your project from the container to the local '$HOME/my-extract/directory' on your local system.`,
```
helenmasters: Two local's in one sentence is confusing I'm not sure which is the most appropriate to remove but if it's a local directory it must be on your local system... so perhaps the second occurrence should be removed?

cmd/extract.go
```diff
@@ -37,9 +37,12 @@ func newExtractCmd(rootConfig *RootCommandConfig) *cobra.Command {
 
 	var extractCmd = &cobra.Command{
 		Use:   "extract",
-		Short: "Extract the stack and your Appsody project to a local directory",
-		Long: `This copies the full project, stack plus app, into a local directory
-in preparation to build the final container image.`,
+		Short: "Extract your Appsody project to a local directory.",
+		Long: `Extract the full application (the stack and your Appsody project), into a local directory, in preparation to build the final container image.
```
helenmasters: in preparation to build? that wording doesn't read very easily. Please can you work with @doveye on this wording and the wording for the flag '--target-dir'?

The 'build' flag if it's going to reference docker needs to refer to Docker with a capital D.

cmd/extract.go
```diff
@@ -37,9 +37,12 @@ func newExtractCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
+		
+By default, your project gets extracted into your local '$HOME/.appsody/extract' directory, unless specified otherwise.`,
```
doveye: You don't need "By default" and "unless specified otherwise", as they basically say the same thing. How about:
"Your project is extracted into your local '$HOME/.appsody/extract' directory, unless you use the `--target-dir` flag to specify a different location."

cmd/extract.go
doveye: I agree, it makes me think you can have local and remote directories on your local system.

cmd/extract.go
doveye: You don't need the comma after the second bracket. I don't understand the "preparation to build" bit - is it explaining why you would want to extract a project? The docs say the build command does an extract anyway, so why would you want to do one separately?

skoh7645: @doveye I included it there to give some context about when Appsody will use this. I will remove it from here as we say it in the `build` command. Thanks

cmd/extract.go
```diff
@@ -38,18 +38,18 @@ func newExtractCmd(rootConfig *RootCommandConfig) *cobra.Command {
...
-  Extracts your project from the container to the local '$HOME/my-extract/directory' on your local system.`,
+  Extracts your project from the container to the local '$HOME/my-extract/directory' on your system.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return extract(config)
 		},
 	}
 
-	extractCmd.PersistentFlags().StringVar(&config.targetDir, "target-dir", "", "Directory path to place the extracted files. This dir must not exist, it will be created.")
-	extractCmd.PersistentFlags().BoolVar(&rootConfig.Buildah, "buildah", false, "Extract project using buildah primitives instead of docker.")
+	extractCmd.PersistentFlags().StringVar(&config.targetDir, "target-dir", "", "Directory path to place the extracted files. This directory must not exist, it will be created.")
```
doveye: Suggest:
"The directory to extract the files into. This directory must not exist, as it will be created."
Might be worth mentioning if the directory path has to be absolute, or if it is relative to the current dir or somewhere else?

skoh7645: updated it to include it should be the absolute directory path

Issue #581, opened 2019-11-05
Standardize 'appsody operator' help text format


skoh7645:
Apply standard created for CLI commands' help text under issue #463, to the commands:
- [ ] `appsody operator install`
- [ ] `appsody operator uninstall`

PR #643, opened 2019-11-15
appsody operator: refactor help text

skoh7645:
Refactor help text for the following commands:
- `appsody operator install`
- `appsody operator uninstall`

Fixes: #581


cmd/operator_install.go
```diff
@@ -33,8 +33,12 @@ func newOperatorInstallCmd(operatorConfig *operatorCommandConfig) *cobra.Command
 	// installCmd represents the "appsody deploy install" command
 	var installCmd = &cobra.Command{
 		Use:   "install",
-		Short: "Install the Appsody Operator into the configured Kubernetes cluster",
-		Long:  ``,
+		Short: "Install the Appsody Operator.",
+		Long: `Install the Appsody Operator into your configured Kubernetes cluster.
+		
+The Appsody Operator deploys an Operator container with the controller, and listens for incoming AppsodyApplication resources.`,
```
kylegc: This line doesn't sound quite right. Maybe try something like:
`The Appsody Operator listens for incoming AppsodyApplication resources on your cluster. For more information, see https://operatorhub.io/operator/appsody-operator`

kylegc: We should also clarify that unless the `--watch-all` flag us used, the operator only watches one namespace, so additional operators must be installed to watch additional namespaces.

cmd/operator_install.go
```diff
@@ -33,8 +33,12 @@ func newOperatorInstallCmd(operatorConfig *operatorCommandConfig) *cobra.Command
...
+		Example: `  appsody operator install --watchspace my-namespace
```
kylegc: We should include `--namespace` in the example, in addition to `--watchspace`
namespace is where the operator will be installed
watchspace is where the operator will watch for AppsodyApplication resources

cmd/operator_uninstall.go
```diff
@@ -35,8 +35,8 @@ func newOperatorUninstallCmd(operatorConfig *operatorCommandConfig) *cobra.Comma
 	// uninstallCmd represents the "appsody deploy uninstall" command
 	var uninstallCmd = &cobra.Command{
 		Use:   "uninstall",
-		Short: "Uninstall the Appsody Operator from the configured Kubernetes cluster",
-		Long:  ``,
+		Short: "Uninstall the Appsody Operator.",
+		Long:  `Uninstall the Appsody Operator from your configured Kubernetes cluster.`,
```
kylegc: Should include an example with the `--namespace` flag

cmd/operator_uninstall.go
```diff
@@ -37,6 +37,8 @@ func newOperatorUninstallCmd(operatorConfig *operatorCommandConfig) *cobra.Comma
 		Use:   "uninstall",
 		Short: "Uninstall the Appsody Operator.",
 		Long:  `Uninstall the Appsody Operator from your configured Kubernetes cluster.`,
+		Example: `  appsody operator uninstall --namespace my-namespace
+  Uninstalls the Appsody Operator in the 'my-namespace- namespace from your Kubernetes cluster.`,
```
kylegc: Maybe double quote "my-namespace" to be consistent with the install help text.

cmd/operator_uninstall.go
```diff
@@ -38,7 +38,7 @@ func newOperatorUninstallCmd(operatorConfig *operatorCommandConfig) *cobra.Comma
 		Short: "Uninstall the Appsody Operator.",
 		Long:  `Uninstall the Appsody Operator from your configured Kubernetes cluster.`,
 		Example: `  appsody operator uninstall --namespace my-namespace
-  Uninstalls the Appsody Operator in the 'my-namespace- namespace from your Kubernetes cluster.`,
+  Uninstalls the Appsody Operator in the "my-namespace-"namespace from your Kubernetes cluster.`,
```
helenmasters: Why is this written as:
... "my-namespace-"namespace...
and not
... "my-namespace" namespace...
?

cmd/operator_install.go
```diff
@@ -33,8 +33,12 @@ func newOperatorInstallCmd(operatorConfig *operatorCommandConfig) *cobra.Command
 	// installCmd represents the "appsody deploy install" command
 	var installCmd = &cobra.Command{
 		Use:   "install",
-		Short: "Install the Appsody Operator into the configured Kubernetes cluster",
-		Long:  ``,
+		Short: "Install the Appsody Operator.",
+		Long: `Install the Appsody Operator into your configured Kubernetes cluster.
+		
+The Appsody Operator listens for incoming AppsodyApplication resources on your cluster. For more information, see https://operatorhub.io/operator/appsody-operator. By default, the operator only watches one namespace, unless the '--watch-all' flag is used. Additional operators must be installed to watch additional namespaces.`,
+		Example: `  appsody operator install --namespace my-namespace --watchspace my-watchspace
```
helenmasters: "By default, the operator only watches one namespace, unless the '--watch-all' flag is used. Additional operators must be installed to watch additional namespaces."

I found these sentences a bit puzzling.

Can we separate these out into 1. default behaviour, 2. what happens with the --watch-all flag 3. how/when you get into a situation where you need to specify another operator to watch another namespace?

"By default, the operator only watches one namespace. If the '--watch-all' flag is specified, the operator watches all namespaces."

If I'd specified the `--watch-all` flag surely I wouldn't need to add another namespace? So is 3. only required if you're watching one namespace and want to watch another namespace (but didn't specify --watch-all)?

cmd/operator_uninstall.go
skoh7645: I have fixed this in the most recent commit

cmd/operator_install.go
skoh7645: Case 3 is if you want to watch more than one namespace, but not all namespaces. I have made the changes to reflect this

cmd/operator_install.go
```diff
@@ -36,7 +36,9 @@ func newOperatorInstallCmd(operatorConfig *operatorCommandConfig) *cobra.Command
 		Short: "Install the Appsody Operator.",
 		Long: `Install the Appsody Operator into your configured Kubernetes cluster.
 		
-The Appsody Operator listens for incoming AppsodyApplication resources on your cluster. For more information, see https://operatorhub.io/operator/appsody-operator. By default, the operator only watches one namespace, unless the '--watch-all' flag is used. Additional operators must be installed to watch additional namespaces.`,
+The Appsody Operator listens for incoming AppsodyApplication resources on your cluster. For more information, see https://operatorhub.io/operator/appsody-operator. 
+
+By default, the operator only watches one namespace. You can specify the '--watch-all' flag is used to tell the Appsody Operator to watch all namspaces. If you want to match multiple namespaces, but not all namespaces, additional operators must be installed to watch additional namespaces.`,
```
helenmasters: "You can specify the '--watch-all' flag is used to tell the Appsody Operator to watch all namspaces." 
doesn't make sense and namspaces is misspelt.

How about:
 If the '--watch-all' flag is specified, the operator watches all namespaces in the cluster.
or 
You can specify the '--watch-all' flag to tell the operator to watch all namespaces in the cluster.
?

helenmasters: The third sentence isn't quite clear either and I think it says match where it means watch...
"If you want to match multiple namespaces, but not all namespaces, additional operators must be installed to watch additional namespaces."?

Can we change the whole piece to read as follows...?

By default, the operator watches a single namespace. You can specify the ‘--watch-all’ flag to tell the operator to watch all namespaces in the cluster. If you want to watch multiple, but not all, namespaces within your cluster, install an additional operator to watch each additional namespace."

skoh7645: I like "You can specify the '--watch-all' flag to tell the operator to watch all namespaces in the cluster."

Issue #588, opened 2019-11-05
Standardize 'appsody version' help text format


skoh7645:
Apply standard created for CLI commands' help text under issue #463, to the `appsody version` command

PR #646, opened 2019-11-18
version: refactor help text

skoh7645:
Refactor help text for `appsody version`

Fixes: #588


cmd/version.go
```diff
@@ -22,8 +22,8 @@ func newVersionCmd(rootCmd *cobra.Command) *cobra.Command {
 	// versionCmd represents the version command
 	var versionCmd = &cobra.Command{
 		Use:   "version",
-		Short: "Show Appsody CLI version",
-		Long:  ``,
+		Short: "Show Appsody CLI version.",
```
Kamran64: nit: "Output" instead of "Show"?

skoh7645: Thanks, sounds better to me

cmd/version.go
```diff
@@ -22,8 +22,8 @@ func newVersionCmd(rootCmd *cobra.Command) *cobra.Command {
 	// versionCmd represents the version command
 	var versionCmd = &cobra.Command{
 		Use:   "version",
-		Short: "Show Appsody CLI version",
-		Long:  ``,
+		Short: "Output Appsody CLI version.",
```
neeraj-laad: Displays the version of Appsody CLI?

cmd/version.go
```diff
@@ -22,8 +22,8 @@ func newVersionCmd(rootCmd *cobra.Command) *cobra.Command {
...
+		Long:  `Output the Appsody CLI version that is currently installed on your local system.`,
```
neeraj-laad: Displays the version of Appsody CLI that is in use?

cmd/version.go
skoh7645: I prefer the word "output" over "display" in this case, but I will change the working from "Output Appsody CLI version" --> "Output the version of the Appsody CLI"

doveye: I prefer the original "show" (as in "Show the version of the Appsody CLI"), as "output" can be something generated from something else :)

skoh7645: @doveye okay, we'll go with that :)

PR #650, opened 2019-11-18
Fix issue with opening a repository index file referenced with file:/// 

groeges:
Signed-off-by: Steve Groeger <GROEGES@uk.ibm.com>


groeges:
Also fixes: #658

PR #660, opened 2019-11-19
Only performs actions when minimum requirements have been set.

### LINKED ISSUE #655: Minimum requirements checker should completely skip if no requirements are set
Currently, when no requirements are set for a stack, the user sees the following logs:
```
Checking stack requirements...
Skipping Docker - No requirements set.
Skipping Appsody - No requirements set.
Skipping Buildah - No requirements set.
```
Since no requirements are set, logging all of this is unnecessary (especially if we add more technologies to enforce later down the line)

Implement a check in the `CheckStackRequirements` function in `utils.go` where:
If no requirements have been set for a stack, output: `No requirements set for stack, continuing...` or words to that effect and avoid outputting logs for skipping individual technologies.

upLukeWeston:
Fixes #655

- If _none_ of the technology requirements have been set, `CheckStackRequirements` is no longer called.

- If _any_ requirements have been set, then `CheckStackRequirements` is called, but only technologies with an assigned value are acted upon. 

- No longer logs if there has not been requirements set per unique technology, only if none of the requirements are set - `"No custom stack requirements set. Skipping"`.


cmd/init.go
```diff
@@ -231,10 +231,23 @@ func initAppsody(stack string, template string, config *initCommandConfig) error
...
+
+		// If no requirements have been set, this function doesn't need to be called
+		if !mapEmpty {
+			checkErr := CheckStackRequirements(config.LoggingConfig, reqsMap, config.Buildah)
+			if checkErr != nil {
+				config.Error.log(checkErr)
+				os.Exit(1)
+			}
+		} else {
+			config.Info.log("No custom stack requirements set. Skipping")
```
Kamran64: Since stack maintainers can't currently set their own custom restrictions, maybe: "No stack requirements set. Skipping" would be better here for now.

cmd/utils.go
```diff
@@ -1883,13 +1883,11 @@ func CheckStackRequirements(log *LoggingConfig, requirementArray map[string]stri
 	log.Info.log("Checking stack requirements...")
 
 	for technology, minVersion := range requirementArray {
-		if minVersion == "" {
-			log.Info.log("Skipping ", technology, " - No requirements set.")
-		} else if technology == "Docker" && buildah {
+		if technology == "Docker" && buildah {
 			log.Info.log("Skipping Docker requirement - Buildah is being used.")
```
Kamran64: Could we be using Debug.log instead of Info so that we don't clutter up the logs?

upLukeWeston: Yeah sure - I did think about this; now Info.log is only used to let the user know that the stack requirements are being checked, each technology check is now a Debug.log.

cmd/init.go
upLukeWeston: Yep, will do. 👍Thanks for clarifying that.

PR #665, opened 2019-11-19
Refactor InitConfig and EnsureConfig so they are testable

djones6:
This PR:
1. Exports `InitConfig` and `EnsureConfig` so that they are testable, and refactors the assignment of the default appsody config values into a separate `ConfigDefaults()` function that produces a `ConfigValues` type.  This would allow direct testing of these functions, for example, to create an Appsody directory in a location other than the default.
2. Replaces `cmd.RootCommandConfig` with `viper.Viper` in various config related functions that did not need the full root config, so that `InitConfig` and `EnsureConfig` can be driven without requiring a full root config.

Motivation: while developing #640, I initially used these functions to generate the appsody configuration file. Later I changed to generating a minimal file manually, but these functions could now be unit tested.  (though I didn't actually add any tests here)

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -136,7 +136,7 @@ func build(config *buildCommandConfig) error {
 		return perr
 	}
 
-	extractDir := filepath.Join(getHome(config.RootCommandConfig), "extract", projectName)
+	extractDir := filepath.Join(getHome(config.CliConfig), "extract", projectName)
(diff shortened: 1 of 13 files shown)


kylegc:
I agree that `initconfig` and `ensureconfig` could use some refactoring. This proposal improves in some ways, but also makes the code more complex in other ways. 
Since we don't have any unit tests ready for this, I don't think it is worth merging at this point. We can always reopen or reference this in the future.

Issue #552, opened 2019-10-31
`appsody stack validate` doesn't validate all the templates in the stack


helenmasters:
**Is your feature request related to a problem? Please describe.**
`appsody stack validate` only validates the default template in the stack.

**Describe the solution you'd like**
I'd like it to loop over all the templates in the stack and create an Appsody project for each one (i.e. performing int, run, test and build against each). This is what I expected validate to do, it should also help prevent stack creators from checking in broken templates. 

If we don't do this, how are we suggesting users validate their other templates, when they have more than one in a stack?

**Describe alternatives you've considered**
N/A

**Additional context**

Issue #636, opened 2019-11-14
stack validate does not return 1 if any of its tests fail


tnixa:
**Describe the bug**
I'm running a bats validation tests that calls `appsody stack validate` and it shows this passes even when there are failures. For example the java-spring-boot2 stack...
```
@@@@@@@@@ Validate Summary Start @@@@@@@@@@
PASSED: Lint for stack: java-spring-boot2
FAILED: Package for stack: java-spring-boot2
PASSED: Init for stack: java-spring-boot2
PASSED: Run for stack: java-spring-boot2
FAILED: Test for stack: java-spring-boot2
FAILED: Build for stack: java-spring-boot2
Total PASSED: 3
Total FAILED: 3
@@@@@@@@@ Validate Summary End @@@@@@@@@@
supra:20191113_162147 tnixa$ echo $?
0
```

**To Reproduce**
1. run `appsody stack validate` on java-spring-boot2

**Expected behavior**
Expecting the command to have a `1` return code if any of the tests fail

**Actual behaviour**
The command has a `0` return code regardless of if the tests pass or fail

**Environment Details (please complete the following information):**
 - OS: [e.g. iOS] mac

**If applicable please specify:**
 - CLI version: 0.4.10
 - Stack you are using: java-spring-boot2

PR #666, opened 2019-11-19
validate all stack templates

tnixa:


tnixa:
fixes issues #552 and #636

PR #668, opened 2019-11-19
Zsh completion support

### LINKED ISSUE #645: appsody completion could support zsh

kubectl, helm, and minikube completion commands all support zsh, appsody does not.

To support `appsody completion zsh`

Not using completion?

**Additional context**

I ran into this on Linux, where I chose zsh, but this will become a problem for mac users very soon, if not already: https://scriptingosx.com/2019/06/moving-to-zsh/

```
source <(helm completion zsh)
source <(kubectl completion zsh)
source <(minikube completion zsh)
#source <(appsody completion zsh) # :-(
```

It would also be useful if `appsody completion` errored when given invalid arguments, rather than ignoring them.

kewegner:
Fixes #645

Initial fullfillment of appsody completion zsh support

1. Will not support --flags until CLI moves to cobra 0.0.5


kylegc:
We will not be upgrading to cobra 0.0.5, it is too risky. I'll merge this in so that we at least have zsh completion for the appsody commands (but not the flags)

PR #686, opened 2019-11-21
Ensure consistent labels in image and app-deploy

EnriqueL8:
Currently the `created` label from the image build and the `app-deploy.yaml` are not matching because we compute the labels twice. In this PR, I cache the labels inside the `RootCommandConfig`. I thought of packaging up the labels into a struct, but it could be the case that a function will be used independently - would like to hear thoughts on this!

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -487,7 +487,6 @@ func generateDeploymentConfig(config *buildCommandConfig, imageName string, labe
 	return nil
 }
 
-
 func updateDeploymentConfig(config *buildCommandConfig, imageName string, labels map[string]string) error {
 	configFile := config.appDeployFile
 
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -202,7 +202,7 @@ func build(config *buildCommandConfig) error {
 	}
 
 	// Generate app-deploy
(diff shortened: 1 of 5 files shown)


kylegc:
First, good catch on this! The labels should definitely be reused and consistent between the build image and the app-deploy.yaml.

For the implementation, caching then in the `RootCommandConfig` works, but I think I prefer a simpler solution. Since the "cache" only needs to happen in the build command, we can simply call `getLabels()` once and then pass the `labels` map as a parameter to `updateDeploymentConfig()`. This we we don't impact the root config for everyone.
What do you think?

EnriqueL8:
okay - sounds good I can modify that. I was just following that approach because we already cache the stack labels.

kylegc:
Do you remember why we added the `cachedStackLabels`? Looks like it is only being called once so the cache isn't used.

kylegc:
@EnriqueL8 I think this looks good, but the build is failing. Please rebase this on top of PR #711 so you can take advantage of the new test sandboxing.

EnriqueL8:
I have rebased this change on top of the test sandboxing PR

PR #688, opened 2019-11-21
Add no-build flag to deploy and convert generate-only to just build

EnriqueL8:
`config.nobuild` or `config.build`?

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -26,7 +26,7 @@ import (
 type deployCommandConfig struct {
 	*RootCommandConfig
 	appDeployFile, namespace, tag, pushURL, pullURL string
-	knative, generate, force, push, nobuild         bool
+	knative, generate, force, push, build           bool
 	dockerBuildOptions                              string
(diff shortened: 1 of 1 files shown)


cmd/deploy.go
```diff
@@ -137,6 +139,7 @@ generates a deployment manifest (yaml) file if one is not present, and uses it t
 	}
 
 	deployCmd.PersistentFlags().BoolVar(&config.generate, "generate-only", false, "DEPRECATED - Only generate the deployment configuration file. Do not deploy the project.")
+	deployCmd.PersistentFlags().BoolVar(&config.build, "no-build", true, "Deploys the application without building a new image or modifying the deployment configuration file.")
```
kylegc: Did you test this? I have doubts that it will work. I think flag defaults to `true`, and if it exists, it is set to `true`. So the way I see it can never be `false`.
Probably better to have the variable be `config.nobuild` and the flag default to `false`

EnriqueL8: Revert these changes, good catch!

PR #689, opened 2019-11-21
Fix build test temp dir

kylegc:


functest/build_test.go
```diff
@@ -134,10 +134,9 @@ func TestBuildLabels(t *testing.T) {
 	}
 
 	// create a temporary dir to create the project and run the test
-	projectDir, err := ioutil.TempDir("", "appsody-build-labels-test")
-	if err != nil {
-		t.Fatal(err)
-	}
+	projectDir := cmdtest.GetTempProjectDir(t)
+	defer os.RemoveAll(projectDir)
+	t.Log("Created project dir: " + projectDir)
```
EnriqueL8: We are doing these operations just below.

EnriqueL8:
Also there are 6 commits showing up, might have been that the branch wasn't originaly rebased on master.

functest/build_test.go
```diff
@@ -134,10 +134,9 @@ func TestBuildLabels(t *testing.T) {
 	}
 
 	// create a temporary dir to create the project and run the test
-	projectDir, err := ioutil.TempDir("", "appsody-build-labels-test")
-	if err != nil {
-		t.Fatal(err)
-	}
+	projectDir := cmdtest.GetTempProjectDir(t)
```
EnriqueL8: I was also thinking if you would want to name the TempDir for debugging purposes?

EnriqueL8: Might also be nicer to return a `cleanup` function which would be equivalent to the `os.RemoveAll` passed to `defer`

functest/build_test.go
kylegc: removed the duplicates

functest/build_test.go
kylegc: I agree with these comments, however `GetTempProjectDir()` has been used all over in the tests for quite some time now, so I'd like to get this fix in and address these comments in a separate PR.

EnriqueL8: Sounds good!

PR #695, opened 2019-11-23
src: change init and extract steps to use sh instead of bash

lance:
Some stack images may be based on intentionally constrained systems
such as alpine. The init step for stacks based on alpine, for example,
fail with a message similar to:

```console
Running command: docker run --rm --entrypoint /bin/bash dev.local/ce-functions:0.0.1 -c find /project -type f -name .appsody-init.sh
[Warning] The stack init script failed: Failed to run the docker find command: exit status 127: docker: Error response from daemon: OCI runtime create failed: container_linux.go:346: starting container process caused "exec: \"/bin/bash\": stat /bin/bash: no such file or directory": unknown.
[Warning] Your local IDE may not build properly, but the Appsody container should still work.
```

This commit modifies those stack processes that depend on bash to instead use `/bin/sh`. And apparently my IDE cleaned up a few trailing whitespaces.


neeraj-laad:
This will impact init and extract.

Will need proper testing, especially the extract side of things.

PR #697, opened 2019-11-25
Templating fix to skip binaries

CameronMcWilliam:
Previous fix skipped JSON, python and javascript files.  We will allow these through while in the future we will create a .templatingIgnore file to allow for future stack creators to add in their own blacklist.


cmd/stack_package.go
```diff
@@ -611,3 +618,13 @@ func ApplyTemplating(stackPath string, templateMetadata interface{}) error {
 	return nil
 
 }
+
+// stringInSlice is used to check for allowed application files
+func stringInSlice(a string, list []string) bool {
```
EnriqueL8: I think we could rename this function to `contains`, because in the end we are just checking that a string is in an array of strings. A def example would be `contains(array, value)`.

cmd/stack_package.go
```diff
@@ -548,6 +548,13 @@ func CreateTemplateMap(labels map[string]string, stackYaml StackYaml, imageNames
 // previously created templateMetada to all files in the target directory
 func ApplyTemplating(stackPath string, templateMetadata interface{}) error {
 
+	// current method means we allow certain application file types through
+	allowedApplicationFiles := []string{
+		"application/json",
+		"application/javascript",
+		"application/x-python",
```
EnriqueL8: FYI: https://stackoverflow.com/questions/45614485/what-is-the-mime-type-for-py-files

CameronMcWilliam: mime type on this package (the one we are using) is application/x-python

https://github.com/gabriel-vasile/mimetype/blob/master/supported_mimes.md

CameronMcWilliam:
Now using a new package that tests if the file is a binary.  Much cleaner and future proof.

PR #698, opened 2019-11-25
templating: add built-in template variables for bumped versions

skoh7645:
This PR adds the bumped versions of major/minor/patch versions to the existing built-in template variables. As seen in the PR to use go-templating in the `java-sprint-boot2` stack (https://github.com/appsody/stacks/pull/517), the stack makes use of the bumped minor version. We could include these bumped versions as built-in variables to help stack creators (currently, this would be the java stacks that require a version range).

### DIFF
--- a/cmd/stack_package.go
+++ b/cmd/stack_package.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 	"text/template"
 	"unicode"
+	"strconv"
 
 	"github.com/gabriel-vasile/mimetype"
 	"github.com/pkg/errors"
@@ -221,7 +222,7 @@ The packaging process builds the stack image, generates the "tar.gz" archive fil
 				return errors.Errorf("Error applying templating: %v", err)
 			}
 
-			// tag with the full version then mojorminor, major, and latest
+			// tag with the full version then majorminor, major, and latest
 			cmdArgs := []string{"-t", buildImage}
 			semver := templateMetadata["stack"].(map[string]interface{})["semver"].(map[string]string)
 			cmdArgs = append(cmdArgs, "-t", namespaceAndRepo+":"+semver["majorminor"])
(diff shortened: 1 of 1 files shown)


skoh7645:
@BarDweller @ebullient what do you think of this addition?

ebullient:
I'm not quite sure how this helps. Each language will have a different syntax for version ranges, and to some extent, it should be up to the stack maintainer to decide what that version string is. Ozzy had a draft that specified the version range in the stack.yaml as a custom attribute (where it stays near the stack version at least).  That may be the better option.

The problem with next is that you don't know what that is.. do you? You have a "next full" .. which is really out there. Next major could be major + 1, next minor could me minor + 1, and next patch could be patch + 1, though this is almost never useful in a version range. Putting these next values together makes no sense, as you wouldn't go from 1.2 to 2.3 .. if you see what I mean. Would have to be current major + minor next, or next major with a 0 for minor.

kylegc:
@skoh7645 What is the use case for this? 
It is a creative way to get the next version number in line, but I'm not really sure how it will be used.

skoh7645:
@ebullient @kylegc This PR would create 3 individual templating variables {{.stack.semver.majornext}}, {{.stack.semver.minornext}}, and {{.stack.semver.patchnext}}. The idea was that the stack creator could use these existing variables instead creating custom ones. So in relation to @BarDweller's PR (appsody/stacks#517), instead of having to create a custom variable `parentpomrange: '[0.3, 0.4)'`, and having to update it when bumping eventually bumping the minor versionof the stack, he would be able to use `[{{.stack.semver.major}}.{{.stack.semver.minor}}, {{.stack.semver.major}}.{{.stack.semver.minornext}})` in the template `pom.xml` to have the same effect.

skoh7645:
Closing this in favour of having stack creator specify their own version range as it is harder to read

PR #701, opened 2019-11-25
Fix default image name

EnriqueL8:
An issue was found where the default name for the image build was different from the one found in the `app-deploy.yaml`. Default image name is now always set to `dev.local/<projectname>`.
Build name is now passed to the `generateDeploymentConfig`, and the logic for a custom tag is only done once.


cmd/build.go
```diff
@@ -527,11 +527,6 @@ func updateDeploymentConfig(config *buildCommandConfig) error {
 
 	appsodyApplication.Spec.CreateKnativeService = &config.knative
 
-	imageName := appsodyApplication.Spec.ApplicationImage
```
kylegc: Looks like this is now going to always overwrite the image in app-deploy.yaml? What if the user had manually edited the app-deploy.yaml with their own image? Maybe we don't care and this is fine, just thinking through the changes.

EnriqueL8: I can imagine that they would run `appsody deploy --no-build`. We could read the `app-deploy.yaml` and if they haven't specified a tag check if the applicationImage is the same as the one we are about to build.

kylegc: @chilanti what are your thoughts on this? Was there a scenario where the user would want to manually change the image in `app-deploy.yaml`?

Kamran64:
Looks good, I've tested this and the deployment works for me without having to tag. I'll hold off from approving until Kyle's comments have been addressed.

cmd/build.go
EnriqueL8: FYI, our current docs state the following `If the `--generate only` flag is used in conjunction with the `--tag (-t)` flag, the value of the tag is preserved in the app-deploy.yaml file and used during subsequent `appsody deploy` invocations.
`. With this change, the user will need to pass the tag flag in every invocation of `appsody build` and `deploy`.

EnriqueL8:
Sound good - I have already update the docs, and the pr is ready here: https://github.com/appsody/website/pull/433

PR #723, opened 2019-11-28
Added deploy_delete tests

### LINKED ISSUE #722: Code coverage analysis for deploy_delete.go
- [x] `if !config.Dryrun && !exists {`
  Ensure that deploy_delete doesn't continue if the `deployConifgFile` cannot be found.

- [x] `if err != nil {`
   Ensure that a valid (the expected) error is returned and propagated from `KubeDelete` if it fails to delete the `config` file

upLukeWeston:
Resolves #722 

Code coverage for `deploy_delete.go` increased from 26.7% to 80.0%


functest/deploy_test.go
```diff
@@ -157,3 +157,67 @@ func TestGenerationDeploymentConfig(t *testing.T) {
 		cleanup()
 	}
 }
+
+// Testing deploy delete when the required config file cannot be found
+func TestDeployDeleteNotFound(t *testing.T) {
+
+	// Not passing a config file so it will use the default, which shouldn't exist
+	_, err := cmdtest.RunAppsodyCmd([]string{"deploy", "delete"}, ".", t)
```
EnriqueL8: This needs to change after the sandboxing PR get's merged

Issue #719, opened 2019-11-27
stack validate nodejs fails, "[Error] Could not find the image either in docker hub or locally: docker.io/dev.local/nodejs:0.2.5"


tnixa:
**Describe the bug**
I am running `stack validate nodejs` via bats test framework which clones the master branch of both the `stacks` and `appsody` repo. The `run` `test` and `build` part of `validate` are failing with what looks like a similar error:
```
[Info] Running command: docker pull docker.io/dev.local/nodejs:0.2.5
[Info] Error response from daemon: Get https://dev.local/v2/: Service Unavailable
[Warning] Docker image pull failed: exit status 1
[Error] Could not find the image either in docker hub or locally: docker.io/dev.local/nodejs:0.2.5
```
Full log is here:
[stack_validate.log](https://github.com/appsody/appsody/files/3899049/stack_validate.log)

**To Reproduce**
1. clone the `appsody` and `stacks` branch
2. from `appsody` branch run `go build` to build latest appsody binary
3. from stacks` branch in `incubator/nodejs` run `appsody stack validate`

**Environment Details (please complete the following information):**
 - OS: [e.g. iOS] macOS, RHEL 7.7

**If applicable please specify:**
 - CLI version: latest
 - Stack you are using: nodejs

kylegc:
Duplicate of #703

PR #702, opened 2019-11-25
Fix local pull by removing docker.io and index.docker.io locally like docker does

kylegc:
Fixes #703

PR #725, opened 2019-11-28
utils: don't use registry for dev.local

skoh7645:
Only use image registry when pulling docker image if the stack isn't `dev.local`

Fixes: #719


tnixa:
@skoh7645  thanks for fixing this! I tested this fork/branch and it does fix #719

kylegc:
I believe this is a duplicate of #702, but I'll take a closer look at the two approaches and see which parts make sense to keep.

kylegc:
Closing, duplicate of #749

PR #730, opened 2019-11-29
Templating uses isBinary package

CameronMcWilliam:
Much cleaner approach using: https://github.com/andrew-d/isbinary

Allows more future proofing, as no need to hard code application files.


Kamran64:
Looks good - just need to fix the Linter errors. I'll allow someone else to take a look/approve. I've added this to Milestone 11 as you said this was a cleaner approach to the previous solution.

Issue #669, opened 2019-11-19
Stack package and validate should fully qualify the image registry


kylegc:
**Is your feature request related to a problem? Please describe.**
Yes, the current design defaults the image registry for stacks to `index.docker.io`. This can be overwritten by an appsody user in their CLI `~/.appsody/.appsody.yaml` config. However, this applies to all appsody projects and is not granular enough. Each project should be fully qualified to the registry.

**Describe the solution you'd like**
The stack package command should use `docker.io/appsody/stack-name:0.0.0` format to fully qualify the stack location. This should get embedded into the template `.appsody-config.yaml` files.

We should also provide a flag to change the default registry from `docker.io` to something else.

For backward compatibility, if a registry doesn't exist, we will still fall back on `~/.appsody/.appsody.yaml`'s `images` field.

**Describe alternatives you've considered**
Using the appsody repo index.yaml files. However these currently don't point to stack images, only the template tars. And the index.yaml is only read on `appsody init` so it would have to embed the registry into the `.appsody-config.yaml` anyway.

PR #749, opened 2019-12-02
utils: Dont use docker.io when checking if local image exists

groeges:
Fixes #703 

Signed-off-by: Steve Groeger <GROEGES@uk.ibm.com>

PR #750, opened 2019-12-02
package/validate: image registry flag

skoh7645:
The `stack package` and `validate` command now uses `registry/namespace/stack:0.0.0` format to fully qualify the stack location. This is embedded into the template `.appsody-config.yaml` files.

Added a flag so user can change the default registry.

Fixes: #669 
This PR should wait for https://github.com/appsody/appsody/pull/725 and https://github.com/appsody/appsody/pull/749

### DIFF
--- a/cmd/stack_package.go
+++ b/cmd/stack_package.go
@@ -79,7 +79,7 @@ func newStackPackageCmd(rootConfig *RootCommandConfig) *cobra.Command {
 
 	var imageNamespace string
 	var imageRegistry string
-	var namespaceAndRepo string 
+	var namespaceAndRepo string
 
 	log := rootConfig.LoggingConfig
 
@@ -198,8 +198,8 @@ The packaging process builds the stack image, generates the "tar.gz" archive fil
 			// docker build
 			// create the image name to be used for the docker image
 			if imageNamespace == "dev.local" && imageRegistry != "docker.io" {
-				return errors.Errorf("Error creating the image name. When specifying the image registry: %v, you must also specify the image namespace.", imageRegistry)
-			} else if imageNamespace == "dev.local"{
+				return errors.Errorf("Error creating the image name. When specifying the image registry: %v: you must also specify the image namespace.", imageRegistry)
+			} else if imageNamespace == "dev.local" {
 				namespaceAndRepo = imageNamespace + "/" + stackID
 			} else {
 				namespaceAndRepo = imageRegistry + "/" + imageNamespace + "/" + stackID
--- a/cmd/stack_package.go
+++ b/cmd/stack_package.go
@@ -197,12 +197,16 @@ The packaging process builds the stack image, generates the "tar.gz" archive fil
 
 			// docker build
 			// create the image name to be used for the docker image
+			imageRepo := rootConfig.CliConfig.GetString("images")
+
 			if imageNamespace == "dev.local" && imageRegistry != "docker.io" {
 				return errors.Errorf("Error creating the image name. When specifying the image registry: %v: you must also specify the image namespace.", imageRegistry)
 			} else if imageNamespace == "dev.local" {
 				namespaceAndRepo = imageNamespace + "/" + stackID
-			} else {
+			} else if imageRegistry != "docker.io" {
 				namespaceAndRepo = imageRegistry + "/" + imageNamespace + "/" + stackID
+			} else {
+				namespaceAndRepo = imageRepo + "/" + imageNamespace + "/" + stackID
 			}
 
 			buildImage := namespaceAndRepo + ":" + stackYaml.Version
--- a/cmd/stack_package.go
+++ b/cmd/stack_package.go
@@ -219,7 +219,7 @@ The packaging process builds the stack image, generates the "tar.gz" archive fil
 			}
 
 			// create the template metadata
-			templateMetadata, err := CreateTemplateMap(labels, stackYaml, imageNamespace)
+			templateMetadata, err := CreateTemplateMap(labels, stackYaml, imageNamespace, imageRegistry)
 			if err != nil {
 				return errors.Errorf("Error creating templating mal: %v", err)
 			}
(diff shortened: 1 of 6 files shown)


neeraj-laad:
@skoh7645 Please can you resolve the conflicts on this PR?

cmd/stack_package.go
```diff
@@ -195,7 +197,18 @@ The packaging process builds the stack image, generates the "tar.gz" archive fil
 
 			// docker build
 			// create the image name to be used for the docker image
-			namespaceAndRepo := imageNamespace + "/" + stackID
+			imageRepo := rootConfig.CliConfig.GetString("images")
+
+			if imageNamespace == "dev.local" && imageRegistry != "docker.io" {
```
kylegc: Hi Sandy. I've been thinking about this more and talking to Michele. We've really been using `dev.local` wrong (not your fault). 
Keeping in mind that the image format is REGISTRY/NAMESPACE/REPOSITORY:TAG
The `dev.local` part should really be the registry, not the namespace. It is implying that this is a non-existing, local only registry. 

Please investigate making the `--image-registry` default to `dev.local` and the `--image-namespace` default to something else, like `stacks`. I haven't fully thought through all the coding challenges of this, so please try it out and let us know what this complicates and/or simplifies.

skoh7645: I spoke with @neeraj-laad about this and it makes sense to have `dev.local` as the registry. `appsody` might be a better default option for the namespace over `stacks`

cmd/stack_package.go
```diff
@@ -195,7 +197,16 @@ The packaging process builds the stack image, generates the "tar.gz" archive fil
 
 			// docker build
 			// create the image name to be used for the docker image
-			namespaceAndRepo := imageNamespace + "/" + stackID
+			imageRepo := rootConfig.CliConfig.GetString("images")
+
+			if imageRegistry != "dev.local" && imageNamespace == "appsody" {
```
kylegc: I don't think we need any of this logic anymore. Just make 
`namespaceAndRepo = imageRegistry + "/" + imageNamespace + "/" + stackID`

Now that we've moved `dev.local` to the registry, we don't need to read "images" from the CliConfig. Just always use the value in the flag.

cmd/stack_validate.go
```diff
@@ -78,6 +79,10 @@ Runs the following validation tests against the stack and its templates:
 			stackName := filepath.Base(stackPath)
 			rootConfig.Info.Log("stackName is: ", stackName)
 
+			if imageRegistry != "dev.local" && imageNamespace == "appsody" {
```
kylegc: Is this necessary?

cmd/stack_validate.go
```diff
@@ -148,7 +153,7 @@ Runs the following validation tests against the stack and its templates:
 				rootConfig.Info.Log("Created project dir: " + projectDir)
 
 				// init
-				err = TestInit(rootConfig.LoggingConfig, "dev.local/"+stackName, templates[i], projectDir)
+				err = TestInit(rootConfig.LoggingConfig, imageRegistry+imageNamespace+stackName, templates[i], projectDir)
```
kylegc: Need a "/"s between the parts. Since this is used multiple times, create a variable with the fully qualified image once to use it in all the commands.

cmd/utils.go
```diff
@@ -482,19 +482,16 @@ func getProjectConfig(config *RootCommandConfig) (*ProjectConfig, error) {
 		stack := v.GetString("stack")
 		config.Debug.log("Project stack from config file: ", projectConfig.Stack)
 		imageRepo := config.CliConfig.GetString("images")
-		config.Debug.log("Image repository set to: ", imageRepo)
```
kylegc: Can you justify all of these changes to `utils.go`? I don't see why they are necessary.

cmd/stack_package.go
skoh7645: My intention was to use "images" if they set a namespace but not registry, and only use `dev.local` if neither the namespace nor registry was set. If I don't use "images" here, where else is it being used/why do we need it?

cmd/stack_validate.go
skoh7645: I thought it would make more sense to say "if you specify a registry, you must specify the namespace". I will remove this

cmd/utils.go
skoh7645: The `if len(imageComponents) < 3` no longer makes sense as this would never be the case anymore. This part overrides the project stack registry with "images" from CliConfig. I'm not sure why we would want this to be the case and I think we can remove `OverrideStackRegistry` altogether?

cmd/stack_validate.go
skoh7645: Just realised it doesn't make sense to have registry and namespace here...I will remove it altogether

cmd/utils.go
skoh7645: I have tested it without `OverrideStackRegistry` and `init` without overriding the project stack registry with "images" works now. I will push the changes. If you test it and it works fine for you I can remove the `OverrideStackRegistry` function altogether

cmd/stack_package.go
kylegc: Thanks for removing this. 
"images" from `~/.appsody/.appsody.yaml` is being used elsewhere. Specifically when pulling a stack that was not fully qualified during `appsody run`

cmd/utils.go
kylegc: TLDR: Please undo all changes to utils.go from this PR.

Hi Sandy, I can see why you don't think this is necessary, but first you have to remember backward compatibility. There will be existing template tars out in the field that don't yet fully qualify their stack registry. In those cases, we need this code to add the "images" registry. 
And second, `OverrideStackRegistry` is needed because of the new `--stack-registry` flag on `appsody run`. It will win over all other registries.

skoh7645: Okay I will undo these changes and push them (as this might be from my lack of understanding?). But without these changes, if you do an `appsody stack package` and then an `appsody init`, the `--image-registry` value you set when doing the `appsody stack package` would just get replaced with `docker.io`...

kylegc: Okay, this was the justification I originally asked for. I agree this is bad behavior. So we have to figure out a way to preserve the OverrideStackRegistry flag behavior, but not default to docker.io. Let me take a closer look and get back to you.

kylegc: Alright. I reproduced the behavior and I think the proper fix is to reset `config.StackRegistry` during `init`. 

I hope you don't mind, I've pushed a commit to your branch (this PR) to fix it.

Please pull and test this now. I'm going to approve it, so if your testing passes, feel free to merge.

skoh7645: I don't mind at all! Thank you!

skoh7645:
I tested the whole Appsody flow `package|validate|init|run|test|debug|deploy` with the `nodejs-express` stack and it works well. Merging

Issue #710, opened 2019-11-26
Code coverage analysis for stack_yaml_lint.go


Kamran64:
- [ ] `err = yaml.Unmarshal([]byte(stackyaml), s)`
  Run linter against a stack.yaml with an invalid structure and check for error

- [ ] `if !defaultTemplateFound`
  Run linter against stack with a missing value for `default-template` in the stack.yaml and check for error

- [ ] `if yamlValues[i] == ""`
  Run linter against stack.yaml with missing fields e.g. name, version, license etc. and check for error.

- [ ] `if len(maintainerEmails) == 0`
  Run linter against stack.yaml with no maintainers specified and check for error.

- [ ] `if len(versionNo) < 3`
  Run linter against stack.yaml with invalid version and check for error. In this case, a version will less than 3 elements or more than 4.

- [ ] `if len(s.Description) > 70 `
  Run linter against stack.yaml with description field with greater than 70 characters and check for error.

- [ ] `if len(s.Name) > 30`
  Run linter against stack.yaml with name field with greater than 30 characters and check for error.

**Note:** This list is subject to change after https://github.com/appsody/appsody/pull/627 is merged.

Issue #708, opened 2019-11-26
Code coverage analysis for stack_lint.go


Kamran64:
- [x] `validStackID, err := IsValidProjectName(stackID)`
    Run linter with invalid stack name and test that it fails.

- [x] `fileCheck, err = Exists(filepath.Join(stackPath, "/stack.yaml"))`
    Run linter with missing `stack.yaml` and validate error message

- [x] `fileCheck, err = Exists(imagePath)`
    Run linter when stack is missing an image directory and validate the error message.

- [x] `fileCheck, err = Exists(filepath.Join(imagePath, "/Dockerfile-stack"))`
    Run linter when Dockerfile-stack is missing and validate the error message.

- [x] `fileCheck, err = Exists(filepath.Join(imagePath, "/LICENSE"))`
    Run linter when LICENSE file is missing and validate the error message.

- [x] `fileCheck, err = Exists(configPath)`
    Run linter when config folder is missing and validate warning message. The linter should not fail here.

- [x] `fileCheck, err = Exists(templatePath)`
    Run linter when templates folder is missing and validate error message. There should be an error complaining about the missing folder and no templates within the template dir.

- [x] `IsEmptyDir(templatePath)`
    Run linter with templates folder but with no templates within the dir.

- [x] `fileCheck, err = Exists(filepath.Join(templatePath, f.Name(), ".appsody-config.yaml"))`
    Run linter with an `appsody-config.yaml` inside a template.

**Note**: Some of these tests cases could be combined into one test.

PR #751, opened 2019-12-02
Improve code coverage for linter files

Kamran64:
- Improves code coverage from ~64% to ~80% for `stack_lint.go`
- Improves code coverage from 98% to 100% for `stack_lint_dockerfile_stack.go`
- Improves code coverage from 68.2% to 95.5% for `stack_yaml_lint.go`

- Adds check for project directory in Linter

Fixes: https://github.com/appsody/appsody/issues/708 Fixes: https://github.com/appsody/appsody/issues/710


cmd/stack_lint_test.go
```diff
@@ -286,44 +287,327 @@ func TestLintWithValidStack(t *testing.T) {
...
+
+	if appsodyErr == nil { //Lint check should fail, if not fail the test
+		t.Fatal("Expected failure - Missing stack.yaml")
+	}
+
+	RestoreSampleStack(removeArray, testStackPath)
+	writeErr := ioutil.WriteFile(removeYaml, []byte(file), 0644)
+	if writeErr != nil {
+		t.Fatal(writeErr)
+	}
```
upLukeWeston: Could you add the `RestoreSampleStack` into a defer (or make the current defer into it a defer function which calls it alongside `cleanup`) so if the test is quit during the run, the files are still restored?

It should probably be done for all other instances too.

upLukeWeston: Apart from this, the rest looks good to me!

Kamran64: Good idea - Will add this in the next commit.

Issue #743, opened 2019-11-29
Code coverage analysis for repo_remove.go


skoh7645:
- [ ] func newRepoRemoveCmd & if repoErr != nil  
test repo remove with an incorrectly formatted `repository.yaml` and verify the error

- [ ] func newRepoRemoveCmd & if repoName != defaultRepoName
test repo remove with the default repository and verify the error

PR #760, opened 2019-12-04
repo_remove.go: improve code coverage

skoh7645:
Write tests to bump code coverage from 78.3% to 91.3%

Fixes: #743


cmd/stack_package.go
```diff
@@ -367,7 +367,7 @@ The packaging process builds the stack image, generates the "tar.gz" archive fil
 				}
 				if repoNameToDelete != "" {
 					log.Info.logf("Appsody repo %s is configured with %s's URL. Deleting it to setup %s.", repoNameToDelete, repoName, repoName)
-					repos.Remove(repoNameToDelete)
+					repos.Remove(repoNameToDelete, rootConfig)
 				}
```
upLukeWeston: And same for these two

cmd/repo.go
```diff
@@ -415,11 +415,12 @@ func (r *RepositoryFile) GetDefaultRepoName(rootConfig *RootCommandConfig) (stri
 	return repoName, nil
 }
 
-func (r *RepositoryFile) Remove(name string) {
+func (r *RepositoryFile) Remove(name string, rootConfig *RootCommandConfig) {
```
upLukeWeston: If the only thing `rootConfig` is doing is related to logging, instead of passing `RootCommandConfig`, you should pass a `LoggingConfig` :)

cmd/repo_remove.go
```diff
@@ -52,12 +52,12 @@ You cannot remove the default repository, but you can make a different repositor
 						return err
 					}
 					if repoName != defaultRepoName {
-						repoFile.Remove(repoName)
+						repoFile.Remove(repoName, config)
 					} else {
```
upLukeWeston: Then here you would pass `config.LoggingConfig`

cmd/stack_package.go
skoh7645: Makes sense! thanks :)

PR #767, opened 2019-12-06
Run git commands in project dir rather than the current working directory

kylegc:
Although not yet supported by the command line options, all of our code runs relative to an absolute path ProjectDir instead of the current working directory - except for the git commands. This should be fixed for tests to run properly and future proof the code for command line support.

### DIFF
--- a/functest/build_test.go
+++ b/functest/build_test.go
@@ -95,6 +95,14 @@ var appsodyStackLabels = []string{
 	"configured",
 }
 
+var appsodyCommitKey = "dev.appsody.image.commit."
+var appsodyCommitLabels = []string{
+	"message",
+	"date",
+	"committer",
+	"author",
+}
+
 func TestBuildLabels(t *testing.T) {
 	stacksList = "incubator/nodejs"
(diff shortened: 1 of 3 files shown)


functest/build_test.go
```diff
@@ -158,6 +175,18 @@ func TestBuildLabels(t *testing.T) {
...
+	for _, label := range appsodyCommitLabels {
+		if labelsMap[appsodyCommitKey+label] == nil {
+			t.Errorf("Could not find %s%s label in Docker image!", appsodyCommitKey, label)
+		}
+	}
+
+	if labelsMap[appsodyCommitKey+"message"] != commitMessage {
+		t.Errorf("Expected commit message \"%s\" but found \"%s\"", commitMessage, labelsMap[appsodyCommitKey+"message"])
+	}
+
```
EnriqueL8: Just a suggestion: could we also check for the upstream url?

kylegc: Yes, we could check for a lot more labels. However, I didn't want to get carried away in this PR. The commit message was giving me grief so I wanted to make sure that one was working.

PR #769, opened 2019-12-06
Unit tests for stack_create.go

CameronMcWilliam:
Puts coverage up to 84.3%


cmd/stack_create_test.go
```diff
@@ -214,3 +214,36 @@ func TestStackAlreadyExists(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestStackCreateMissingArguments(t *testing.T) {
```
upLukeWeston: I think that maybe it should be made clear that the test is expecting a failure in its title `TestStackCreateFailMissingArguments` or `TestStackCreateMissingArgumentsFail`,  because the `if err == nil` sometimes throws people off when scanning through (if this test were to fail, someone might assume it's because of err is being returned as nil - generally what we'd want after running a command)

cmd/stack_create_test.go
```diff
@@ -214,3 +214,36 @@ func TestStackAlreadyExists(t *testing.T) {
...
+	args := []string{"stack", "create"}
+	_, err := cmdtest.RunAppsodyCmd(args, ".", t)
+
+	if err == nil {
```
upLukeWeston: Rather than blindly testing that an error is being returned, it'd be wiser to confirm that the error being returned is what we are expecting, because it could have returned for a completely unrelated issue and go unnoticed. Instead I would suggest something like:
```
if err != nil {
	if !strings.Contains(err.Error(), "<your expected error here>") {
		t.Errorf("String \"<your expected error here>\" not found in output: '%v'", err.Error())
	}
} else {
	t.Error("Expected an error to be returned from command, but error was nil")
}
```

cmd/stack_create_test.go
```diff
@@ -214,3 +214,36 @@ func TestStackAlreadyExists(t *testing.T) {
...
+		t.Fatal(err)
```
upLukeWeston: Related to the above comment, but there isn't much point in returning the err when it's been confirmed to be `nil`

PR #773, opened 2019-12-06
Lint APPSODY Mounts for single file mounts

### LINKED ISSUE #774: Lint APPSODY_MOUNTS during appsody run and appsody stack lint
Single file Docker mounts will be deprecated in the future we need to lint the mounts to warn users of this and detect any issues.

kewegner:
Do Not Merge.
Fixes #774


cmd/utils.go
```diff
@@ -2113,3 +2114,107 @@ func checkValidLicense(log *LoggingConfig, license string) error {
...
+			log.Warning.logf("Path %s for mount points to a single file.  Single file Docker mount paths cause unexpected behavior and will be deprecated in the future.", path)
+		}
+
+	}
+}
+
+func lintMountVar(mountListSource string, log *LoggingConfig, stackPath string) (int, int) {
+
+	if mountListSource == "" {
+		log.Warning.log("No APPSODY MOUNTS exists, mount paths can not be validated.")
```
kylegc: This should be an Error log

cmd/utils.go
```diff
@@ -2113,3 +2114,107 @@ func checkValidLicense(log *LoggingConfig, license string) error {
...
+	return warningCount, errCount
+}
+func validateMountPath(path string, mount string, log *LoggingConfig) (int, error) {
+	log.Debug.log("Attempting to validate mount path: ", path)
+	var mountError error
+	warningCount := 0
+	file, err := os.Stat(path)
+	if err != nil {
+
+		mountError = errors.Errorf("Could not stat path: %s for mount %s", path, mount)
```
kylegc: This should be printed somewhere. Maybe print here and return a count instead of error

cmd/utils.go
```diff
@@ -2113,3 +2114,107 @@ func checkValidLicense(log *LoggingConfig, license string) error {
...
+
+	// loop through the template directories
+	for _, f := range templates {
+
+		for _, mount := range mountList {
+			traceMount := strings.Trim(mount, "\"")
+			log.Debug.log("mount pair: ", traceMount)
+			localPaths := strings.Split(traceMount, ":")
+			if len(localPaths) != 2 {
+				log.Error.log("Mount is not properly formatted: ", traceMount)
```
kylegc: Clarify that it requires a single colon

cmd/utils.go
```diff
@@ -2113,3 +2114,107 @@ func checkValidLicense(log *LoggingConfig, license string) error {
...
+			log.Debug.logf("Path %s for mount is a directory", path)
+		} else {
+
+			log.Warning.logf("Path %s for mount points to a single file.  Single file Docker mount paths cause unexpected behavior and will be deprecated in the future.", path)
+		}
+
+	}
+}
+
+func lintMountVar(mountListSource string, log *LoggingConfig, stackPath string) (int, int) {
```
kylegc: These functions will only be used in `stack_lint_dockerfile_stack.go` so I think they should be moved there.

cmd/utils.go
kewegner: done

Issue #779, opened 2019-12-08
stack package lint should check the requirement constraints are valid


kylegc:
The stack linter should check each technology `requirements` and use `setConstraint, err := semver.NewConstraint(minVersion)` to ensure they have a proper semver constraint syntax.

PR #780, opened 2019-12-09
Update the linter to check for stack requirements

Kamran64:
- Update the linter to check that stack requirements meet the format of the package we use to set constraints.

Fixes: https://github.com/appsody/appsody/issues/779

### DIFF
--- a/cmd/stack_yaml_lint.go
+++ b/cmd/stack_yaml_lint.go
@@ -21,6 +21,7 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/Masterminds/semver"
 	"gopkg.in/yaml.v2"
 )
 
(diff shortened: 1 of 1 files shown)


cmd/stack_yaml_lint.go
```diff
@@ -139,3 +142,26 @@ func (stackDetails *StackYaml) checkLicense(log *LoggingConfig) int {
 	}
 	return stackLintErrorCount
 }
+
+func (stackDetails *StackYaml) checkRequirements(log *LoggingConfig) int {
+	stackLintErrorCount := 0
+
+	reqsMap := map[string]string{
+		"Docker":  stackDetails.Requirements.Docker,
```
kylegc: What happens when `requirements` is not specified in `stack.yaml`? I believe this like will panic. Please test this and fix accordingly.

Kamran64: If requirements aren't set in the `stack.yaml`, the linter will continue. Just tested this now. I believe the values are set as empty strings if the requirement isn't set hence why I added the `if req == ""; continue` lines.

kylegc: OK, I was worried `stackDetails.Requirements` would be nil, but I suppose since it is part of the StackYaml struct it gets a value. 

Thanks for testing. I'll approve and merge.

Issue #729, opened 2019-11-29
appsody build with buildah push integration failed


brightzheng100:
**Describe the bug**
/kind bug

Using `buildah` directly works fine:

```sh
$ buildah push --tls-verify=false "$imgid" docker://registry2:5000/spring-boot-demo:temp
```

But `appsody build` with `buildah` failed:

```sh
$ appsody -v  build --buildah -t "spring-boot-demo:k8s" --buildah-options "--tls-verify=false" --push --push-url docker://registry2:5000
```

With below errors:

```
...
[Buildah] STEP 15: COPY --from=compile /project/user-app/target/app.jar /app.jar
[Buildah] STEP 16: EXPOSE 8080
[Buildah] STEP 17: EXPOSE 8443
[Buildah] STEP 18: ENTRYPOINT [ "sh", "-c", "exec java $JVM_ARGS -Djava.security.egd=file:/dev/./urandom -jar /app.jar" ]
[Buildah] STEP 19: COMMIT docker://registry2:5000/spring-boot-demo:temp
[Buildah] Getting image source signatures
[Buildah] Copying blob sha256:d9ff1782a1ccf0ed26b6481ed70d1001e5a33a76abdbbe4920614ea5d6e14104
[Buildah] Copying blob sha256:2ad4dcd79a7924d70f2bdddddf0cc363ce2981bc43bda2b88f4a374e2b7dff97
[Buildah] Copying blob sha256:dbab956cded4b1c4ad14dd257d32c0cc5d45040fb9862c101541d3bbd753c9be
[Buildah] Copying blob sha256:813e9a2dd79d946e817b4e6fe678a51a2cd8b59cb2cc7ed62251b139081988fe
[Buildah] Copying blob sha256:be184efd1818c38604382198fd2020769e5e582524af3fdfd6f54e19806975ea
[Buildah] Copying config sha256:36680b886e51a89a3fab9f8efb928fea2d8309febb409cd4ea4946b7b2efabaf
[Buildah] Writing manifest to image destination
[Buildah] Storing signatures
[Buildah]
[Info] Pushing image docker://registry2:5000/spring-boot-demo:temp
[Error] Could not push the image: exit status 1
[Error] Could not push the docker image - exiting. Error: exit status 1

Could not push the docker image - exiting. Error: exit status 1
github.com/appsody/appsody/cmd.build
	/home/travis/gopath/src/github.com/appsody/appsody/cmd/build.go:196
[...]

PR #782, opened 2019-12-09
fix buildah missing options

brightzheng100:
This is a fix for #729.

In `appsody build` command like:

```sh
$ appsody build -v --buildah -t "spring-boot-demo:docker" --buildah-options "--tls-verify=false" --push-url docker://registry:5000
```

The `buildah` options `"--tls-verify=false"` would be missing because of the bug.

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -194,7 +194,7 @@ func build(config *buildCommandConfig) error {
 		return execError
 	}
 	if config.pushURL != "" || config.push {
-		err := ImagePush(config.LoggingConfig, buildImage, config.Buildah, config.Dryrun)
+		err := ImagePush(config.LoggingConfig, buildImage, config.Buildah, config.buildahBuildOptions, config.Dryrun)
 		if err != nil {
 			return errors.Errorf("Could not push the docker image - exiting. Error: %v", err)
 		}
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -1300,14 +1300,15 @@ func DockerTag(log *LoggingConfig, imageToTag string, tag string, dryrun bool) e
 }
 
 //ImagePush pushes a docker image to a docker registry (assumes that the user has done docker login)
-func ImagePush(log *LoggingConfig, imageToPush string, buildah bool, dryrun bool) error {
+func ImagePush(log *LoggingConfig, imageToPush string, buildah bool, pushArgs string, dryrun bool) error {
 	log.Info.log("Pushing image ", imageToPush)
 	cmdName := "docker"
 	if buildah {
 		cmdName = "buildah"
 	}
 
-	cmdArgs := []string{"push", imageToPush}
+	//the order matters especially with buildah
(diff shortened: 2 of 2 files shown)


EnriqueL8:
@brightzheng100 looking at this change, we could have a potential issue with the flags between `buildah push` and `buildah bud`  being different since we currently only allow to pass the flags through the `buildah-options` flag.

brightzheng100:
While looking at the code, it becomes complicated as it mixes some groups of flags:
- The flags for `appsody build` itself;
- The flags for `buildah bud` if we're using `appsody build --buildah`;
- Additional flags for `buildah push`, if `buildah` is enabled and `--push` and/or `----push-url` are enabled

For a longer-term perspective, a code refactoring should be seriously considered but as of now, this fix might still be acceptable.

There are two potential solutions that I can think of:

1. Separate `appsody build` and `appsody push` to simplify the overall command line experience;
2. Make the flags clearer, say `--buildah-build-options`, `--docker-build-options`, `--buildah-push-options`, `--docker-push-options`, etc.

what do you think? @EnriqueL8

EnriqueL8:
I agree, we need to think about the longer-term. We need to be conscious about incrementing the number of flags to one command, and adding a lot of complexity to the code base cause we would need to manage more scenarios. 

From the two options proposed:

1. `appsody push` would be directly `docker push` in the local development case. The only value we would add is `buildah`, and then in that case we need to add options for `buildah` so that would be very similar flags to the current `appsody build` command.

2. I would suggest just a `build-options` and a `push-options` flags. If those flags are for **buildah** then the `--buildah` flag is passe, if not they are **docker** options.

We should discuss this future before attempting to implement a quick fix. 
What are your thoughts @kylegc @chilanti @neeraj-laad ?

kylegc:
What's wrong with doing `appsody build` then `buildah push` separately? 

I'm concerned with supporting and maintaining too many flags/options. My opinion is that we continue to support the `--push` flag on appsody as a convenience to the default docker/buildah push. But if you need to specify custom options on the push, then you should run it as a separate command.

Would like to hear other's opinions as well @chilanti @neeraj-laad

brightzheng100:
The only issue while separating `appsody build` and `appsody push` is that, if you're using `buildah push`, an image ID must be provided (e.g. `buildah push imageID docker://registry.example.com/repository:tag`), which creates some dependency from `appsody push` to `appsody build` -- please consider whether it may break the UX.

As of now, I actually agreed with @EnriqueL8's fine-tuned second approach, with a little bit extra idea:
- to have dedicated `build-options` and `push-options` in `appsody build` command;
- there might be some "drivers" to enable, for now it's `buildah` and these drivers may grow. So it might be a good idea to reserve such a potential to have a dedicated flag for growing options, other than just `docker` and `buildah` -- so, it can be something like `--driver buildah` and it defaults to docker.

I can help to refine this PR for such an approach, if this is agreeable. @EnriqueL8 @kylegc

PR #788, opened 2019-12-10
Updated all current tests to use new sandboxing

### LINKED ISSUE #711: Update all tests to use new sandboxing
Follow up to PR #709 we need to update all the tests to use the new sandboxing methods. 

1. All tests should either call `TestSetup()` or `TestSetupWithSandbox()`
2. All tests should use the new methods from #709

When finished, you should be able to remove the following testutils methods:
- RunAppsodyCmd
- AddLocalFileRepo
- GetTempProjectDir

upLukeWeston:
Resolves #711 

Updated all tests to use new sandboxing functionality, and removed the now unnecessary functions from `testutils.go`


Kamran64:
Changes look good from my perspective. Will allow @skoh7645 and @CameronMcWilliam to take a look also before approving.

cmd/cmdtest/testutils.go
```diff
@@ -31,6 +31,8 @@ import (
 
 const CLEANUP = true
 
+const TestDirPath = "../cmd/testdata"
```
kylegc: Best to use `filepath.Join()` here too. You probably can't use `const` then, but that's okay. Rather have it work on multiple platforms.

kylegc: Also, I think all references to "testdata" below can also use this `TestDirPath`

functest/package_test.go
```diff
@@ -11,23 +11,35 @@
...
 	sandbox, cleanup := cmdtest.TestSetupWithSandbox(t, true)
 	defer cleanup()
 
-	args := []string{"stack", "package", filepath.Join("..", "cmd", "testdata", "starter")}
-	_, err := cmdtest.RunAppsody(sandbox, args...)
+	var outBuffer bytes.Buffer
+	log := &cmd.LoggingConfig{}
+	log.InitLogging(&outBuffer, &outBuffer)
+
+	stackDir := filepath.Join("..", "cmd", "testdata", "starter")
```
kylegc: Use `TestDirPath` here

functest/package_test.go
```diff
@@ -11,23 +11,35 @@
...
+	err := cmd.CopyAllContents(log, stackDir, sandbox.ProjectDir)
+	if err != nil {
+		t.Errorf("Problem copying %s to %s: %v", stackDir, sandbox.ProjectDir, err)
+	} else {
+		t.Logf("Copied %s to %s", stackDir, sandbox.ProjectDir)
+	}
 
+	args := []string{"stack", "package"}
+	_, err = cmdtest.RunAppsody(sandbox, args...)
 	if err != nil {
```
kylegc: There should be more validation that the `stack package` command works here. Making sure there was no error is good, but we should also check for instance that a new image exists with the proper tag. And a proper template tar exists.

This comment should be resolved in a separate PR. The current PR is about updating the sandboxing not improving the test cases, but I wanted to point it out while I was here reviewing.

cmd/utils.go
```diff
@@ -636,6 +636,51 @@ func CopyDir(log *LoggingConfig, fromDir string, toDir string) error {
 	return nil
 }
 
+// CopyAllContents copies the files and folders from one place to another, without the parent folder
+func CopyAllContents(log *LoggingConfig, fromDir, toDir string) error {
+	var files []string
+
+	// Get the filepaths within the given folder
+	err := filepath.Walk(fromDir, func(path string, info os.FileInfo, err error) error {
```
kylegc: The `filepath.Walk` method will traverse the entire directory tree, including subdirectories. I'm not sure you want this, because below you use `CopyDir` which copies all the contents of a directory. You probably only need to list the top level files and dirs from this method. I suggest looking at `ioutil.ReadDir` - this will already give you a FileInfo so you don't need to call `os.Stat` later.

Alternatively, you could avoid writing this `CopyAllContents` function altogether. Use `CopyDir` to copy the `stackDir` into the `sandbox.ProjectDir`, then reset `sandbox.ProjectDir` to be one level lower (aka the new stackDir)

cmd/cmdtest/testutils.go
upLukeWeston: Oops, missed this! Yeah I initially did this for `const`, but of course it makes more sense to make it platform unspecific.

functest/package_test.go
upLukeWeston: Of course, yes! Done, and changed for most places throughout the code!

cmd/utils.go
upLukeWeston: I've chosen to go with an alternative method :) Realised I was being a bit silly and overcomplicated it; the new utils.go function has been removed and instead the `sandbox.ProjectDir` now goes one level deeper into the `starter` folder before running `appsody stack package`

kylegc: Yep I like this much better! Look at it as good coding exercise ;-)
Also great job commenting the code for this exception! Thanks!

cmd/stack_lint_test.go
```diff
@@ -206,16 +221,19 @@ func Test_KILLValue(t *testing.T) {
 	}
 
 	output = strings.Join(lines, "\n")
-	err = ioutil.WriteFile("../cmd/testdata/test-stack/image/Dockerfile-stack", []byte(output), 0644)
+	err = ioutil.WriteFile(filepath.Join(cmdtest.TestDirPath, "test-stack", "image", "Dockerfile-stack"), []byte(output), 0644)
 
 	if err != nil {
 		t.Fatal(err)
```
EnriqueL8: These tests will fail if they are run in parallel, since they all write and read from the same file. I ran `go test` locally and it failed. It doesn't fail in travis since this runs `make test` which passes `--parallel 1`. A way to solve this would be to copy the `testdata` directory in the `setupSandbox` function.

functest/init_test.go
```diff
@@ -221,7 +214,8 @@ func TestWhiteList(t *testing.T) {
 	}
 
 	// appsody init nodejs-express
-	_, _ = cmdtest.RunAppsodyCmd([]string{"init", "nodejs-express"}, projectDir, t)
+	args := []string{"init", "nodejs-express"}
+	_, _ = cmdtest.RunAppsody(sandbox, args...)
```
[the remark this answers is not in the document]
EnriqueL8: I agree

cmd/stack_lint_test.go
EnriqueL8: Yes - we could do this as part of a different PR.

cmd/stack_lint_test.go
```diff
@@ -288,9 +308,12 @@ func TestLintWithMissingConfig(t *testing.T) {
...
 
 func TestLintWithMissingProject(t *testing.T) {
+	sandbox, cleanup := cmdtest.TestSetupWithSandbox(t, true)
+	defer cleanup()
+
 	currentDir, _ := os.Getwd()
 	testStackPath := filepath.Join(currentDir, "testdata", "test-stack")
-	args := []string{"stack", "lint"}
+	args := []string{"stack", "lint", testStackPath}
 	removeProj := filepath.Join(testStackPath, "image", "project")
```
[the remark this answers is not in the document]
upLukeWeston: I guess not! I've removed them and the tests didn't brick themselves after multiple runs - I'll give it a go on Travis

functest/init_test.go
```diff
@@ -159,7 +153,8 @@ func TestNoTemplate(t *testing.T) {
 	shouldExist(appjs, t)
 
 	// appsody init nodejs-express
-	_, _ = cmdtest.RunAppsodyCmd([]string{"init", "nodejs-express", "--no-template"}, projectDir, t)
+	args := []string{"init", "nodejs-express", "--no-template"}
+	_, _ = cmdtest.RunAppsody(sandbox, args...)
```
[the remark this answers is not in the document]
upLukeWeston: I'll leave this for another PR too - I didn't want to change too many moving parts at the same time, there are a few changes to tests I noticed too that I'll probably alter after this goes in!

Issue #560, opened 2019-11-01
Update stack package command to support templating in stacks


neeraj-laad:
This will be follow on work from the exploratory work done under https://github.com/appsody/stacks/issues/430

We now need to:

- [ ] Agree on what information from stack.yaml will be available to stack creators
- [ ] Agree on keys/names used for the template variable
- [ ] Agree how will users specify custom template strings
- [ ] Update stack package command to do this template parsing/replacement before building the stack.
- [ ] Validate this works on all platforms and retains file permissions

neeraj-laad:
We need the following variables as a minimum:

- StackID
- StackName
- SackMajorVersion
- StackMinorVersion
- StackPatchVersion
- StackImageNamespace

kylegc:
Lots of this information overlaps with the image labels we have added. We don't necessarily need to use the same variable names, but we should reuse code and values as much as possible. For instance, we calculate a created timestamp for the labels, if we add that as a template variable, it should use the **exact same** timestamp, not one that is off by even a couple milliseconds.

kylegc:
I definitely think we should have a way for the stack developer to define their own arbitrary template variable=value pairs for replacement. These would likely go in the stack.yaml under a defined array. 
We could consider adding these to the image labels as well.

kylegc:
If the stack developer is using templating within the stack, I don't like making them wrap their templates with `{{"{{.UserVariableName}}"}}`. That could become very cumbersome for the stack developer.

Thus, I think we should use a different delimiter for the stack variables, but one similar enough that it still looks like a template. Here are some ideas:
- `{{{.StackId}}}`
- `{{$.StackId}}`  - yes I do believe the starting delimiter can be different from the ending
- `{{appsody.StackId}}`

neeraj-laad:
@kylegc @CameronMcWilliam I've been thinking more on this and have discussed this with a few others to see what they find intuitive. 

The general feedback I'm getting at the moment is that users would prefer to use something that is simple and standard. That way they will know exactly how to work around limitations like escaping. 

Most did like {{stack`.id`}} as it looks meaningful to user and could provide some scoping if we wanted to provide templating from other files in future. But there are 2 big negatives too:
- IDEs might format `{{ stack.id }}` with spaces and that might break things.
- Also using other go templating capabilities like logic etc will be cumberson.

[...]

PR #795, opened 2019-12-11
Change templating delimiters to use {{.stack}}

CameronMcWilliam:
This is not a breaking change to existing stacks using templating.  This is a temporary solution due to many errors happening with the new templating feature.  This means that it will avoid code collisions from other templating libraries that use `{{` `}}` as delimiters, as the new delimiters are `{{.stack` `}}`.

### DIFF
--- a/cmd/stack_package.go
+++ b/cmd/stack_package.go
@@ -223,7 +223,7 @@ The packaging process builds the stack image, generates the "tar.gz" archive fil
 
 			// tag with the full version then mojorminor, major, and latest
 			cmdArgs := []string{"-t", buildImage}
-			var semver = make(map[string]string)
+			semver := templateMetadata["semver"].(map[string]string)
 			cmdArgs = append(cmdArgs, "-t", namespaceAndRepo+":"+semver["majorminor"])
 			cmdArgs = append(cmdArgs, "-t", namespaceAndRepo+":"+semver["major"])
 			cmdArgs = append(cmdArgs, "-t", namespaceAndRepo)
--- a/cmd/stack_package.go
+++ b/cmd/stack_package.go
@@ -223,7 +223,7 @@ The packaging process builds the stack image, generates the "tar.gz" archive fil
 
 			// tag with the full version then mojorminor, major, and latest
 			cmdArgs := []string{"-t", buildImage}
-			semver := make(map[string]interface{})["semver"].(map[string]string)
+			var semver = make(map[string]string)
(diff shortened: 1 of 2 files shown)


CameronMcWilliam:
**Why this change has come about:**

Templating was introduced as part of issue #560 in the PR #528 to allow stack creators ease-of-use in using common values across a stack and keeping them in sync.  

Since introducing the templating feature, many have expressed complaints about applying templating to an entire stack and how it can be irritating how if code collisions occur with other templating libraries using the same `{{` `}}` delimiters.  We offered work arounds for this as to avoid it crashing out, such as putting `{{"{{ }}"}}` around variables, but in reality this is not practical, as it can be time consuming to going through source code and individually change wherever these occur, and it looks ugly too.

Another instance of where code collisions frequently happened, is during local development of a stack. For example if you were developing a node stack locally.  If you tried to apply templating to this stack that had node modules installed, and somewhere deep inside those modules, a dependency uses the `{{ }}` delimiters, it would crash out and cause an error, meaning your stack could not be packaged, and you are unable to edit the source code.  The only way round this is to delete the modules before every  package, which again is time consuming.

There were further discussions of how to tackle this problem, such as including an opt-in design or an opt-out design for files that you want or don't want templating applied to respectively.  This would solve issues of unwanted templating applied to files that already use templating widely.  However, we decided against this, as it means an extra step for the stack developer, and we want it to be as simple as possible to implement templating: declare it in your stack, value appears after packaging.  That's the premise we went with.

The middle ground of all of this is to have the delimiters being `{{.stack }}`.  This means that when `stack package` parses the files for templating, it is only looking for variables that begin with `{{.stack`, meaning templating variables prefixed with only `{{` are ignored, therefore leaving them intact and not throwing up an error when a stack is packaged.  This also means that stacks currently using the templating, will not be broken by this change.  It is important now to document that you must **not** include spaces in-between `{{` and `.stack`, otherwise templating will not be applied.

kylegc:
Thanks for the details! Just one more thing to mention from #560 
We did originally discuss a delimiter such as `{{.stack` , however it was decided not to be use then because:
- IDEs might format {{ .stack.id }} with spaces and that might break things.
- Also using other go templating capabilities like logic etc will be cumbersome.

Are we willing to say now that the other issues above are worse than these two issues? Therefore we will take the tradeoff with the lesser impact and implement this?

I tend to think so and am in favor of this change, but would like to hear other opinions too.

Issue #463, opened 2019-10-16
Standardize the help text for the CLI commands (and add examples)


helenmasters:
The number of Appsody CLI commands has grown significantly recently and the help text for these commands could do with some standardizing and editing. I also think it would help users if we added examples (in addition to the Usage: section).

I'm putting together a proposal of what this might look like for `appsody stack create` if we can find consensus this can be rolled out across the other commands.

helenmasters:
@neeraj-laad, suggested --help text for the `appsody stack create` command is attached below:

[Sample CLI help for stack create.docx](https://github.com/appsody/appsody/files/3734715/Sample.CLI.help.for.stack.create.docx)

doveye:
- I think the examples would be best on a line by themselves, with explanation underneath.
- Is the "string" part of the flag necessary? I'm thinking it might be clearer to say what's needed there rather than in the explanation
- I think flag explanations might be better distinguised by a capital letter at the start
Questions for global flags:
- What's the config file (do we need to mention more about it here or can we assume users will know what it is?)
- What's dry run mode (and ditto re do we need to mention more about it here)?
- What's the name of the verbose log file? In case the user needs to find it amongst a load of other log files)

--------

```
Examples:
`appsody stack create myStack`
Creates a copy of the example stack, “samples/sample-stack”, called myStack.

`appsody stack create myStack --copy incubator/nodejs-express`
Creates a copy of the Node.js.Express stack, called myStack.

Usage:
 `appsody stack create <name> [flags]`
where <name> starts with a letter or number, and can contain only lowercase letters, numbers, or dashes.

Flags:
  --copy <repository_name>/<stack_name>   Copies the specified stack. The default value is “samples/sample-stack”.
   -h, --help                             Prints this help information

Global Flags:
   --config <filename>      Specifies a configuration file. The default file is $HOME/.appsody/.appsody.yaml
   --dryrun                 Turns on dry run mode
   -v, --verbose            Turns on debug output and logging to a file in $HOME/.appsody/logs
```

helenmasters:
Note to self, the example stack name has capital letters which contravenes our stack naming convention so I need to change it!

helenmasters:
I hope this version addresses all the requested changes (excluding questions from @doveye about global flags). I've changed the tense on the flags back to 'Copy', 'Print' etc as this seems to be the tense universally adopted by other CLIs (Docker, man pages, Helm etc).

[...]

PR #805, opened 2019-12-12
refactor global flags definitions

skoh7645:
Refactor global flags definitions as part of https://github.com/appsody/appsody/issues/463

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -92,9 +92,9 @@ Complete documentation is available at https://appsody.dev`,
 		//Run: no run action for the root command
 	}
 
-	rootCmd.PersistentFlags().StringVar(&rootConfig.CfgFile, "config", "", "Location of Appsody configuration file to use. (default '$HOME/.appsody/.appsody.yaml')")
-	rootCmd.PersistentFlags().BoolVarP(&rootConfig.Verbose, "verbose", "v", false, "Prints out more detailed logs on the console, and logs them to a file in $HOME/.appsody/logs.")
-	rootCmd.PersistentFlags().BoolVar(&rootConfig.Dryrun, "dryrun", false, "Lists terminal commands that would be run without executing them.")
+	rootCmd.PersistentFlags().StringVar(&rootConfig.CfgFile, "config", "", "The absolute path to the Appsody config file. Use this option when you want to specify your own, customized config file. (default '$HOME/.appsody/.appsody.yaml')")
+	rootCmd.PersistentFlags().BoolVarP(&rootConfig.Verbose, "verbose", "v", false, "Prints more detailed log output, to the console and to a file in $HOME/.appsody/logs")
+	rootCmd.PersistentFlags().BoolVar(&rootConfig.Dryrun, "dryrun", false, "Shows the commands that are called by this command, without running them.")
 
 	// parse the root flags and init logging before adding all the other commands in case those log messages
 	rootCmd.SetArgs(args)
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -92,7 +92,7 @@ Complete documentation is available at https://appsody.dev`,
 		//Run: no run action for the root command
 	}
 
-	rootCmd.PersistentFlags().StringVar(&rootConfig.CfgFile, "config", "", "The absolute path to the Appsody config file. Use this option when you want to specify your own, customized config file. (default '$HOME/.appsody/.appsody.yaml')")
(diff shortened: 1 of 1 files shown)


cmd/root.go
```diff
@@ -92,9 +92,9 @@ Complete documentation is available at https://appsody.dev`,
 		//Run: no run action for the root command
 	}
 
-	rootCmd.PersistentFlags().StringVar(&rootConfig.CfgFile, "config", "", "config file (default is $HOME/.appsody/.appsody.yaml)")
-	rootCmd.PersistentFlags().BoolVarP(&rootConfig.Verbose, "verbose", "v", false, "Turns on debug output and logging to a file in $HOME/.appsody/logs")
-	rootCmd.PersistentFlags().BoolVar(&rootConfig.Dryrun, "dryrun", false, "Turns on dry run mode")
+	rootCmd.PersistentFlags().StringVar(&rootConfig.CfgFile, "config", "", "Location of Appsody config files to use. (default '$HOME/.appsody/.appsody.yaml')")
```
helenmasters: Does `--config` allow you to change the location and name of the default Appsody config file, or just the location (i.e. does the file have to be called .appsody.yaml)?

helenmasters: Ok, so the expectation is that you make your own configuration file, called whatever you choose, and use `--config` to point at your own configuration file (so this definitely shouldn't say "files").

doveye: Judging by the earlier conversation, it sounds like this is actually a path to a single file? Is it relative or absolute? Here's a possible alternative:
"The absolute path to the Appsody config file. Use this option when you want to specify your own, customized config file, instead of the default: '$HOME/.appsody/.appsody.yaml'."
(That's assuming there isn't an explanation nearby for when/why you would want to use this option :)

cmd/root.go
```diff
@@ -92,9 +92,9 @@ Complete documentation is available at https://appsody.dev`,
...
+	rootCmd.PersistentFlags().BoolVarP(&rootConfig.Verbose, "verbose", "v", false, "Prints out more detailed logs on the console, and logs them to a file in $HOME/.appsody/logs.")
```
doveye: Assuming the same thing is sent to the console and to the file, I suggest:

"Prints more detailed log output, to the console and to a file in $HOME/.appsody/logs."

Do we need to mention the name of the log file or will it be obvious, eg does it have a date/timestamp, or something to say which stack/app/whatever this log file came from?

cmd/root.go
```diff
@@ -92,9 +92,9 @@ Complete documentation is available at https://appsody.dev`,
...
+	rootCmd.PersistentFlags().BoolVar(&rootConfig.Dryrun, "dryrun", false, "Lists terminal commands that would be run without executing them.")
```
doveye: Not sure I understand this - is the following accurate?
"Shows the commands that are called by this command, without running them."

cmd/root.go
skoh7645: Yes, the log file will have the timestamp on it

cmd/root.go
```diff
@@ -92,9 +92,9 @@ Complete documentation is available at https://appsody.dev`,
 		//Run: no run action for the root command
 	}
 
-	rootCmd.PersistentFlags().StringVar(&rootConfig.CfgFile, "config", "", "config file (default is $HOME/.appsody/.appsody.yaml)")
-	rootCmd.PersistentFlags().BoolVarP(&rootConfig.Verbose, "verbose", "v", false, "Turns on debug output and logging to a file in $HOME/.appsody/logs")
-	rootCmd.PersistentFlags().BoolVar(&rootConfig.Dryrun, "dryrun", false, "Turns on dry run mode")
+	rootCmd.PersistentFlags().StringVar(&rootConfig.CfgFile, "config", "", "The absolute path to the Appsody config file. Use this option when you want to specify your own, customized config file. (default '$HOME/.appsody/.appsody.yaml')")
+	rootCmd.PersistentFlags().BoolVarP(&rootConfig.Verbose, "verbose", "v", false, "Prints more detailed log output, to the console and to a file in $HOME/.appsody/logs")
```
helenmasters: V minor but as you've taken out the suggested wording "instead of the default" (presumably to match the format of the other Appsody default values which are added in brackets) please can you remove the full stop before the '(' so it reads:

... Use this option when you want to specify your own, customized config file (default '$HOME/.appsody/.appsody.yaml')

PR #806, opened 2019-12-12
Code coverage: version.go

CameronMcWilliam:
Simple test that just calls `appsody version` Puts coverage up to 100%


cmd/version_test.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"testing"
+
+	"github.com/appsody/appsody/cmd/cmdtest"
+)
+
+func TestVersion(t *testing.T) {
+
+	args := []string{"version"}
+	_, err := cmdtest.RunAppsodyCmd(args, ".", t)
```
EnriqueL8: This will need to change, since the sandboxing PR #788 will remove `RunAppsodyCmd`.

EnriqueL8: Change to `RunAppsody(sandbox)`

PR #810, opened 2019-12-12
Tests for utils.go

upLukeWeston:
A lot of the tests can't currently be ran on Travis due to the use of Kubernetes, but they're written and tested locally ready to be uncommented when the functionality is implemented (if you could still review them or pull them down to test locally that'd be fantastic but I wouldn't say it's a definite requirement).

The tests for Move/Copy and Dryrun are still applicable, though.


kylegc:
Thanks for writing the tests, even the ones that can't be run on travis. Instead of commenting them out, we could implement some sort of toggle, such as one or more environment variables, that turns these tests on and off. Take a look at `t.Skip` and think about this a little.

chilanti:
@neeraj-laad - is this another case where the CLA doesn't get reported as signed?

PR #816, opened 2019-12-16
Propagate stack package verbose option

### LINKED ISSUE #814: appsody stack validate always uses verbose options for run/debug/test
`appsody stack validate` always invokes commands with the verbose flag, like this: 

[/usr/bin/appsody -v init dev.local/java-spring-boot2 kotlin]

I would like `appsody stack validate` to run without any verbose output, unless it is invoked with the verbose option.

Plain invocation (not verbose)
```
$ appsody stack validate
...
/usr/bin/appsody stack package --image-namespace [secure] --image-registry docker.io
...
/usr/bin/appsody init dev.local/java-spring-boot2 kotlin
```

and then with verbose option: 
```
$ appsody stack validate -v
...
/usr/bin/appsody stack package -v --image-namespace [secure] --image-registry docker.io
...
/usr/bin/appsody init -v dev.local/java-spring-boot2 kotlin
```

ebullient:
Propagate the verbose option through to appsody command execution during stack validate.

Fixes #814

### DIFF
--- a/cmd/stack_validate.go
+++ b/cmd/stack_validate.go
@@ -92,7 +92,7 @@ Runs the following validation tests against the stack and its templates:
 
 			// lint
 			if !noLint {
-				_, err = RunAppsodyCmdExec([]string{"stack", "lint"}, stackPath)
+				_, err = RunAppsodyCmdExec([]string{"stack", "lint"}, stackPath, rootConfig)
 				if err != nil {
 					//logs error but keeps going
 					rootConfig.Error.Log(err)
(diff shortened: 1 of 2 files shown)


cmd/stack_validate.go
```diff
@@ -92,7 +92,7 @@ Runs the following validation tests against the stack and its templates:
 
 			// lint
 			if !noLint {
-				_, err = RunAppsodyCmdExec([]string{"stack", "lint"}, stackPath)
+				_, err = RunAppsodyCmdExec([]string{"stack", "lint"}, stackPath, rootConfig.Verbose)
```
chilanti: Wondering if we should pass just `rootConfig` and let the function extract `Verbose` - rather than explicitly passing in a boolean. This has been a common practice throughout Appsody CLI. Unless there are strong objections, I'd request this change.

Issue #789, opened 2019-12-10
Check pre-reqs should check buildah on the appropriate code path


chilanti:
**Is your feature request related to a problem? Please describe.**
Currently we only check whether docker is available and running on the docker path.
No pre-req check is done for buildah.

**Describe the solution you'd like**
Check whether buildah is installed when appsody runs on the buildah path

PR #818, opened 2019-12-17
Add prereq check for buildah when used

Kamran64:
Fixes: https://github.com/appsody/appsody/issues/789

- Checks prerequisite that buildah is running if used.


Kamran64:
@chilanti I don't have Buildah running on my machine since I'm using mac, please could you verify this works :) Thanks.

cmd/utils.go
```diff
@@ -686,6 +686,13 @@ func CopyDir(log *LoggingConfig, fromDir string, toDir string) error {
 // CheckPrereqs checks the prerequisites to run the CLI
 func CheckPrereqs(config *RootCommandConfig) error {
 	if config.Buildah {
+		buildahCmd := "buildah"
+		buildahArgs := []string{"containers"}
+		checkBuildahCmd := exec.Command(buildahCmd, buildahArgs...)
+		_, buildahCmdErr := checkBuildahCmd.Output()
+		if buildahCmdErr != nil {
+			return errors.New("buildah does not seem to be installed or running - failed to execute buildah containers")
```
chilanti: I would suggest changing the message slightly and returning the error too. The message could be something like:
```
errors.Errorf("buildah does not seem to be installed or capable of running in this environment - failed to execute buildah containers: %v", buildahCmdErr)
```
`buildah containers` is a good test - but it may fail when the docker container that runs buildah isn't running on the correct platform (for example, macOS). In that case, you get an error even if buildah is installed correctly.

PR #830, opened 2020-01-06
Changed templating-data in stack.yaml to be stack-variables

CameronMcWilliam:
Further reduction in references to the word templating as to avoid confusion over conflicting pre-existing name templates.  In the `stack.yaml` where a user declares their custom variables, it is now declared using `stack-variables` instead.  The website and stacks already using templating need to be updated accordingly.

### DIFF
--- a/cmd/stack_package.go
+++ b/cmd/stack_package.go
@@ -38,7 +38,7 @@ type StackYaml struct {
 	Language        string `yaml:"language"`
 	Maintainers     []Maintainer
 	DefaultTemplate string            `yaml:"default-template"`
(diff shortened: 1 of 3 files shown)


kylegc:
This is a breaking change, so we need to be careful. Do you know how many stacks are using this?

CameronMcWilliam:
> This is a breaking change, so we need to be careful. Do you know how many stacks are using this?

Yeah, might be best to leave it for another release, we could do something where we allow both `templating-data` and `stack-variables` as to avoid breaking change, encouraging users to use `stack-variables`.  But this needs more thought.

Issue #827, opened 2020-01-05
jsonpath results in invalid query


haf-tech:
**Describe the bug**
The resulting jsonpath is invalid. the last char is an ´ instead of '
`Running command: kubectl get pods -o=jsonpath='{.items[?(@.metadata.labels.name=="appsody-operator")].metadata.namespace}’ --all-namespaces`

Verify both lines
correct: https://github.com/appsody/appsody/blob/3582a88d037ce6ab372fe6ce546e320c7e5493ae/cmd/operator_utils.go#L134

incorrect: https://github.com/appsody/appsody/blob/3582a88d037ce6ab372fe6ce546e320c7e5493ae/cmd/operator_utils.go#L153

**To Reproduce**
```
appsody deploy -t image-registry.openshift-image-registry.svc:5000/${PRJ_NAME}/k101-nodejs-express:v0.1 --namespace ${PRJ_NAME} --no-build
Attempting to get resource from Kubernetes ...
Running command: kubectl get pods -o=jsonpath='{.items[?(@.metadata.labels.name=="appsody-operator")].metadata.namespace}’ --all-namespaces
[Error] kubectl command failed: '’
```

**Expected behavior**
Valid k8s/jsonpath command

**Actual behaviour**
jsonpath is invalid. after correction of the last char to ' is the query valid

**Environment Details (please complete the following information):**
 - Version 0.5.3

**If applicable please specify:**
 - CLI version: appsody 0.5.3
 - Stack you are using: nodejs-express

**Screenshots**
appsody command
`appsody deploy -t image-registry.openshift-image-registry.svc:5000/${PRJ_NAME}/k101-nodejs-express:v0.1 --namespace ${PRJ_NAME} --no-build
Attempting to get resource from Kubernetes ...
Running command: kubectl get pods -o=jsonpath='{.items[?(@.metadata.labels.name=="appsody-operator")].metadata.namespace}’ --all-namespaces
[Error] kubectl command failed: '’`

**Additional context**
n/a

PR #834, opened 2020-01-07
Fix generated jsonpath. 

haf-tech:
Fix generated jsonpath. Fixes: https://github.com/appsody/appsody/issues/827


haf-tech:
@kylegc this is the new one, with the same change :-)

PR #839, opened 2020-01-09
Buildah support for stack package

CameronMcWilliam:
Adds the option to use `--buildah` and --buildah-options` flags when using `stack package` command.

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -54,7 +54,7 @@ var supportedKubeLabels = []string{
 	"app.appsody.dev/name",
 }
 
-func CheckDockerBuildOptions(options []string) error {
+func checkBuildOptions(options []string) error {
 	buildOptionsTest := "(^((-t)|(--tag)|(--help)|(-f)|(--file))((=?$)|(=.*)))"
 
 	blackListedBuildOptionsRegexp := regexp.MustCompile(buildOptionsTest)
@@ -162,7 +162,7 @@ func build(config *buildCommandConfig) error {
 
 	if buildOptions != "" {
 		options := strings.Split(buildOptions, " ")
-		err := CheckDockerBuildOptions(options)
+		err := checkBuildOptions(options)
 		if err != nil {
 			return err
 		}
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -130,6 +130,11 @@ func build(config *buildCommandConfig) error {
 		buildOptions = strings.TrimSpace(config.buildahBuildOptions)
 	}
 
+	// Issue 529 - if you specify --push or --push-url without --tag, error out
+	if config.tag == "" && (config.push || config.pushURL != "") {
+		return errors.New("Cannot specify --push or --push-url without a --tag")
+	}
+
 	extractConfig := &extractCommandConfig{RootCommandConfig: config.RootCommandConfig}
 
 	projectName, perr := getProjectName(config.RootCommandConfig)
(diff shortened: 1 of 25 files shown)


cmd/build.go
```diff
@@ -54,7 +54,7 @@ var supportedKubeLabels = []string{
 	"app.appsody.dev/name",
 }
 
-func checkDockerBuildOptions(options []string) error {
+func CheckDockerBuildOptions(options []string) error {
```
EnriqueL8: Just a suggestion: we could change the name of this function

kylegc:
This is looking good. One thought, if we add `--buildah-options` flag, we should probably also add the `--docker-options` flag as well, just like the build command has for consistency.

How much testing has been done on this with buildah?

kylegc:
Also, don't worry too much about the build failure, it is a separate issue which I think I've got a fix for it.

CameronMcWilliam:
> Also, don't worry too much about the build failure, it is a separate issue which I think I've got a fix for it.

ahhh, good ! thought it was a bit random.  yeah I'll add in docker options now.

EnriqueL8:
> This is looking good. One thought, if we add `--buildah-options` flag, we should probably also add the `--docker-options` flag as well, just like the build command has for consistency.
> 
> How much testing has been done on this with buildah?

I was also thinking about the buildah test, we could add buildah to the linux test and implement a series of functional tests.

cmd/stack_package.go
```diff
@@ -93,12 +100,34 @@ The packaging process builds the stack image, generates the "tar.gz" archive fil
...
+			}
+
+			if config.dockerBuildOptions != "" {
+				if config.Buildah {
+					err = errors.New("Cannot specify --docker-options flag with --buildah")
+				}
+				buildOptions = strings.TrimSpace(config.dockerBuildOptions)
+			}
+
+			if err != nil {
```
kylegc: Why not return the errors as they are happening above?

functest/package_test.go
```diff
@@ -49,3 +49,61 @@ func TestPackage(t *testing.T) {
...
+	} else {
+		t.Logf("Copied %s to %s", stackDir, sandbox.ProjectDir)
+	}
+
+	// Because the 'starter' folder has been copied, the stack.yaml file will be in the 'starter'
+	// folder within the temp directory that has been generated for sandboxing purposes, rather than
+	// the usual core temp directory
+	sandbox.ProjectDir = filepath.Join(sandbox.ProjectDir, "starter")
+
+	args := []string{"stack", "package", "--docker-options", "-q"}
```
EnriqueL8: It could be great if we could test that the `docker` command actually gets the `-q` (quiet flag). I believe here we are just testing that the `stack package --docker-options -q` command is not falling, and not that the docker options are correctly being passed down?

EnriqueL8: After checking the `docker_options_test`, we currently only run the Appsody commands. This could be a future improvement.

functest/package_test.go
```diff
@@ -49,3 +49,61 @@ func TestPackage(t *testing.T) {
...
+	_, err = cmdtest.RunAppsody(sandbox, args...)
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+/*
+func TestPackageBuildahOptions(t *testing.T) {
```
EnriqueL8: Maybe we could set up the linux travis builds to install buildah?

PR #840, opened 2020-01-09
Fix build failures by setting up the sandbox and sub-tests properly for parallel testing

### LINKED ISSUE #826: Investigate why getRepoListOutput doesn't always return the expected result (and fix it)
**Describe the bug**
It seems that `getRepoListOutput` intermittently returns an unexpected number of repos (as if there was a thread safety issue) during Travis testing.

**Expected behavior**
`getRepoListOutput` should work in isolation during Travis test execution. 

**Actual behaviour**
It seems that - occasionally - the function returns an unexpected number of repos, as if some other thread had modified the repository file.

 - CLI version:
 - Stack you are using:

**Screenshots**

**Additional context**

kylegc:
Fixes #826


kylegc:
This PR reverts #825 and fixes it properly

PR #844, opened 2020-01-09
created developer best practices

tnixa:


docs/testing/DEVELOPER-BEST-PRACTICES.md
```diff
@@ -0,0 +1,90 @@
...
+* the github webpage for the upstream repo should show an indicator after pushing your changes, click the "Compare & pull request" button
+![image info](pull-request.png)
+
+* if you would like your code changes to be known/tracked but are not ready for them to be reviewed/approved you can create a draft pull request using the pull-down
+![image info](draft-pull-request.png)
+
+## sync changes from upstream to fork master branch
+* git checkout master
+* git pull master
+* git checkout <branch>
```
kylegc: `<branch>` doesn't show up in markdown preview. I think you need to escape the brackets

docs/testing/DEVELOPER-BEST-PRACTICES.md
```diff
@@ -0,0 +1,90 @@
...
+* create a branch, ex:
+    * git checkout -b testbranch
+* set tracking of branch
+    * make changes to the files you want
+* stage the changes
+    * git add .
+* commit the changes
+    * git commit -m "commit message"
+* push the changes and set tracking of the branch
+    * git push -u tnixa testbranch
```
kylegc: Maybe add a comment here that the `-u` and after is only needed the first time you push the branch. After that, `git push` is good enough.

docs/testing/DEVELOPER-BEST-PRACTICES.md
```diff
@@ -0,0 +1,90 @@
+# Developer Best Practices
```
helenmasters: This file is called developer best practices but it seems to contain - something about go, a set of instructions on installing and configuring git, and then are we really telling users how we suggest that they work with git within the Appsody project (without actually saying that)? What is the intent of this information - it isn't what I think of as "best practices"?

docs/testing/DEVELOPER-BEST-PRACTICES.md
```diff
@@ -0,0 +1,90 @@
...
+* git checkout <branch>
+* git merge master (some people prefer rebase here, but that should only be used if you haven't created a PR yet)
+## testing:
+* test your code changes against the test bucket
+    * make test
+* build the appsody binary files for the various platforms
+    * make build-darwin (mac)
+    * make build-windows (windows)
+    * make build-linux (linux)
+        * this creates an apposdy binary in the build dir of the repo
```
helenmasters: apposdy?

docs/DEVELOPER-GUIDE.md
```diff
@@ -39,7 +43,8 @@
     * git remote add upstream git@github.com:appsody/appsody.git
 * set master to point to upstream
     * git branch master -u upstream/master
```
helenmasters: If I run this on a clean setup following these instructions I get:
```
error: the requested upstream branch 'upstream/master' does not exist
hint:
hint: If you are planning on basing your work on an upstream
hint: branch that already exists at the remote, you may need to
hint: run "git fetch" to retrieve it.
hint: 
hint: If you are planning to push out a new local branch that
hint: will track its remote counterpart, you may want to use
hint: "git push -u" to set the upstream config as you push.```

File: docs/DEVELOPER-GUIDE.md [hunk 6]
```diff
@@ -39,7 +43,8 @@
     * git remote add upstream git@github.com:appsody/appsody.git
```
helenmasters: Given https://help.github.com/en/github/using-git/which-remote-url-should-i-use#cloning-with-https-urls-recommended is it not clearer to use:
`git remote add upstream https://github.com/appsody/appsody.git`
?
(This is certainly referenced in other tutorials on github.com such as https://help.github.com/en/github/using-git/removing-a-remote)

docs/DEVELOPER-GUIDE.md
helenmasters: Do we need this level of detail in this guide? Surely we could link out to github.com for some of this?

docs/DEVELOPER-GUIDE.md
```diff
@@ -0,0 +1,99 @@
...
+![image info](draft-pull-request.png)
+
+## sync changes from upstream master
+This section describes how to sync the upstream master and your branch. You may want to do this to check for merge conflicts.
+* git checkout master
+* git pull master
+* git checkout `<branch>`
+* git merge master (some people prefer rebase here, but that should only be used if you haven't created a PR yet)
+## testing:
+This section describes various ways to test both your code and other's code.
```
helenmasters: Suggest reword to - "There are various ways to test your own code or someone else’s pull request (or branch)".

docs/DEVELOPER-GUIDE.md
```diff
@@ -0,0 +1,99 @@
...
+    * make build-linux (linux)
+        * this creates an apposdy binary in the build dir of the repo
+* test appsody binary file
+    * fully qualify the appsody binary file for the command you want to test, ex:
+        * ./Users/tnixa/go/src/github.com/appsody/appsody/build/appsody-0.0.0-darwin-amd64 list
+    * to make this easier to run from any terminal window
+        * cd ~/go/bin
+        * ln -s ../src/github.com/appsody/appsody/build/appsody-0.0.0-darwin-amd64 appsody-local
+        * now you can call appsody-local from any terminal (assumes GOPATH/bin is on your PATH)
+* test someone's PR/branch:
```
helenmasters: test someone else's PR/branch:

docs/DEVELOPER-GUIDE.md
```diff
@@ -0,0 +1,99 @@
...
+This section describes various ways to test both your code and other's code.
+* test your code changes against the test bucket
+    * make test
+* build the appsody binary files for the various platforms
+    * make build-darwin (mac)
+    * make build-windows (windows)
+    * make build-linux (linux)
+        * this creates an apposdy binary in the build dir of the repo
+* test appsody binary file
+    * fully qualify the appsody binary file for the command you want to test, ex:
```
helenmasters: Please can you replace ex: with e.g.: or for example:

docs/DEVELOPER-GUIDE.md
```diff
@@ -0,0 +1,99 @@
...
+* set master to point to upstream
+    * git branch master -u upstream/master
+## branch setup
+This section describes how to create and setup a branch for working on a new issue/PR
+* checkout and update master
+    * git checkout master
+    * git pull master
+* create a branch, ex:
+    * git checkout -b testbranch
+* set tracking of branch
```
helenmasters: why is this called set tracking of branch - isn't this making the changes?

kylegc:
> I think it will be best if this document described Appsody specific aspects detail but provides only general guidance to non-Appsody specific aspects.
> 
> So, providing exact commands on how to run tests locally is great.. but perhaps we should refrain from specifying exactly how contributors set up their git environment. We can describe what we want tom to achieve - fork appsody/stacks repository, create a new branch for every feature/defect, submit a PR etc.
> 
> As long as we get a PR with all the information that we are after, the rest does not really matter as much to Appsody.

While it's true that contributors can setup their local git however they want, there is some value in having the same setup when it comes to discussions and debugging issues. I also think this info is valuable for new team members who are not familiar with git. 

I don't want to throw this away, so if we don't actually want it in the appsody repo, maybe Terrence can host in his personal github repo or an alternate location.

tnixa:
I revised the guidelines per @neeraj-laad . I'll leave specific instructions in the box note.

docs/DEVELOPER-GUIDE.md
```diff
@@ -0,0 +1,38 @@
...
+
+## testing:
+There are various ways to test your own code or someone else’s pull request (or branch):
+* test your code changes against the test bucket
+    * make test
+* build the appsody binary files for the various platforms
+    * make build-darwin (mac)
+    * make build-windows (windows)
+    * make build-linux (linux)
+        * this creates an apposdy binary in the build dir of the repo
```
helenmasters: apposdy

docs/DEVELOPER-GUIDE.md
```diff
@@ -0,0 +1,38 @@
...
+* test appsody binary file
+    * fully qualify the appsody binary file for the command you want to test, for example:
```
helenmasters: Having read the building from source instructions for Appsody I'm surprised that we don't suggest here that the user specifies the target `localbin-{target system}` as it'd save the requirement for the linking you're suggesting? See https://github.com/appsody/appsody/blob/master/build.md#running-the-locally-built-cli

Or are we expecting that developers would prefer to have an executable named `appsody-local` to another called `appsody`?

docs/DEVELOPER-GUIDE.md
```diff
@@ -0,0 +1,38 @@
...
+## testing:
+There are various ways to test your own code or someone else’s pull request (or branch):
+* test your code changes against the test bucket
+    * make test
+* build the appsody binary files for the various platforms
+    * make build-darwin (mac)
+    * make build-windows (windows)
+    * make build-linux (linux)
+        * this creates an apposdy binary in the build dir of the repo
+* test appsody binary file
```
helenmasters: This looks more like a command than a title, if it's meant to be a section title it needs to be something like "test the local appsody binary file"

docs/DEVELOPER-GUIDE.md
tnixa: I think the idea is to have `appsody` run the released/installed version and then `local-appsody` would run the dev/test version. I'm not sure what `localbin-{target system}` does. Maybe @kylegc can shed some light...

kylegc: I forgot about the build.md doc. There are overlaps between it and this, but I think that is okay as both docs serve a different purpose.

The `localbin` part almost solves the same problem, except as Terrence pointed out, we want the binary to have a different name so the user can have a stable `appsody` command and an experimental `appsody-local` command at the same time.

Issue #813, opened 2019-12-16
Operator fails install to a specified namespace when query of all-namespaces is restricted


leochr:
@richard-trotter commented on [Mon Dec 16 2019](https://github.com/appsody/appsody-operator/issues/201)

## Bug Report

I recently received notice of some RHOCP enablement resources, including a course on OpenShift 4. The course includes labs that use a provided RHOCP shared cluster. Since IBM Public Cloud does not provide an OpenShift Lite (free) option, this cluster seemed like a good place to evaluate deployment of an AppsodyApplication to an OpenShift cluster. Of course this requires installation of the Appsody Operator. 

I can only install resources to my own namespace, and installation of Appsody to my own namespace fails.

It looks like this is failing while trying to determine if the operator is already installed, and is failing because 'get pods' is run with '-all-namespaces' but I don't have auth (in this shared cluster) to query all namespaces.  Seems like a bug, that '-all-namespaces'  is used instead of '--namespace eda4-appsody-deployment '.
$ appsody operator install --namespace eda4-appsody-deployment --verbose
...
[Debug] watchNamespace is:  eda4-appsody-deployment
[Info] Running command: kubectl get deployments -o=jsonpath='{.items[?(@.metadata.name=="appsody-operator")].metadata.namespace}' -n eda4-appsody-deployment
[Info] There are no deployments with appsody-operator
[Debug] Looking for an operator matching watchspace: eda4-appsody-deployment
[Info] Running command: kubectl get pods -o=jsonpath='{.items[?(@.metadata.labels.name=="appsody-operator")].metadata.namespace}' --all-namespaces
[Error] kubectl command failed: ''

### Environment

* Appsody CLI version: 0.5.3

* Appsody Stack: java-spring-boot2

* Kubernetes version information: 
$ oc version --short
Client Version: v4.0.22
Server Version: v1.13.4+28c3de7

---

@arturdzm commented on [Mon Dec 16 2019](https://github.com/appsody/appsody-operator/issues/201#issuecomment-566174662)

The installation code you referring to is in Appsody CLI
https://github.com/appsody/appsody

chilanti:
This is a valid issue. Appsody tries to do "the right thing" by preventing the installation of redundant operators that concurrently watch the same namespace, but in order to do so, it requires the ability of detecting (i.e. querying) operators all around the cluster. In most shared cluster environments, this isn't viable, because of RBAC constraints. 

Possible ways to address this behavior:
[...]

PR #845, opened 2020-01-10
Add a no-check flag to to suppress all-namespaces kubectl command

Kamran64:
Fixes: https://github.com/appsody/appsody/issues/813

Adds a new flag, `no-check` to `appsody deploy` and `appsody operator install` to suppress `kubectl get pods --all-namespaces` when checking if an operator already exists in the namespace.

Instead only the namespace specified will be queried (In case permission issues occur with `all-namespaces` mentioned in the issue above)

- Opening as draft so that I can write tests

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -204,7 +204,7 @@ Run this command from the root directory of your Appsody project.`,
 	deployCmd.PersistentFlags().BoolVar(&config.knative, "knative", false, "Deploy as a Knative Service")
 	deployCmd.PersistentFlags().StringVar(&config.pushURL, "push-url", "", "Remote repository to push image to.  This will also trigger a push if the --push flag is not specified.")
 	deployCmd.PersistentFlags().StringVar(&config.pullURL, "pull-url", "", "Remote repository to pull image from.")
-	deployCmd.PersistentFlags().BoolVar(&config.noOperatorCheck, "no-operator-check", false, "Suppresses check for operator existing in namespace")
+	deployCmd.PersistentFlags().BoolVar(&config.noOperatorCheck, "no-operator-check", false, "Do not check whether existing operators are already watching the namespace")
 	deployCmd.PersistentFlags().BoolVar(&config.noOperatorInstall, "no-operator-install", false, "Deploy your application without installing the Appsody operator")
 	deployCmd.AddCommand(newDeleteDeploymentCmd(config))
 
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -25,10 +25,10 @@ import (
 
 type deployCommandConfig struct {
 	*RootCommandConfig
-	appDeployFile, namespace, tag, pushURL, pullURL string
-	knative, generate, force, push, nobuild         bool
-	dockerBuildOptions                              string
-	buildahBuildOptions                             string
+	appDeployFile, namespace, tag, pushURL, pullURL  string
+	knative, generate, force, push, nobuild, nocheck bool
+	dockerBuildOptions                               string
+	buildahBuildOptions                              string
 }
 
 func findNamespaceRepositoryAndTag(image string) string {
@@ -97,7 +97,7 @@ The command performs the following steps:
 			}
 
 			// Check for the Appsody Operator
-			operatorExists, existingNamespace, operatorExistsErr := operatorExistsWithWatchspace(config.LoggingConfig, namespace, config.Dryrun)
+			operatorExists, existingNamespace, operatorExistsErr := operatorExistsWithWatchspace(config.LoggingConfig, namespace, config.Dryrun, config.nocheck)
 			if operatorExistsErr != nil {
 				return operatorExistsErr
 			}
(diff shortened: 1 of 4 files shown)


chilanti:
Update after the team discussion - we have  a new proposal to address this scenario:
* Add both a `--no-check` flag and a `--no-operator` flag to `appsody deploy` (these flags are mutually exclusive)
* With `--no-check`, `deploy` will deploy the operator and the app on the target namespace (`watchnamespace`) without querying the entire cluster
* with `--no-operator`, `deploy` will deploy the Appsody application skipping all queries and logic related to the operator.
* with neither flag, `appsody deploy` behaves like it does today - except for the handling of the failure to query the entire cluster for operators. If that failure occurs, a message should be produced that instructs the user to check with the administrator for the presence of an operator that watches the "watch namespace" in question. If one exists, user `--no-operator`, otherwise use `--no-check`.

@neeraj-laad @kylegc @EnriqueL8 @Kamran64  - any additional comments?

kylegc:
Thanks for the writeup Michele. I think that is the best we can do. Just a couple notes:
- `--no-operator` is a bad name, because we still need an operator. How about `--no-operator-install` and `--no-operator-check`? They are longer but clearer.
- `--no-operator-check` still needs to be added to the `operator install` command in addition to `appsody deploy`

cmd/deploy.go
```diff
@@ -96,27 +96,29 @@ The command performs the following steps:
...
+				if operatorExistsErr != nil {
+					return operatorExistsErr
 				}
-			} else {
-				config.Debug.logf("Operator exists in %s, watching %s ", existingNamespace, namespace)
 
+				//kargs := []string{"service/appsody-operator"}
+				//_, err := KubeGet(kargs)
+				// Performing the kubectl apply
+				if !operatorExists {
```
chilanti: @Kamran64 - there's this corner scenario that I'd like you to test if possible. Say you are running `appsody deploy` on namespace `xyz`. The logic now checks whether there's an operator in `xyz` that watches `xyz` (in line 101). If we don't find one, we install it. But what about the case in which there's an operator in `xyz` that watches `abc`? The query in line 101 would report no matches, because the operator doesn't watch `xyz`. We'd then try to install the operator in `xyz` again (this time watching `xyz`). Would that work? Fail?

Kamran64: @chilanti - I tested this just now:

I installed an operator to namespace `xyz` watching `abc`. I then attempted to deploy my project with a target namespace of `xyz` (and had the `no-operator-check` flag set) I.e. `appsody deploy --namespace xyz --no-operator-check`.

As you said, this attempts to install the operator to `xyz` whilst watching `xyz` but the command fails because an operator already exists in namespace `xyz`. Full error:
```
[Error] Failed to install an Appsody operator in namespace xyz watching namespace xyz. Error was: An operator already exists in namespace xyz and it is watching the abc namespace.
```
I reattempted without the `--no-operator-check` flag and saw the same error.

chilanti: Thanks for testing this! 
I think we should handle the situation. If there's an operator in `xyz`, but it's watching another namespace, we should error out with a message that says "There's an appsody operator in xyz, but it's watching another namespace... talk to your cluster adminstrator".

Kamran64: Added in the latest commit - See `operator_install.go` line 86. Thanks @chilanti :)

cmd/operator_install.go
```diff
@@ -83,6 +83,9 @@ func operatorInstall(config *operatorInstallCommandConfig) error {
 		if existingOperatorWatchspace == "" {
 			existingOperatorWatchspace = "all namespaces"
 		}
+		if existingOperatorWatchspace != operatorNamespace {
```
chilanti: `getOperatorWatchspace` returns a string, but it may be a CSV list of watchspaces, which needs to be processed. Appsody will always install an operator with a single watchspace, but Kabanero may install operators with multiple watchspaces. The string needs to be processed using `getWatchSpaces` as we do in `operatorExistsWithWatchspace`.

cmd/operator_install.go
```diff
@@ -83,8 +83,11 @@ func operatorInstall(config *operatorInstallCommandConfig) error {
 		if existingOperatorWatchspace == "" {
 			existingOperatorWatchspace = "all namespaces"
 		}
-		if existingOperatorWatchspace != operatorNamespace {
-			return errors.Errorf("An Appsody operator already exists in namespace %s but is watching another namespace: %s. Contact your cluster administrator for more details.", operatorNamespace, existingOperatorWatchspace)
+		watchSpaces := getWatchSpaces(existingOperatorWatchspace, config.Dryrun)
+		for _, existingWatchspace := range watchSpaces {
+			if existingWatchspace != operatorNamespace {
```
chilanti: I think the logic needs to be adjusted here. We need to error out only if none of the `existingWatchspace` match the operatorWatchspace. Say the operator is installed in `xyz`, and it watches `abc`, `def`, and `xyz`. The code would error out on `abc` != `xyz`, but it shouldn't because the operator does watch `xyz`

Issue #802, opened 2019-12-11
Appsody commands should fail if extra arguments are given


kylegc:
**Is your feature request related to a problem? Please describe.**
Some appsody commands take in positional arguments, and others don't. Either way, we should have each command check that the proper number of arguments exist and fail otherwise. This includes failing if no arguments are expected but one exists.

**Describe the solution you'd like**
All appsody commands check argument length and fail if unexpected arguments exist.

For example:
`appsody stack package ../my/stack/dir/stack` should fail because the path is not expected. `stack package` should be run in the package dir and does not accept positional args.

**Implementation details**
Every command has a code function signature like this:
`RunE: func(cmd *cobra.Command, args []string) error {`

The `args` array should be checked for the proper length on every command.

PR #846, opened 2020-01-10
cli commands will fail if extra arguments are given

skoh7645:
Added checks to CLI so appsody commands will fail if extra propositional arguments are added (and some other minor help text changes)

Fixes: #802

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -79,13 +79,18 @@ func newBuildCmd(rootConfig *RootCommandConfig) *cobra.Command {
 
 By default, the built image is tagged with the project name that you specified when you initialised your Appsody project. If you did not specify a name, the image is tagged with the name of the root directory of your Appsody project.
 
-If you want to push the built image to an image repository using the [--push] options, you must specify the relevant image tag.`,
+If you want to push the built image to an image repository using the [--push] options, you must specify the relevant image tag.
+
+Run this command from the root directory of your Appsody project.`,
 		Example: `  appsody build -t my-repo/nodejs-express --push
   Builds the container image, tags it with my-repo/nodejs-express, and pushes it to the container registry the Docker CLI is currently logged into.
 
   appsody build -t my-repo/nodejs-express:0.1 --push-url my-registry-url
   Builds the container image, tags it with my-repo/nodejs-express, and pushes it to my-registry-url/my-repo/nodejs-express:0.1.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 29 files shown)


cmd/stack_removefromrepo.go
```diff
@@ -54,7 +54,7 @@ The updated repository index file is created in  ~/.appsody/stacks/dev.local dir
 				return errors.New("Required parameter missing. You must specify a repository name and a stack name")
 			}
 			if len(args) > 2 {
-				return errors.New("expected exactly 2 arguments: <repository> <stack>")
+				return errors.Errorf("Expected exactly 2 arguments: %v", cmd.Use)
```
helenmasters: I would rather the word "two" than 2, which would be consistent with the other messages

cmd/repo_set_default.go
```diff
@@ -34,7 +34,7 @@ The default repository is used when you run the "appsody init" command without s
 				return errors.New("you must specify desired default repository")
 			}
 			if len(args) > 1 {
-				return errors.New("expected exactly one arguments: <repository>")
+				return errors.Errorf("Expected exactly one arguments: %v", cmd.Use)
```
helenmasters: arguments -> argument

cmd/repo_set_default.go
```diff
@@ -34,7 +34,7 @@ The default repository is used when you run the "appsody init" command without s
 				return errors.New("you must specify desired default repository")
```
helenmasters: Whilst we're in here this error message doesn't make sense, how about "You must specify the repository to set as the default."

cmd/repo_remove.go
```diff
@@ -36,7 +36,7 @@ You cannot remove the default repository, but you can make a different repositor
 				return errors.New("Error, you must specify repository name")
```
helenmasters: Can't this be "You must specify the repository name." as the error text should be generated?

cmd/build.go
```diff
@@ -89,7 +89,7 @@ Run this command from the root directory of your Appsody project.`,
   Builds the container image, tags it with my-repo/nodejs-express, and pushes it to my-registry-url/my-repo/nodejs-express:0.1.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if len(args) > 0 {
-				return errors.New("expected no additional arguments")
+				return errors.New("Expected no additional arguments")
```
helenmasters: The word "additional" seems strange in this sentence - aren't we saying that this command doesn't accept any arguments?

PR #849, opened 2020-01-13
Appsody init to support --stack-registry

### LINKED ISSUE #696: Appsody init should support --stack-registry flag
The `appsody deploy/build/run/test/debug` commands support a `--stack-registry` flag that allows overriding of the Stack registry URL.
The `appsody init` command does not support that flag - forcing a user that wants to override the registry URL to use the flag on every command.

Appsody `init` should support the flag. When the flag is used, the CLI should store the registry URL in the `stack` element of the `.appsody-config.yaml` file.

chilanti:
Fixes #696


cmd/utils.go
```diff
@@ -532,6 +537,42 @@ func saveProjectNameToConfig(projectName string, config *RootCommandConfig) erro
 	config.Info.log("Your Appsody project name has been set to ", projectName)
 	return nil
 }
+func setStackRegistry(stackRegistry string, config *RootCommandConfig) error {
+	valid, err := ValidateHostNameAndPort(stackRegistry)
```
kylegc: OverrideStackRegistry also calls this, so it probably doesn't need to be here

cmd/utils.go
```diff
@@ -532,6 +537,42 @@ func saveProjectNameToConfig(projectName string, config *RootCommandConfig) erro
...
+	if !valid {
+		return err
+	}
+
+	// update the in-memory project name
```
kylegc: should this comment be removed?

cmd/init.go
```diff
@@ -307,8 +307,15 @@ func install(config *initCommandConfig) error {
 	config.Info.log("Setting up the development environment")
 
 	// reset config.StackRegistry and get it again from the newly untarred .appsody-config.yaml
-	// this will need to be updated when we support --stack-registry on the init command
-	config.StackRegistry = ""
+
+	if config.StackRegistry != "" {
```
kylegc: I'm not convinced this is going to work. What if the stack template's `.appsody-config.yaml` already fully qualifies the stack registry and the user doesn't set `--stack-registry` flag? In this case we shouldn't modify the file, but I'm worried this code will change it.

cmd/init.go
```diff
@@ -307,8 +307,15 @@ func install(config *initCommandConfig) error {
...
+		config.Debug.Log("The will be set to be: ", config.StackRegistry)
```
kylegc: A better debug message: "Setting the stack registry in .appsody-config.yaml to "

chilanti: Done.

cmd/init.go
chilanti: I've done some significant refactoring - please take a look.

cmd/utils.go
```diff
@@ -532,60 +517,115 @@ func saveProjectNameToConfig(projectName string, config *RootCommandConfig) erro
...
+		return "", errors.Errorf("Invalid stack image name detected in project config file: %s", stack)
+	}
+	return "", errors.New("No stack image name detected in project config file")
 
-		config.Debug.Logf("Project stack after override: %s is: %s", config.StackRegistry, projectConfig.Stack)
-		config.ProjectConfig = &projectConfig
+}
+func getProjectConfig(config *RootCommandConfig) (*ProjectConfig, error) {
+
+	projectConfig, err := getProjectConfigFileContents(config)
```
kylegc: Lets put back the `config.ProjectConfig == nil` caching

PR #850, opened 2020-01-14
Fix namespace issue

### LINKED ISSUE #464: Support kubectl semantics of namespace in metadata of deployment manifest
With kubectl, you can define a namespace for your deployment in the deployment manifest or via a command line option to `kubectl apply`. If both these options are used, and the namespace differs, then an error is generated.

With appsody deploy, it appears that although you can define a namespace in the metadata of app-deploy.yaml, unless you also use the` --namespace` cmd line option, this doesn't work. This appears to be because if you do not supply a `--namespace` option, to `appsody deploy`, it will add an explicit `--namespace default `to the eventual `kubectl apply` command. The `appsody deploy delete` actually gives you the smoking gun:

```
$ appsody deploy delete
.
.
Deleting deployment using deployment manifest app-deploy.yaml
Attempting to delete resource from Kubernetes...
Running command: kubectl delete -f app-deploy.yaml --namespace default
[Error] error: the namespace from the provided object "i2" does not match the namespace "default". You must pass '--namespace=i2' to perform this operation.
```

The same problem occurs on the original deploy (although the error message isn't as clear).

I expected, if I chose, to be able to define the namespace in the metadata, and have that used throughout, without any need for command line options

As an aside, I would expect an explicit declaration mismatch of namespaces to cause an error, e.g.:

`appsody deploy --namespace foo`

would fail if the metadata in the app-deploy.yaml specified a different namespace.

EnriqueL8:
Fix #464

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -41,6 +41,7 @@ type buildCommandConfig struct {
 	pullURL             string
 	appDeployFile       string
 	knative             bool
+	namespace           string
 }
 
 //These are the current supported labels for Kubernetes,
@@ -533,6 +534,10 @@ func updateDeploymentConfig(config *buildCommandConfig, imageName string, labels
 	}
 
 	appsodyApplication.Spec.ApplicationImage = imageName
+
+	if appsodyApplication.Namespace == "" {
+		appsodyApplication.Namespace = config.namespace
+	}
 
 	err = writeAppsodyApplication(appsodyApplication, config)
 	if err != nil {
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -535,6 +535,10 @@ func updateDeploymentConfig(config *buildCommandConfig, imageName string, labels
 
 	appsodyApplication.Spec.ApplicationImage = imageName
 
+	// This only applies to the deploy command flow:
+	// - if the namespace doesn't exist in the manifest, and a namespace flag is not set: we write a "default" namespace
+	// - if the namespace does exist in the manifest, and a namespace flag is set: we verify that they are the same. If they are not we throw an error.
+	// - if the namespace doesn't exist in the manifest, and a namespace flag is set: we write the value passed an argument with the flag.
 	if appsodyApplication.Namespace == "" {
 		appsodyApplication.Namespace = config.namespace
 	}
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -73,7 +73,42 @@ The command performs the following steps:
 			namespace := config.namespace
 			configFile := filepath.Join(projectDir, config.appDeployFile)
 
+			exists, err := Exists(configFile)
+			if err != nil && !strings.Contains(err.Error(), "no such file") {
+				return err
+			}
+
+			if !exists && config.nobuild {
+				return errors.Errorf("--no-build flag was used, but deployment manifest %s was not found. Please remove the --no-build, or generate a deployment manifest first by running \"appsody build\"", configFile)
+			}
+
+			if exists {
+				config.Info.Logf("Found deployment manifest %s", configFile)
+
+				appsodyApplication, err := getAppsodyApplication(configFile)
+				if err != nil {
+					return err
+				}
+
+				manifestNamespace := appsodyApplication.Namespace
+				if manifestNamespace != "" {
+					if namespace != "" && manifestNamespace != namespace {
+						return errors.Errorf("the namespace \"%s\" from the deployment manifest does not match the namespace \"%s\" passed as an argument.", manifestNamespace, namespace)
+					}
+
+					namespace = manifestNamespace
+				}
+			}
+
+			if namespace == "" {
+				namespace = "default"
+			}
+
+			config.Info.Logf("Using namespace %s for deployment", namespace)
+
 			if !config.nobuild {
+				config.Info.Log("Building the production image")
+
 				buildConfig := &buildCommandConfig{RootCommandConfig: config.RootCommandConfig}
 				buildConfig.Verbose = config.Verbose
 				buildConfig.pushURL = config.pushURL
@@ -85,6 +120,7 @@ The command performs the following steps:
 				buildConfig.pullURL = config.pullURL
 				buildConfig.knative = config.knative
 				buildConfig.appDeployFile = configFile
+				buildConfig.namespace = namespace
 
 				buildErr := build(buildConfig)
 				if buildErr != nil {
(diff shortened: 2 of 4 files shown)


cmd/deploy.go
```diff
@@ -73,7 +73,42 @@ The command performs the following steps:
 			namespace := config.namespace
 			configFile := filepath.Join(projectDir, config.appDeployFile)
 
+			exists, err := Exists(configFile)
+			if err != nil && !strings.Contains(err.Error(), "no such file") {
```
kylegc: is the string contains "no such file" necessary? the Exists() func already checks for `os.IsNotExist`

functest/deploy_test.go
```diff
@@ -127,19 +127,56 @@ func TestGenerationDeploymentConfig(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	secondNamespace := "secondNamespace"
+	// appsody deploy
+	t.Log("Running appsody deploy...")
+	_, err = cmdtest.RunAppsody(sandbox, "deploy", "--generate-only", "-n", secondNamespace)
+
+	if err != nil && !strings.Contains(err.Error(), "the namespace \""+firstNamespace+"\" from the deployment manifest does not match the namespace \""+secondNamespace+"\" passed as an argument.") {
```
kylegc: You are expecting an error here right? Then it should be `if err == nil || !strings.Contains(...`
However, it would probably be best to split into separate if checks so you can write better messages on `t.Fatal()`

functest/deploy_test.go
```diff
@@ -127,19 +127,56 @@ func TestGenerationDeploymentConfig(t *testing.T) {
...
+		checkDeploymentConfig(t, expectedDeploymentConfig{filepath.Join(sandbox.ProjectDir, deployFile), pullURL, imageTag, namespace, true})
 	}
 }
 
+func TestDeployNamespaceMismatch(t *testing.T) {
+	sandbox, cleanup := cmdtest.TestSetupWithSandbox(t, true)
+	defer cleanup()
+
+	// first add the test repo index
+	_, err := cmdtest.AddLocalRepo(sandbox, "LocalTestRepo", filepath.Join(cmdtest.TestDirPath, "index.yaml"))
```
kylegc: Is this necessary?

cmd/deploy.go
```diff
@@ -153,7 +188,7 @@ The command performs the following steps:
 	deployCmd.PersistentFlags().BoolVar(&config.nobuild, "no-build", false, "Deploys the application without building a new image or modifying the deployment manifest file.")
 	deployCmd.PersistentFlags().StringVarP(&config.appDeployFile, "file", "f", "app-deploy.yaml", "The file name to use for the deployment manifest.")
 	deployCmd.PersistentFlags().BoolVar(&config.force, "force", false, "DEPRECATED - Force the reuse of the deployment manifest file if one exists.")
-	deployCmd.PersistentFlags().StringVarP(&config.namespace, "namespace", "n", "default", "Target namespace in your Kubernetes cluster")
+	deployCmd.PersistentFlags().StringVarP(&config.namespace, "namespace", "n", "", "Target namespace in your Kubernetes cluster")
```
kylegc: I see why you did this, but as a side-effect, the `appsody deploy -h` help text won't specify the default namespace anymore. Not sure if there's an easy way to avoid this though.

cmd/build.go
```diff
@@ -534,6 +535,10 @@ func updateDeploymentConfig(config *buildCommandConfig, imageName string, labels
 
 	appsodyApplication.Spec.ApplicationImage = imageName
 
+	if appsodyApplication.Namespace == "" {
```
kylegc: Should we always overwrite the namespace from the config? I'm not sure of the answer here, but it seems odd to ignore the `config.namespace` if it was set.

EnriqueL8: In the case where a namespace is specified in the config file and a namespace is passed as an argument, we check if they are the same and if not we through an error. This means that it only make sense to write to the config file if the namespace field is empty, if not we will just be writting the same thing.

cmd/deploy.go
EnriqueL8: Yes, I also thought about this - but I can't find another way to know if the flag was actually used...

functest/deploy_test.go
EnriqueL8: No, good spot!

cmd/deploy.go
EnriqueL8: Yup, I was getting an err directly from the Exists when trying to look for the file that is why I added that check.. but I checked and it works without it

functest/deploy_test.go
EnriqueL8: yeah, I have split the ifs and changed to `t.Error`

cmd/deploy.go
kylegc: That's fine. Michele and I have struggled with this in the past too. Lets not worry about it now.

cmd/build.go
kylegc: Ok, I think that makes sense, but it is not clear in the code. Would you please put a code comment here explaining this and its dependency with deploy?

functest/deploy_test.go
```diff
@@ -145,34 +145,32 @@ func TestDeployNamespaceMismatch(t *testing.T) {
...
 	// appsody deploy
-	t.Log("Running appsody deploy...")
+	t.Logf("Running appsody deploy with namespace: %s ...", secondNamespace)
 	_, err = cmdtest.RunAppsody(sandbox, "deploy", "--generate-only", "-n", secondNamespace)
 
-	if err != nil && !strings.Contains(err.Error(), "the namespace \""+firstNamespace+"\" from the deployment manifest does not match the namespace \""+secondNamespace+"\" passed as an argument.") {
-		t.Fatal(err)
+	if err != nil {
+		if !strings.Contains(err.Error(), "the namespace \""+firstNamespace+"\" from the deployment manifest does not match the namespace \""+secondNamespace+"\" passed as an argument.") {
+			t.Errorf("Expecting namespace error to be thrown, but another error was thrown: %s", err)
```
kylegc: I mentioned this in another PR too, but I think it is actually better to check for the string in the command output from RunAppsody rather than the err. This is because it is most important for the message to be printed to the screen for the user to read (a.k.a the output)

cmd/deploy.go
CameronMcWilliam: Missing a full stop at the end of the comment, might be good to include this to keep it in line with the other ones.

Issue #811, opened 2019-12-13
Error message for incorrect push-url parameter indicates wrong cause


nastacio:
**Describe the bug**
Per slack thread (https://appsody.slack.com/archives/CKQ9R1X6V/p1576207357117900) , if you issue

**To Reproduce**
Issue
```
appsody deploy --push-url docker-registry-default.apps.example.com/csantana/demo-12-12:6988 --pull-url docker-registry.default.svc:5000/csantana/demo-12-12:6988
```

gives you the error message below, which indicates a problem with a "-t" or "--tag", where the user has not specified that tag:

```
Running command: docker build -t docker-registry-default.apps.example.com/csantana/demo-12-12:6988/dev.local/demo-12-12 --label dev.appsody.stack.version=0.2.10 --label dev.appsody.stack.tag=appsody/nodejs-express:0.2.10 --label dev.appsody.stack.created=2019-11-22T09:54:22+0000 --label org.opencontainers.image.created=2019-12-12T22:19:16-05:00 --label org.opencontainers.image.title=demo-12-12 --label dev.appsody.stack.configured=docker.io/appsody/nodejs-express:0.2 --label dev.appsody.stack.revision=676a405baed71b4532532322cc245b49291de886 -f /Users/csantana23/.appsody/extract/demo-12-12/Dockerfile /Users/csantana23/.appsody/extract/demo-12-12
[Docker] invalid argument "docker-registry-default.apps.example.com/csantana/demo-12-12:6988/dev.local/demo-12-12" for "-t, --tag" flag: invalid reference format
[Docker] See 'docker build --help'.
[Error] exit status 125
```

**Expected behavior**
The error message should reflect the actual user mistake, which was eventually fixed with this:

```
ok I figured it out need to use --tag
appsody deploy --tag csantana/demo-12-12:24988 --push-url docker-registry-default.apps.example.com --pull-url docker-registry.default.svc:5000
$ oc get appsodyapplication 
NAME         IMAGE                                                        EXPOSED   RECONCILED   AGE
demo-12-12   docker-registry.default.svc:5000/csantana/demo-12-12:24988   true      True         4m
```

**Actual behaviour**
What is the actual behaviour.

**Environment Details (please complete the following information):**

**If applicable please specify:**
 - CLI version:
 - Stack you are using:

**Screenshots**

**Additional context**
Thread in the appsody slack channel:
https://appsody.slack.com/archives/CKQ9R1X6V/p1576207357117900

[...]

PR #851, opened 2020-01-14
Added test for build with no --push flag

CameronMcWilliam:
Related to issue #811 .  When a user tries `appsody deploy` and they specify either a `--push` or `--push-url` flag, and there is no `--tag` associated with `deploy` command,  error message detailing how `--tag` is missing and how to format the `--tag` flag appears.

### DIFF
--- a/functest/build_test.go
+++ b/functest/build_test.go
@@ -316,3 +316,32 @@ func verifyImageAndConfigLabelsMatch(t *testing.T, appsodyApplication v1beta1.Ap
 	}
 
 }
+
+func TestBuildMissingTagFail(t *testing.T) {
+
+	sandbox, cleanup := cmdtest.TestSetupWithSandbox(t, true)
+	defer cleanup()
+
+	// appsody init
+	t.Log("Running appsody init...")
+	_, err := cmdtest.RunAppsody(sandbox, "init", "starter")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// set push flag to true with no tag
+	args := []string{"build", "--push"}
+	_, err = cmdtest.RunAppsody(sandbox, args...)
+	if err != nil {
+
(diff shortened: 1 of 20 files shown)


EnriqueL8:
Duplicate of #843 , but could do with these tests being in the build

functest/deploy_test.go
```diff
@@ -210,3 +210,33 @@ func TestDeployDeleteKubeFail(t *testing.T) {
...
+	}
+
+	// set push flag to true with no tag
+	args := []string{"deploy", "--push"}
+	_, err = cmdtest.RunAppsody(sandbox, args...)
+	if err != nil {
+
+		// As tag is missing, appsody verifies user input and shows error
+		if !strings.Contains(err.Error(), "Tag missing from deploy command") {
+			t.Error("String \"Tag missing from deploy command\" not found in output")
```
EnriqueL8: Can you add the `err` to the output of this message?

functest/build_test.go
```diff
@@ -316,3 +316,32 @@ func verifyImageAndConfigLabelsMatch(t *testing.T, appsodyApplication v1beta1.Ap
...
+	}
+
+	// set push flag to true with no tag
+	args := []string{"build", "--push"}
+	_, err = cmdtest.RunAppsody(sandbox, args...)
+	if err != nil {
+
+		// As tag is missing, appsody verifies user input and shows error
+		if !strings.Contains(err.Error(), "Cannot specify --push or --push-url without a --tag") {
+			t.Errorf("String \"Cannot specify --push or --push-url without a --tag\" not found in output: %v", err)
```
kylegc: Although this string should be present in both, it is probably better to check the output of the RunAppsody command rather than the error. Because we know the output gets printed to the screen for the user to see.

PR #853, opened 2020-01-15
Added Buildah tests for stack package

CameronMcWilliam:
Adding tests for PR #839

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,7 @@ jobs:
     - <<: *test-template
       name: Full Test on Linux
       os: linux
+      # installation instructions for Buildah taken from https://github.com/containers/buildah/blob/master/install.md
       dist: xenial
       before_install:
         - sudo apt-get update -qq
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,11 @@
 language: go
 dist: xenial
+before_install:
+  sudo apt-get update -qq
+  sudo apt-get install -qq -y software-properties-common
+  sudo add-apt-repository -y ppa:projectatomic/ppa
+  sudo apt-get update -qq
+  sudo apt-get -qq -y install buildah
 
 go:
   - 1.12.x
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,11 @@
 language: go
 dist: xenial
 before_install:
-  sudo apt-get update -qq
-  sudo apt-get install -qq -y software-properties-common
-  sudo add-apt-repository -y ppa:projectatomic/ppa
-  sudo apt-get update -qq
-  sudo apt-get -qq -y install buildah
+  - sudo apt-get update -qq
+  - sudo apt-get install -qq -y software-properties-common
+  - sudo add-apt-repository -y ppa:projectatomic/ppa
+  - sudo apt-get update -qq
+  - sudo apt-get -qq -y install buildah
 
 go:
   - 1.12.x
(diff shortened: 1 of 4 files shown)


functest/package_test.go
```diff
@@ -83,8 +83,40 @@ func TestPackageDockerOptions(t *testing.T) {
...
+	// the usual core temp directory
+	sandbox.ProjectDir = filepath.Join(sandbox.ProjectDir, "starter")
+
+	args := []string{"stack", "package", "--buildah"}
+	output, err := cmdtest.RunAppsody(sandbox, args...)
+	if err != nil {
+		t.Fatal(err)
+	} else {
+		if !strings.Contains(output, "[Buildah] Writing manifest to image destination") {
+			t.Errorf("String \"[Buildah] Writing manifest to image destination\" not found in output")
```
EnriqueL8: What exactly are we testing here? That outputs suggests that the command has passed, and the image is built?

functest/package_test.go
```diff
@@ -83,8 +83,40 @@ func TestPackageDockerOptions(t *testing.T) {
...
+	} else {
+		t.Logf("Copied %s to %s", stackDir, sandbox.ProjectDir)
+	}
+
+	// Because the 'starter' folder has been copied, the stack.yaml file will be in the 'starter'
+	// folder within the temp directory that has been generated for sandboxing purposes, rather than
+	// the usual core temp directory
+	sandbox.ProjectDir = filepath.Join(sandbox.ProjectDir, "starter")
+
+	args := []string{"stack", "package", "--buildah"}
```
EnriqueL8: Do we need a `--format=docker` here?

functest/package_test.go
```diff
@@ -106,9 +138,16 @@ func TestPackageBuildahOptions(t *testing.T) {
 	sandbox.ProjectDir = filepath.Join(sandbox.ProjectDir, "starter")
 
 	args := []string{"stack", "package", "--buildah", "--buildah-options", "--format=docker"}
-	_, err = cmdtest.RunAppsody(sandbox, args...)
+	output, err := cmdtest.RunAppsody(sandbox, args...)
 	if err != nil {
 		t.Fatal(err)
+	} else {
+		if !strings.Contains(output, "format=docker") {
```
EnriqueL8: why not include the dashes in the Contains?

cmd/stack_package.go
```diff
@@ -272,6 +264,14 @@ The packaging process builds the stack image, generates the "tar.gz" archive fil
 				cmdArgs = append(cmdArgs, options...)
 			}
 
+			labelPairs := CreateLabelPairs(labels)
```
EnriqueL8: Could this change also be an issue in `build.go` when using build options?

.travis.yml
```diff
@@ -33,6 +33,13 @@ jobs:
     - <<: *test-template
       name: Full Test on Linux
       os: linux
+      dist: xenial
+      before_install:
```
EnriqueL8: Could you put this link: https://github.com/containers/buildah/blob/master/install.md a line above to show where we have taken these commands from?

functest/package_test.go
CameronMcWilliam: There is a test with build Options and one without build options passed

functest/package_test.go
CameronMcWilliam: Don't think the dashes are passed through in the Buildah command

cmd/stack_package.go
CameronMcWilliam: Think in `build.go` the `buildOptions` are already before labels are added

Issue #841, opened 2020-01-09
CopyDir results vary between Windows and non-Windows


tnixa:
**Describe the bug**
Here's what I'm seeing with the CopyDir function in utils.go when running stack package
**Mac:**
fromDir: `/Users/tnixa/go/src/github.com/appsody/stacks/incubator/nodejs-express`
toDir: `/Users/tnixa/.appsody/stacks/packaging`
When I run `stack package` on mac I end up with  `/Users/tnixa/.appsody/stacks/packaging/nodejs-expres/<contents>`
*it is adding an additional nodejs-express directory which is not expected/desired

**Windows:**
fromDir: `C:\Users\TerrenceNixa\repos\stacks\incubator\nodejs-express`
toDir: `C:\Users\TerrenceNixa\.appsody\stacks\packaging`
When I run `stack package` on windows I end up with `C:\Users\TerrenceNixa\.appsody\stacks\packaging\<contents>`

**Environment Details (please complete the following information):**
 - OS: [e.g. iOS] mac and windows 10

**If applicable please specify:**
 - CLI version: 0.5.4
 - Stack you are using: nodejs-express

**Additional context**
Thinking to append `/*` to the non-windows fromDir in CopyDir to fix this issue but there are workaround for this problem already implemented so we will have to fix/test those workarounds also.

kylegc:
Related: https://github.com/appsody/appsody/issues/838

Note from that issue that appending `/*` isn't good enough, it won't copy hidden files that start with `.`

Once this is fixed, I think the buildah copy code highlighted in #838 should be modified to use the CopyDir func instead of calling `cp` itself.

PR #855, opened 2020-01-16
Issue841

tnixa:
This PR addresses #841 
Changing CopyDir to restrict target directory to only be the name and not the actual directory. This will cause the need for more unit tests and refactoring of files that already use CopyDir.

### DIFF
--- a/cmd/stack_package_test.go
+++ b/cmd/stack_package_test.go
@@ -232,8 +232,7 @@ func TestPackageNoStackYamlFail(t *testing.T) {
 	log.InitLogging(&outBuffer, &outBuffer)
 
 	stackDir := filepath.Join(cmdtest.TestDirPath, "starter")
-	var err error
-	err = cmd.CopyDir(log, stackDir, (filepath.Join(sandbox.ProjectDir, "starter")))
+	err := cmd.CopyDir(log, stackDir, (filepath.Join(sandbox.ProjectDir, "starter")))
 	if err != nil {
 		t.Errorf("Problem copying %s to %s: %v", stackDir, sandbox.ProjectDir, err)
 	} else {
@@ -291,8 +290,7 @@ func TestPackageInvalidStackYamlFail(t *testing.T) {
 	log.InitLogging(&outBuffer, &outBuffer)
 
 	stackDir := filepath.Join(cmdtest.TestDirPath, "starter")
-	var err error
-	err = cmd.CopyDir(log, stackDir, (filepath.Join(sandbox.ProjectDir, "starter")))
+	err := cmd.CopyDir(log, stackDir, (filepath.Join(sandbox.ProjectDir, "starter")))
 	if err != nil {
 		t.Errorf("Problem copying %s to %s: %v", stackDir, sandbox.ProjectDir, err)
 	} else {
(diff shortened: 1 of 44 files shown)


cmd/stack_package_test.go
```diff
@@ -375,16 +365,11 @@ func TestPackageNoTemplates(t *testing.T) {
...
-	var err error
-	if runtime.GOOS == "windows" {
-		err = cmd.CopyDir(log, stackDir, (filepath.Join(sandbox.ProjectDir, "starter")))
-	} else {
-		err = cmd.CopyDir(log, stackDir, sandbox.ProjectDir)
-	}
+	err := cmd.CopyDir(log, stackDir, (filepath.Join(sandbox.ProjectDir, "starter")))
 	if err != nil {
-		t.Errorf("Problem copying %s to %s: %v", stackDir, sandbox.ProjectDir, err)
+		t.Errorf("Problem copying %s to %s: %v", stackDir, (filepath.Join(sandbox.ProjectDir, "starter")), err)
```
EnriqueL8: Just a suggestion: we could set a variable for the `filepath.Join(...)` and then use is where needed. This could be applied to all these similar cases.

cmd/utils_test.go
```diff
@@ -569,14 +571,16 @@ func TestMoveFailPermissions(t *testing.T) {
 	if err != nil {
 		t.Errorf("Error creating the file: %v", err)
 	}
+	println("existing file:", existingFile)
```
EnriqueL8: I think ww could use `t.log` here

cmd/utils_test.go
```diff
@@ -587,40 +591,278 @@ func TestMoveFailPermissions(t *testing.T) {
...
+		t.Fatal("Error writing file3: ", err)
+	}
+	t.Log("Created file3: ", file3)
+
+	// Create test file4 <.extension>
+	file4 := filepath.Join(sourceSubDir, ".file4")
+	data = []byte("home: " + sourceSubDir + "\n" + "generated-by-tests: Yes" + "\n")
+	err = ioutil.WriteFile(file4, data, 0644)
+	if err != nil {
+		t.Fatal("Error writing file4: ", err)
```
EnriqueL8: It would be interesting to convert these into table test, but not very important

cmd/utils.go
```diff
@@ -661,10 +662,21 @@ func MoveDir(log *LoggingConfig, fromDir string, toDir string) error {
...
-		return err
+		log.Error.logf("Source %s does not exist.", fromDir)
+		//return err
+		return errors.Errorf("Source %s does not exist.", fromDir)
+	}
+
+	// fail if toDir exists on the file system
+	// toDir should just be the name of the target directory, not an existing file or directory
+	// if toDir is an existing file or directory it causes inconsistent results between windows and non-windows
+	_, err = os.Stat(toDir)
```
kylegc: For this `os.Stat` and the one above, we should instead use `Exists()` func from utils.go. It returns a bool and err. The bool indicates if the file/dir exists or not, and the err is != nil if there was a problem checking the file/dir.

tnixa: changes incorporated

Issue #618, opened 2019-11-12
Stack create should be able to copy custom stacks


neeraj-laad:
**Is your feature request related to a problem? Please describe.**
Today we can copy a stack to use that as a basis for a new stack using the command `appsody stack create <name> --copy <base-stack>`

This has a restriction though that the base stack has to be a public appsody stack. It would be great if it could use the user's local appsody configuration and use those repos to be able to create a copy in some way.

**Describe the solution you'd like**
appsody stack create should be able to copy non-public/unofficial Appsody stacks.

henrynash:
This would be particularly useful within private Kabanero collections - but also if we want enterprises who have bought into Appsody but not Kabanero, and are building their own private enterprise-wide Appsody Hubs/Repos, then they will absolutely need this.

timroster:
Scratching my head as to why:
`appsody stack create custom-nodejs-express --copy kabanero/nodejs-express`
does not work and found this issue. Agree with @henrynash that an enterprise solution architect that is looking to extend a kabanero collection, even the public one, is going to get stuck on this limitation.

kylegc:
I agree this is a problem that needs to be addressed. Unfortunately there are architectural challenges with the implementation of this that need to get sorted out first.

neeraj-laad:
So today for a custom stack, we do know which git repo has the source for the stack to be copied. Also, there is no requirement that the stack has to be on a git repo to be in an index.

Could we include the git url for each stack in the index.yaml and use that for copying the stack over?

Initial thoughts:
`appsody stack add-to-repo` - can add path/git url for the folder that contains the stack. When running locally it sill just put the file:///url and when running as part of our CI it can add proper url.

`appsody stack create` can then lookup for this information in the user's local configuration and then use that to copy over.

@kylegc @groeges would like to get your thoughts.

neeraj-laad:
Thinking a bit more, there were issues with when the src for a custom stack will get to a git repo with respect to the packaging of the stack. My proposal above seems to add a  dependency which might or might not be true. Moreover, stack creators might prefer to use a diff SCM system, so having that flexibility will be good.

Here's another proposal:

`appsody stack package` - should package the src folder for the stack (the folder it is packaging) and create an archive (exactly what it does for templates today). The URL for accessing this archive is also added to `src: <url>` field in the generated index.

This src gets published as part of stack release. We can keep all the handling of release-url to be same as that is for the templates.

[...]

PR #817, opened 2019-12-17
Tests can now run concurrently, with up to 10 threads

upLukeWeston:
The `testdata` folder is now being copied to `sandbox.TestDataPath` so the tests don't step on one another with things such as `Dockerfile` lint tests.

Some of the functional tests had issues running alongside each other because of ports being used and containers clashing - those have had asynch set to false.

In response to #812.

PR #856, opened 2020-01-17
Stack create behaviour change

CameronMcWilliam:
Issue: #618  Unsure about this implementation, am welcome to better ideas.  Ready for a first pass though.


CameronMcWilliam:
Still has the old method in it as a fall back, for stacks that don't have the `src:` field.  This will be removed next release.

cmd/stack_create.go
```diff
@@ -83,54 +87,93 @@ The stack name must start with a lowercase letter, and can contain only lowercas
...
 
-			if !valid {
-				return errors.Errorf("Invalid stack name: " + config.copy + ". Stack name must be in the format <repo>/<stack>")
+			// get specificed repo and umarshal
+			repoInfo := repoFile.GetRepo(repoID)
+
+			// error if repo not found in repository.yaml
+			if repoInfo == nil {
+				//return errors.Errorf("Repository: %s not found in repository.yaml file", repoID)
+				return oldCreateMethod(rootConfig.LoggingConfig, rootConfig, config, stack, currentTime)
```
CameronMcWilliam: This worked fine on my local branch, but on travis the test failed unless I returned the `oldCreateMethod` not sure if this needs to be reworked.

cmd/stack_create.go
```diff
@@ -144,7 +179,145 @@ The stack name must start with a lowercase letter, and can contain only lowercas
...
+func getStack(r *IndexYaml, name string) *IndexYamlStack {
+	for _, rf := range r.Stacks {
+		if rf.ID == name {
+			return &rf
+		}
+	}
+	return nil
+}
+
+// taken from https://medium.com/@skdomino/taring-untaring-files-in-go-6b07cf56bc07
```
EnriqueL8: We could put this in utils

cmd/stack_create.go
```diff
@@ -83,54 +87,85 @@ The stack name must start with a lowercase letter, and can contain only lowercas
...
-					return err
-				}
+			untarErr := untarSource(rootConfig.LoggingConfig, stack, extractFile, config.Dryrun)
+			if untarErr != nil {
+				return untarErr
 			}
 
+			//deleting the stacks repo zip
+			os.Remove(extractDirFile)
+
```
EnriqueL8: Just needs more commenting before I can get my head around it

cmd/stack_create.go
CameronMcWilliam: Fixed now

cmd/utils.go
```diff
@@ -2186,3 +2185,73 @@ func lintMountPathForSingleFile(path string, log *LoggingConfig) {
...
+					return err
+				}
+
+				// manually close here after each file operation; defering would cause each file close
+				// to wait until all operations have completed.
+				f.Close()
+			}
+		}
+	} else {
+		log.Info.logf("Dry Run skipping -Untar of filee: %s to destination %s", r, dst)
```
upLukeWeston: Unnecessary "e" at the end of "file".

cmd/stack_addtorepo.go
```diff
@@ -197,6 +197,14 @@ The updated repository index file is created in  ~/.appsody/stacks/dev.local dir
 			// build up stack struct for the new stack
 			newStackStruct := initialiseStackData(stackID, stackYaml)
 
+			versionArchiveTar := stackID + ".v" + stackYaml.Version + ".source.tar.gz"
```
EnriqueL8: @upLukeWeston Suggests that in Windows tar don't exists, he suggests that we need to consider the case for zip files?

cmd/init.go
```diff
@@ -349,7 +349,7 @@ func install(config *initCommandConfig) error {
 	return nil
 }
 
-func untar(log *LoggingConfig, file string, noTemplate bool, overwrite bool, dryrun bool) error {
+func initUntar(log *LoggingConfig, file string, noTemplate bool, overwrite bool, dryrun bool) error {
```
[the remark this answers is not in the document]
CameronMcWilliam: Yes, and the one I am creating is a more generic version of an untar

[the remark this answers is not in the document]
EnriqueL8: Okay, I have compared both implementations and the only change is the template arg - I am wondering if we could merge both functions? What do you think?

cmd/stack_addtorepo.go
kylegc: I don't think this is an issue because we handle tar files on Windows today with the template.tar files.

CameronMcWilliam:
> Sorry Cameron, there's too much here that I'm not comfortable merging right before a release. I don't see any glaring issues, but this is not simply running the old code until we add a `src` property. There's a lot of new code that will always be run, and the `src` property will be added automatically on new stack releases which has the potential of breaking things.
> 
> Before we merge this, more testing needs to be done. In particular, I think @groeges needs to review this code and test it out with the stack release processes. I'll add him as a reviewer.

Makes sense.  Better to take longer to make sure this all works.  That allows me to add tests to this PR too.

cmd/utils.go
```diff
@@ -2049,8 +2049,7 @@ func downloadFileToDisk(log *LoggingConfig, url string, destFile string, dryrun
 }
 
 // tar and zip a directory into .tar.gz
-func Targz(log *LoggingConfig, source, target string) error {
-	filename := filepath.Base(source)
+func Targz(log *LoggingConfig, source, target, filename string) error {
 	log.Info.log("source is: ", source)
 	log.Info.log("filename is: ", filename)
 	log.Info.log("target is: ", target)
```
groeges: Do we need the above lines to be logged to the console as part of the command. Shouldn't this be debug info?

CameronMcWilliam: Yes, as long as we leave the printed path of where it has put the archived files, that should be fine, I agree with all the lines It can look very cluttered.

kylegc:
Thanks for the update Steve. We discussed this on the side and decided to hold this out of today's release. There is a scenario that would break stack creators because the current CI scripts do not upload the new source.tar file:
- Release a new stack. This PR will add the `src` field to the index.yaml but not upload the source.tar it points to.
- Then running `appsody stack create` on that stack will cause an error because the `src` location doesn't exist.

The stack CI scripts should be updated before this is merged.

EnriqueL8:
The new test changes after merging them with this branch makes the lint fail

upLukeWeston:
Yeah, `cmdtest.TestDirPath` was removed and replaced with `sandbox.TestDataPath` in #817 - that'll fix the lint issue.

PR #864, opened 2020-01-23
Set pull policy to always

EnriqueL8:
Fixes #864 - It does not fix the reload of the image at every `appsody deploy`, this will fix when doing `appsody deploy delete` and `appsody deploy` with the same image.

I'm not a big fan of importing the `corev1` package from Kubernetes API

TODO: Needs testing on Knative

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -481,6 +481,7 @@
     "github.com/spf13/cobra/doc",
     "github.com/spf13/viper",
     "gopkg.in/yaml.v2",
+    "k8s.io/api/core/v1",
     "k8s.io/klog",
     "sigs.k8s.io/yaml",
   ]
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -25,6 +25,7 @@ import (
 	"github.com/appsody/appsody-operator/pkg/apis/appsody/v1beta1"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/yaml"
 
 	"bytes"
@@ -544,6 +545,11 @@ func updateDeploymentConfig(config *buildCommandConfig, imageName string, labels
 	}
 
 	appsodyApplication.Spec.ApplicationImage = imageName
+
+	if appsodyApplication.Spec.PullPolicy == nil {
+		pullPolicy := corev1.PullAlways
+		appsodyApplication.Spec.PullPolicy = &pullPolicy
+	}
 
 	// This only applies to the deploy command flow:
 	// - if the namespace doesn't exist in the manifest, and a namespace flag is not set: we write a "default" namespace
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -44,6 +44,7 @@ type buildCommandConfig struct {
 	knative             bool
 	knativeFlagPresent  bool
 	namespace           string
+	pullPolicy          string
 }
 
 //These are the current supported labels for Kubernetes,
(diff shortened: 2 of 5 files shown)


EnriqueL8:
- Add flag to build for PullPolicy: `--pullPolicy=Always/IfNotPresent/Never` No pull policy is default
- Set as always from the deploy command - no flag on deploy
- Add automatic deploy delete to deploy command

EnriqueL8:
Might need a bit more thought on how we check that the application is deployed.
`kubectl get appsodyapplication.appsody.dev/<projectname>`

EnriqueL8:
I think the tests are failing because the name of the docker image built is too large (> 255 characters)

rykal-nate:
@EnriqueL8  @kylegc  - I was doing some investigations on this topic and I think we are going to have an issue with this approach using docker-desktop with kube.  Changing the pull policy to ALWAYS creates a condition such that kube is no longer able to find the local docker image for some reason? 

**
  applicationImage: dev.local/java-openliberty
  pullPolicy: Always
** 

POD Failure Messages

Events:
  Type     Reason     Age                  From                     Message
  ----     ------     ----                 ----                     -------
  Normal   Scheduled  5m16s                default-scheduler        Successfully assigned default/java-openliberty-6f67f5d97c-l9jxt to docker-desktop
  Normal   Pulling    2m (x4 over 5m15s)   kubelet, docker-desktop  Pulling image "dev.local/java-openliberty"
  Warning  Failed     85s (x4 over 4m40s)  kubelet, docker-desktop  Failed to pull image "dev.local/java-openliberty": rpc error: code = Unknown desc = Error response from daemon: Get https://dev.local/v2/: Service Unavailable
  Warning  Failed     85s (x4 over 4m40s)  kubelet, docker-desktop  Error: ErrImagePull
  Warning  Failed     73s (x6 over 4m40s)  kubelet, docker-desktop  Error: ImagePullBackOff
  Normal   BackOff    8s (x11 over 4m40s)  kubelet, docker-desktop  Back-off pulling image "dev.local/java-openliberty"

**
  applicationImage: java-openliberty
  pullPolicy: Always
** 

POD Failure Messages

Events:
  Type     Reason     Age                From                     Message
  ----     ------     ----               ----                     -------
  Normal   Scheduled  40s                default-scheduler        Successfully assigned default/java-openliberty-77c74b4c75-pzwft to docker-desktop
  Normal   Pulling    25s (x2 over 39s)  kubelet, docker-desktop  Pulling image "java-openliberty"
  Warning  Failed     24s (x2 over 37s)  kubelet, docker-desktop  Failed to pull image "java-openliberty": rpc error: code = Unknown desc = Error response from daemon: pull access denied for java-openliberty, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
  Warning  Failed     24s (x2 over 37s)  kubelet, docker-desktop  Error: ErrImagePull
  Normal   BackOff    11s (x2 over 36s)  kubelet, docker-desktop  Back-off pulling image "java-openliberty"
  Warning  Failed     11s (x2 over 36s)  kubelet, docker-desktop  Error: ImagePullBackOff


For what its worth I tested "kubectl rollout restart deployment <deployment-name>" and everything worked. However, we need this requires kube 1.15 and needs more testing.... just another approach to be aware of. 

1) Without changing app-deploy.yaml  (image pull policy not changed)
2) Make application change
3) appsody build 
4) ("appsody deploy mock") kubectl rollout restart deployment java-openliberty
5) Application updated - new image pulled.

EnriqueL8:
That is interesting! Thanks for trying this out!

I have been testing it with minikube locally and on Openshift remotely. I also encountered issues on linking the Kubernetes node from Docker for desktop to my local Docker registry. I believe this change could have been introduced in a newer version of Docker for desktop but not sure... 


I also like the `rollout restart` option!. Just as a suggestion, we could add `appsody deploy restart` command, and call that instead of delete then kubectl apply. I agree this will need more testing because there could be other things in the Spec that the user could change and that the `rollout restart` doesn't pick up?

kylegc:
Docker Desktop's Kubernetes uses the local docker daemon and thus shares a local image cache with the docker command. So it seems that a pull policy of "Always" breaks because it ignores the local cache and tries to remotely pull the image, but the image isn't available remotely.

So it seems that we need pull policy of "Always" everywhere except Docker Desktop's k8s :-(

Does the new rollout restart "always" pull images regardless of the pull policy? If so, I like this option too, however it may be difficult to say everyone must use kube 1.15. OpenShift doesn't get this feature until 4.3. 

We may have to go back to the digest idea or a unique tag. Bummer I know.

EnriqueL8:
Closing this PR fow now - I think this needs more discussion and design.
As a workaround, users will be have to pass a new `tag` everytime they want to deploy. Maybe users will not run `appsody deploy` very frequently, and thus this shouldn't be too much of a problem. We could also think about the digest idea mentioned above.

Issue #868, opened 2020-01-24
Add image information to Index.yaml


kilnerm:
**Is your feature request related to a problem? Please describe.**

Presently the only way to see what image a particular stack in my index will use is to look in the `.appsody-config.yaml` from either creating an application for the stack or extracting the template zip and relies on an understanding of how the stacks are put together.

Further, if we implement the function as described in #867 then having the image tag clearly accessible will help developers understand what patching behaviour they would get (we could even enhance `appsody list` to display this).

**Describe the solution you'd like**

I would like the image tag meta-data added to the `.apsody-config.yaml` to be duplicated in the generated index files for ease of reference.

**Describe alternatives you've considered**

Extracting the templates zip or creating an application to extract the information. Both these actions are processing-intensive and will not scale well for large volumes of stacks.

**Additional context**
This will be very useful for the current Kabanero implementation.

neeraj-laad:
I am wondering if the index should just list out the full image tag and digest information that is built as part of appsody stack package. 

stack:
  image: 
    tag: reg/name/version
    digest: <image digest>

This way the index becones a single place for understanding what was built (src), where it was published (docker image, src url, templates url).

Kamran64:
I have the fully qualified image tag in the index now but I'm struggling to get the digest. I think an image digest is only generated if you push or pull to/from a registry hence why I'm seeing `<none>` for my digest values after building locally. See: https://github.com/moby/moby/issues/32016 and [here](https://stackoverflow.com/questions/44087012/sha256-of-locally-built-docker-image).

@kilnerm - Would it be okay to only have the image tag?

kilnerm:
@Kamran64 Yes, looking at the digest situation I don't think it is reasonable to add that in the context of our current tooling.

Issue #867, opened 2020-01-24
Enable stack creators to opt out of automatic patching


kilnerm:
**Is your feature request related to a problem? Please describe.**

Currently Appsody defaults to using MAJOR.MINOR tags when generating stack config files. This means that any stack built using the Appsody CLI will automatically adopt any patch level updates made to the stack which could be undesirable to some users.

**Describe the solution you'd like**

I would like to see an additional flag added to the stack package command (and stack publish maybe) that would allow a user to override the default behavior and allow them to use MAJOR.MINOR.PATCH or the docker image digest in the image meta-data tag.

**Describe alternatives you've considered**

**Additional context**

PR #869, opened 2020-01-30
Add image tag to index during appsody stack package

Kamran64:
Fixes: https://github.com/appsody/appsody/issues/868

- Adds the fully qualified image tag to the index generated by appsody stack package.
- Changes the behaviour of appsody stack add-to-repo to check if appsody stack package has been run before add-to-repo.
- For add-to-repo the image value is copied from the dev.local-index.yaml into the add-to-repo index.


cmd/stack_package.go
```diff
@@ -251,7 +253,7 @@ The packaging process builds the stack image, generates the "tar.gz" archive fil
 				return errors.Errorf("Error applying templating: %v", err)
 			}
 
-			// tag with the full version then mojorminor, major, and latest
+			// tag with the full version then majorminor, major, and latest
```
EnriqueL8: We will need to tag the image with `MAJOR.MINOR.PATCH`, in order for the metadata to be accurate. 
We could add the flag discussed in #867 as part of this PR?
Another suggestion could be to add a `semver[majorminorpatch]` the same way we have `semver[majorminor]`

Kamran64: Thanks for taking a look - as we discussed the major.minor.patch level is tagged on line 256 (`buildImage` variable).

PR #870, opened 2020-01-31
Rewritten 'stack_create' and 'stack_lint' tests to be in tables

### LINKED ISSUE #854: Restructure tests in table format
A lot of the tests in both `stack_create_test.go` and `stack_lint_test.go` repeat the same actions with different parameters. This had made the files unnecessarily large, and less clear to read.

The tests could be altered to be in a table format, where the different parameters are passed through to the test, making it easier to read, and easier to add additional tests to in the future without writing a whole new function (if the table already includes execution of the desired `appsody` command).

We could leave them as is

upLukeWeston:
Resolves #854.
Makes the test files a lot less cumbersome, making them easier to read, edit, and add future tests to.


cmd/stack_create_test.go
```diff
@@ -23,291 +22,129 @@ import (
...
+			}
+			exists, err := cmdtest.Exists(testStackName)
+			if err != nil {
+				t.Fatal("Failed to check if the stack exists: ", err)
+			}
+			if !exists {
+				t.Fatal("Stack doesn't exist despite appsody stack create executing correctly")
+			}
+			os.RemoveAll(testStackName)
+			if err != nil {
```
EnriqueL8: This `err` is from the Exists command, does the `RemoveAll` return an error?

cmd/stack_lint_test.go
```diff
@@ -25,669 +23,108 @@ import (
...
-
+func TestLinterInvalidValues(t *testing.T) {
+	var linterInvalidValues = []struct {
+		testName     string
+		targetPath   string
+		containsLine string
+		replaceLine  string
+		expectedLog  string
+	}{
+		{"Invalid Run Value", filepath.Join("image", "Dockerfile-stack"), "APPSODY_RUN", "Testing", "Missing APPSODY_RUN"},
```
EnriqueL8: we could store `filepath.Join` in a var?

cmd/stack_create_test.go
EnriqueL8: Actually do we even have to remove? Since we are sandboxing each test?

cmd/stack_create_test.go
```diff
@@ -23,291 +22,129 @@ import (
...
+			if !strings.Contains(output, tt.expectedLogs) {
+				t.Errorf("Did not find expected error '%s' in output", tt.expectedLogs)
+			}
+			exists, err := cmdtest.Exists(tt.stackName)
+			if err != nil {
+				t.Fatal("Error attempting to check stack exists: ", err)
+			}
+			if exists {
+				// It SHOULDN'T exist, but it might
+				err = os.RemoveAll(tt.stackName)
```
EnriqueL8: Do we even have to remove? This is in sandbox

cmd/stack_create_test.go
upLukeWeston: Good question! I guess not, actually? I can give it a go with this bit removed. Didn't really think about that when I took the functionality from each test to turn it into tables.

cmd/stack_lint_test.go
upLukeWeston: As in use a var rather than `filepath.Join("image", "Dockerfile-stack")` every time? Yeah that'd work, probably should have noticed that myself

cmd/stack_create_test.go
EnriqueL8: The exists check should still be there

cmd/stack_create_test.go
upLukeWeston: It left a bunch of files in the `testdata` folder, which then failed the tests the second time running. I can look into why another time?

cmd/stack_create_test.go
upLukeWeston: It left a bunch of files in the testdata folder, which then failed the tests the second time running. I can look into why another time?

cmd/stack_create_test.go
EnriqueL8: Sounds good, but then would need to do `err = os.RemoveAll(testStackName)`

upLukeWeston: Yes, of course!

Makefile
```diff
@@ -3,7 +3,7 @@
 #### Constant variables
 export STACKSLIST ?= incubator/nodejs
 # use -count=1 to disable cache and -p=1 to stream output live
-GO_TEST_COMMAND := go test -v -count=1 -p=5 -parallel 5 -covermode=count -coverprofile=cover.out -coverpkg ./cmd -timeout 15m
+GO_TEST_COMMAND := go test -v -count=1 -p=2 -parallel 5 -covermode=count -coverprofile=cover.out -coverpkg ./cmd
```
EnriqueL8: We will need to look more into this after merging

functest/debug_test.go
```diff
@@ -61,7 +61,7 @@ func TestDebugSimple(t *testing.T) {
 		runChannel := make(chan error)
 		containerName := "testDebugSimpleContainer" + strings.ReplaceAll(stackRaw[i], "/", "_")
 		go func() {
-			_, err := cmdtest.RunAppsody(sandbox, "debug", "--name", containerName)
+			_, err := cmdtest.RunAppsody(sandbox, "debug", "--name", containerName, "-P")
```
EnriqueL8: We could look further into which tests actually use ports... We don't have tests which check that the ports specified in the Dockefile-stack are actually getting used...

Makefile
upLukeWeston: I tested this change on master branch and compared how long the testing run takes to complete with `-p=5` and `-p=2` and there was no clear difference between the two. Both took a similar enough amount of time that I don't think it made any (at least _substantial_) difference in reducing it; but yeah, I'd prefer to investigate it officially a bit more.

functest/debug_test.go
upLukeWeston: Yeah this could definitely be something to look into as a different PR, there is an underlying issue that still intermittently fails which could likely be caused by misallocation of ports.

PR #873, opened 2020-01-31
Don't override registry name unless stack-registry flag is used

neeraj-laad:
Appsody should not add default stack registry unless stack-registry flag is used.

Signed-off-by: Neeraj Laad <neeraj.laad@gmail.com>

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -311,7 +311,7 @@ func install(config *initCommandConfig) error {
 
 	// reset config.StackRegistry and get it again from the newly untarred .appsody-config.yaml
 
-	defaultStackRegistry := getDefaultStackRegistry(config.RootCommandConfig)
+	//defaultStackRegistry := getDefaultStackRegistry(config.RootCommandConfig)
 	configFileStackRegistry, err := getStackRegistryFromConfigFile(config.RootCommandConfig)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


cmd/init.go
```diff
@@ -311,7 +311,7 @@ func install(config *initCommandConfig) error {
 
 	// reset config.StackRegistry and get it again from the newly untarred .appsody-config.yaml
 
-	defaultStackRegistry := getDefaultStackRegistry(config.RootCommandConfig)
+	//defaultStackRegistry := getDefaultStackRegistry(config.RootCommandConfig)
```
chilanti: This change would break the scenario in which Jane changes the `~/.appsody/.appsody.yaml` file and sets `images` to something other than `docker.io` (which is the default). If she sets it to `myregistry.io`, `appsody init` should use `myregistry.io/<namespace>/<image name>` (unless the stack fully qualifies the image name).

cmd/init.go
```diff
@@ -321,13 +321,12 @@ func install(config *initCommandConfig) error {
 	if config.StackRegistryInit != "" {
 		config.Debug.Log("The flag --stack-registry was set to: ", config.StackRegistryInit)
 		stackRegistry = config.StackRegistryInit
-	} else if configFileStackRegistry == "" {
```
chilanti: I would suggest changing this line to
```
else if configFileStackRegistry == "" && defaultRegistry != "docker.io" {
```

cmd/init.go
groeges: if she wants that wouldn't she use `appsody init incubator/noidejs --image-registry myregistry.io`

neeraj-laad: Wouldn't we always use the registry from `~/.appsody/.appsody.yaml` if `appsody_config.yaml` did not have registry specified?

cmd/init.go
neeraj-laad: I see this will update the project config with <custom-registry>/<stack>:<version> . but if this was not done and user's config stayed at <stack>:<version> then appsody will use the <custom-registry> value to pull the image anyways as that is what our default behaviour was.. am I missing something?

kylegc:
Hey Neeraj, this is very close! Just a little bit of code cleanup and two things to add. I was going try to explain them as code review comments here, but since we are in a time crunch on this, I decided to submit the code in a new PR https://github.com/appsody/appsody/pull/874

Feel free to take make those changes here and close 874, or close this one and go with 874. I'm fine either way :-)

cmd/init.go
kylegc: This is actually okay because `~/.appsody/.appsody.yaml` is read later on and performs the overwrite in `getProjectConfig()`

neeraj-laad:
Thanks Kyle! Will close this one..

PR #875, opened 2020-02-03
Add wait for docker stop backround routine to finish

kylegc:
Fixes https://github.com/appsody/controller/issues/66


cmd/dev_common.go
```diff
@@ -231,11 +232,16 @@ func commonCmd(config *devCommonConfig, mode string) error {
...
 	volumeMaps = append(volumeMaps, "-v", controllerVolumeMount)
+
+	var wg sync.WaitGroup
 	if !config.Buildah {
 		c := make(chan os.Signal, 1)
 		signal.Notify(c, os.Interrupt, syscall.SIGTERM, syscall.SIGINT)
 		go func() {
 			<-c
+			config.Debug.Log("Inside signal handler for appsody command")
+			wg.Add(1)
```
chilanti: So - if `Add` is inside the goroutine, there's a chance that the main thread does wg.Wait() before Add is run (and it would end in that case).

kylegc: Yes I thought of this too, BUT we can't put the `wg.Add(1)` outside the signal handler because in some cases the signal handler doesn't get run and the CLI would hang forever. Such as running `appsody stop` or when the `docker run` terminates early. 

We just have to count on the signal handler's `Add` getting called before the main thread's `Wait`, which from testing so far has always happened.

chilanti: Let's go with that - I don't think there's a better pattern to ensure that we won't wait forever (other than a timeout, which we don't want for other reasons).

PR #877, opened 2020-02-04
TESTING: Ps.go

CameronMcWilliam:
Added 2 small tests to increase code coverage of `ps.go`, brings code coverage of `ps.go` to 80+% (remaining bits seem unreachable)


functest/ps_test.go
```diff
@@ -106,3 +106,37 @@ func TestPS(t *testing.T) {
...
+		}
+	}
+}
+
+func TestPsArgumentFail(t *testing.T) {
+
+	sandbox, cleanup := cmdtest.TestSetupWithSandbox(t, false)
+	defer cleanup()
+
+	// appsody ps with no running containers
```
Kamran64: nit: should the comment here be different?

upLukeWeston:
You _should_ be able to change the `false` in both
```
sandbox, cleanup := cmdtest.TestSetupWithSandbox(t, false)
```
to be `true`, because I don't _think_ there is anything stopping these two tests from running concurrently with others. Apart from that, looks good to me.

PR #878, opened 2020-02-04
Do not write default namespace to app-deploy

EnriqueL8:


cmd/build.go
```diff
@@ -549,7 +550,7 @@ func updateDeploymentConfig(config *buildCommandConfig, imageName string, labels
 	// - if the namespace doesn't exist in the manifest, and a namespace flag is not set: we write a "default" namespace
```
kylegc: This comment needs to be updated

PR #884, opened 2020-02-06
HOLD: Removed old create method.

CameronMcWilliam:
In reference to PR: #856 and Issue: #618 .  This now removes the old method of unzipping the github stacks folder.  SHOULD NOT BE MERGED UNTIL ALL STACKS HAVE BEEN RELEASED WITH A SRC URL.

Also added a func test for when the src is missing.

### DIFF
--- a/cmd/stack_create.go
+++ b/cmd/stack_create.go
@@ -15,14 +15,9 @@
 package cmd
 
 import (
-	"archive/zip"
-	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
-	"runtime"
-	"strings"
-	"time"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -51,8 +46,6 @@ The stack name must start with a lowercase letter, and can contain only lowercas
   appsody stack create my-stack --copy incubator/nodejs-express  
   Creates a stack called my-stack, based on the Node.js Express stack.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-
-			currentTime := time.Now().Format("20060102150405")
 
 			if len(args) < 1 {
 				return errors.New("Required parameter missing. You must specify a stack name")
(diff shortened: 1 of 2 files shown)


cmd/stack_create_test.go
```diff
@@ -137,3 +138,19 @@ func TestStackCreateSampleStackDryrun(t *testing.T) {
...
+	defer cleanup()
+
+	createArgs := []string{"stack", "create", "testing-stack", "--copy", "invalid/starter"}
+	output, err := cmdtest.RunAppsody(sandbox, createArgs...)
+	if err != nil {
+		if !strings.Contains(output, "Repository: 'invalid' was not found in the repository.yaml file") {
+			t.Errorf("String \"Repository: 'invalid' was not found in the repository.yaml file\" not found in output")
+		}
+	} else {
+		t.Error("Stack create command unexpectededly passed with an invalid repository name")
```
upLukeWeston: Unexpected typo in the error area.

functest/create_test.go
```diff
@@ -14,163 +14,157 @@
...
-		t.Fatal("Stack doesn't exist despite appsody stack create executing correctly.")
+func TestFuncCreateValidCases(t *testing.T) {
+	var stackCreateValidTests = []struct {
+		testName   string
+		addToRepo  bool
+		createArgs []string
+		stackName  string
+	}{
+		{"Create with dev.local stack", false, []string{"stack", "create", "testing-stack", "--copy", "dev.local/starter"}, "testing-stack"},
+		{"Create with custom repo stack", true, []string{"stack", "create", "testing-stack", "--copy", "test-repo/starter"}, "testing-stack"},
```
upLukeWeston: Just a question: if all of the cases in this test will be `stack create`'s, is it necessary to include it in each argument array? You could instead do a `args := append([]string{"stack", "create"}, tt.createArgs...)` in the future!

cmd/stack_create.go
```diff
@@ -154,8 +147,7 @@ The stack name must start with a lowercase letter, and can contain only lowercas
 
 				if stackEntryURL == "" {
 					//TODO: REMOVE OLD CREATE METHOD AFTER NEXT RELEASE AND UPDATE STACKS
-					//return errors.New("No source URL specified.  Use the add-to-repo command with the --release-url flag to your repo")
-					return oldCreateMethod(rootConfig.LoggingConfig, rootConfig, config, stack, currentTime)
+					return errors.New("No source URL specified.  Use the add-to-repo command with the --release-url flag to your repo")
```
EnriqueL8: Could we be more specific on where the source url is supposed to be? (Also maybe remove the TODO haha)

neeraj-laad:
@CameronMcWilliam PRs like this should be araised as draft PRs. to ensure no one accidentally merges them and it also means we can filter them out easily.

CameronMcWilliam:
> @CameronMcWilliam PRs like this should be araised as draft PRs. to ensure no one accidentally merges them and it also means we can filter them out easily.

It was a draft PR that turned to a normal PR as I thought we were willing to merge this after all the incubator stacks had the source url present, which they now all do.

PR #891, opened 2020-02-10
Added remove packaging folder at the start of stack package

CameronMcWilliam:
Even though we had a defer that removes the packaging folder where templating takes place.  If a user used CTRL-C after `stack package` it will not remove the packaging folder and would error on the next use, then remove it on the defer.  I have added a `removeAll` to the start of the function as to avoid this error.

### DIFF
--- a/cmd/stack_package.go
+++ b/cmd/stack_package.go
@@ -155,7 +155,7 @@ The packaging process builds the stack image, generates the "tar.gz" archive fil
 			defer func() {
 				err = removePackagingFolder(stackPath)
 				if err != nil {
-					errors.Errorf("Error removing packaging folder: %v", err)
+					log.Info.logf("Error removing packaging folder: %v", err)
 				}
 			}()
 
--- a/cmd/stack_package.go
+++ b/cmd/stack_package.go
@@ -148,20 +148,16 @@ The packaging process builds the stack image, generates the "tar.gz" archive fil
 
 			err = removePackagingFolder(stackPath)
 			if err != nil {
-				return err
+				return errors.Errorf("Error removing packaging folder: %v", err)
 			}
 
 			// defer function to remove packaging folder created for stack variables
-			defer func() error {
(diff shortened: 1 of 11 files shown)


cmd/stack_package.go
```diff
@@ -146,6 +146,12 @@ The packaging process builds the stack image, generates the "tar.gz" archive fil
 				return errors.Errorf("Error creating stacks directory: %v", err)
 			}
 
+			err = os.RemoveAll(stackPath)
+
+			if err != nil {
+				return errors.Errorf("Error creating removing packaging directory: %v", err)
```
EnriqueL8: error message typo

cmd/stack_package.go
```diff
@@ -146,6 +146,12 @@ The packaging process builds the stack image, generates the "tar.gz" archive fil
 				return errors.Errorf("Error creating stacks directory: %v", err)
 			}
 
+			err = os.RemoveAll(stackPath)
```
EnriqueL8: should we remove the one in line 158 `os.RemoveAll(stackPath)` and the `defer` in 163?

CameronMcWilliam: Don't think so.  We shouldn't keep the packaging folder if possible.

cmd/stack_package.go
```diff
@@ -673,3 +682,17 @@ func ApplyTemplating(stackPath string, templateMetadata interface{}) error {
 	return nil
 
 }
+
+func removePackagingFolder(stackPath string) error {
```
EnriqueL8: This function looks good, thanks! Just a suggestion would be to rename it to `removePackagingFolderIfExists` or even rename it to `removeFolderIfExists` and move it to `utils.go`, but I will leave this to your judgement!

CameronMcWilliam: I agree I like this more generic approach.  I will do that now.

Issue #736, opened 2019-11-29
Code coverage analysis for repo_list.go


skoh7645:
- [x] func newRepoListCmd & if repoErr != nil
test repo list command with an incorrectly formatted `repository.yaml` file and verify the error

- [ ] func newRepoListCmd & if err != nil
simulate unmarshal error with `repo list --output yaml` and `repo list --output json` commands and verify the error

PR #898, opened 2020-02-11
repo_list.go: improve code coverage

skoh7645:
Bumps code coverage of `repo_list_test.go` from 84% to 90%

Fixes: #736

### DIFF
--- a/cmd/repo_add_test.go
+++ b/cmd/repo_add_test.go
@@ -99,7 +99,7 @@ func TestRepoAddErrors(t *testing.T) {
 		{"No args", nil, "", "You must specify a repository name and URL"},
 		{"One arg", []string{"reponame"}, "", "You must specify a repository name and URL"},
 		{"No url scheme", []string{"test", "localhost"}, "", "unsupported protocol scheme"},
-		{"Non-existing url", []string{"test", "http://localhost/doesnotexist"}, "", "refused"},
+		{"Non-existing url", []string{"test", "http://localhost/doesnotexist"}, "", "Error"},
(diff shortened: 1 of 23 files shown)


cmd/repo_add_test.go
```diff
@@ -99,7 +99,7 @@ func TestRepoAddErrors(t *testing.T) {
 		{"No args", nil, "", "You must specify a repository name and URL"},
 		{"One arg", []string{"reponame"}, "", "You must specify a repository name and URL"},
 		{"No url scheme", []string{"test", "localhost"}, "", "unsupported protocol scheme"},
-		{"Non-existing url", []string{"test", "http://localhost/doesnotexist"}, "", "refused"},
+		{"Non-existing url", []string{"test", "http://localhost/doesnotexist"}, "", "Error"},
```
EnriqueL8: Is there a better error message that could be returned here?

skoh7645: There was an issue with this test @kilnerm brought up a while ago (that I managed to replicate). Depending on whether you are connected to IBM or IBMInternet wifi, we get differing error messages- has to do with the way IBMInternet reroutes unknown URLs I think. We still want the command to fail if the URL is unknown, so I've kept the test in, but we can't predict which error message will be given - only that one will be given

EnriqueL8: Makes sense, maybe we could add a generic error message that wraps the error returned from the download function [here](https://github.com/appsody/appsody/blob/master/cmd/repo_add.go#L75) - something like `Error("Error downloading index: %s", err)`

Issue #717, opened 2019-11-27
Code coverage analysis for git_utils.go


skoh7645:
- [x] func GetGitInfo
test build passes with no error

- [ ] if strings.Contains(value, branchSeparatorString)
simulate build where git labels include a separator string "..." and verify no error

- [x] if strings.Contains(value, noCommits)
simulate build where git info includes a git branch has no commits and verify no error

- [ ] if gitInfo.Upstream != ""
simulate build with and invalid gitInfo.Upstream and verify error

PR #902, opened 2020-02-12
git_utils.go: improve code coverage

skoh7645:
Bumps code coverage of `git_utils.go` from 80% to 91%
I had to capitalise some functions to export them for testing

Fixes: #717

### DIFF
--- a/cmd/git_utils_test.go
+++ b/cmd/git_utils_test.go
@@ -1,0 +1,114 @@
+// Copyright © 2019 IBM Corporation and others.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd_test
+
+import (
+	"bytes"
+	"strings"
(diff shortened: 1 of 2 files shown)


cmd/git_utils_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+func TestStringBefore(t *testing.T) {
+	var StringBeforeTests = []struct {
+		testName       string
+		value          string
+		searchString   string
+		expectedOutput string
+	}{
+		{"Empty values", "", "", ""},
+		{"Non-empty values", "teststring", "string", "test"},
```
EnriqueL8: one more test where the search string is not in the string value, should return empty string

cmd/git_utils_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+func TestGetGitInfoWithNotAGitRepo(t *testing.T) {
+	sandbox, cleanup := cmdtest.TestSetupWithSandbox(t, true)
+	defer cleanup()
+
+	var outBuffer bytes.Buffer
+	loggingConfig := &cmd.LoggingConfig{}
+	loggingConfig.InitLogging(&outBuffer, &outBuffer)
+	config := &cmd.RootCommandConfig{LoggingConfig: loggingConfig}
+	cwd, err := os.Getwd()
```
EnriqueL8: Will need to investigate further on this sandbox and chang dir issue?

upLukeWeston: Done!

cmd/git_utils_test.go
upLukeWeston: Should it return an empty string? Because when I just tested it, it returns the `value` string in full. 

For example, in this case
```
output := cmd.StringBefore("This is what I have", "But this is what I am looking for")
```
`output` = "This is what I have"

Issue #733, opened 2019-11-29
Code coverage analysis for operator_uninstall.go 


Kamran64:
- [ ] `if config.namespace != ""`
    Test with and without namespace set.

- [ ] `removeErr := removeOperator(operatorNamespace, config)`
    Attempt to remove operator using existing and non existing namespace

- [ ] `config.Debug.log("Appsody operator count is: ", operCount)`
    Verify number of operators. For the test case, we could have 1 and verify the printed count is 1.

`func removeOperatorCRDs(config *operatorUninstallCommandConfig) error`
- [ ] `deployConfigDir, err := getDeployConfigDir(config.RootCommandConfig)`
    Run test with missing deploy dir and verify error.

- [ ] ` _, err = downloadCRDYaml(config.LoggingConfig, crdURL, appsodyCRD)`
    Run test with and without invalid URL to a CRD yaml

- [ ] `if !config.Dryrun`
    Run test in dry run mode and verify this code path isn't touched.
    `err = os.Remove(appsodyCRD)`
    Run test out of dry run mode and pass in valid and invalid appsody crd and verify error with invalid crd.

`func removeOperatorRBAC(operatorNamespace string, config *operatorUninstallCommandConfig) error`
- [ ] Repeat tests in func `removeOperatorCRDs` for this function.

`func removeOperator(operatorNamespace string, config *operatorUninstallCommandConfig) error`
- [ ] Run tests after operator has been installed from `operator_install.go` tests and verify that it can be successfully removed.

- [ ] Run tests in dry run mode and verify some code paths aren't touched.

Issue #732, opened 2019-11-29
Code coverage analysis for operator_install.go


Kamran64:
- [ ] `namespace := config.namespace`, `watchspace := config.watchspace`
    Run tests with and without namespace and watchspace set.

- [ ] `if operatorExists`
    Run tests with operator already existing in namespace and verify error message.

- [ ] `if watchExists`
    Run tests with operator already watching namespace and verify error.

- [ ] `deployConfigDir, err := getDeployConfigDir(config.RootCommandConfig)`
    Pass in invald config to function and verify err is returned.

- [ ] `if (operatorNamespace != watchNamespace) || config.all`
    Run tests with operator namespace being different to the watch namespace. Pass in valid and invalid url to RBAC yaml function.

- [ ] `file, err = downloadOperatorYaml(config.LoggingConfig, operatorURL, operatorNamespace, watchNamespace, operatorYaml)`
    Run tests when operatorNamespace and watchNamespace are the same and pass in valid and invalid operator yaml's.

Issue #731, opened 2019-11-29
Code coverage analysis for operator.go


Kamran64:
- [ ] `file, err := downloadYaml(log, url, target)`
    Pass in invalid URL or target and verify that Operator yaml fails to install. Also pass in valid yaml to check that it passes.

- [ ] `yamlReader, err := ioutil.ReadFile(file)`
    Delete downloaded file from test 1, and validate error message when failing to read file.

- [ ]  `err = ioutil.WriteFile(target, output, 0666)`
    Change permissions of file to read only so that write fails. Validate error message.

- [ ] `if dryrun`
    Run command in dry run mode and verify correct messages are outputted

- [ ] Repeat tests 1 - 3 for downloading RBAC yamls.

- [ ] `func downloadYaml(log *LoggingConfig, url string, target string) (string, error)`
    Pass in url to file with invalid yaml structure and verify error and then valid yaml test.

- [ ] `func getDeployConfigDir(rootConfig *RootCommandConfig) (string, error)`
    Run without deploy dir and check error. Create deploy dir and then run again and check that it passes.

PR #903, opened 2020-02-12
operator commands: Improve code coverage

Kamran64:
- These tests can only be run locally as travis currently does not support kube commands.

Fixes: #731 
Fixes: #732 
Fixes: #733

### DIFF
--- a/functest/operator_install_test.go
+++ b/functest/operator_install_test.go
@@ -73,7 +73,12 @@ func TestInstallOperatorWithNamespaceAndWatchspace(t *testing.T) {
 	log := &cmd.LoggingConfig{}
 	log.InitLogging(&outBuffer, &outBuffer)
 
-	defer removeNamespace(log, "namespace-for-test-operator-install")
+	defer func() {
+		err := removeNamespace(log, "namespace-for-test-operator-install")
+		if err != nil {
+			t.Fatalf("%s", err)
+		}
+	}()
 	expectedLogs := "Appsody operator deployed to Kubernetes"
 
 	namespaceKargs := []string{"create", "namespace", "namespace-for-test-operator-install"}
--- a/functest/operator_install_test.go
+++ b/functest/operator_install_test.go
@@ -43,19 +43,12 @@ func TestOperatorInstallCases(t *testing.T) {
 			sandbox, cleanup := cmdtest.TestSetupWithSandbox(t, true)
 			defer cleanup()
 
-			t.Log("Now running appsody init")
-			args := []string{"init", "starter"}
-			_, err := cmdtest.RunAppsody(sandbox, args...)
-			if err != nil {
-				t.Fatal(err)
-			}
-
 			t.Log("Now running appsody operator install")
 			operatorInstallArgs := append([]string{"operator", "install"}, tt.args...)
 			output, operatorErr := cmdtest.RunAppsody(sandbox, operatorInstallArgs...)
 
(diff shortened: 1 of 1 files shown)


functest/operator_install_test.go
```diff
@@ -0,0 +1,127 @@
...
+		if !cmdtest.TravisTesting {
+			t.Skip()
+		}
+
+		tt := testData
+		t.Run(tt.testName, func(t *testing.T) {
+			sandbox, cleanup := cmdtest.TestSetupWithSandbox(t, true)
+			defer cleanup()
+
+			t.Log("Now running appsody init")
```
EnriqueL8: You wouldn't need to initialise a project to install the operator

functest/operator_install_test.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/pkg/errors"
+)
+
+func TestOperatorInstallCases(t *testing.T) {
+	var operatorInstallTests = []struct {
+		testName     string
+		args         []string
+		expectedLogs string
+	}{
+		{"Run in dryrun mode", []string{"--dryrun"}, "Appsody operator deployed to Kubernetes"},
```
EnriqueL8: Not sure the expected log is the desired output for a `dryrun` flag

functest/operator_install_test.go
```diff
@@ -0,0 +1,127 @@
...
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			t.Log("Now running appsody operator install")
+			operatorInstallArgs := append([]string{"operator", "install"}, tt.args...)
+			output, operatorErr := cmdtest.RunAppsody(sandbox, operatorInstallArgs...)
+
+			if !strings.Contains(output, tt.expectedLogs) {
+				t.Fatalf("Expected failure to include: %s but instead receieved: %s. Full error: %s", tt.expectedLogs, output, operatorErr)
```
EnriqueL8: typo in `receieved`

functest/operator_install_test.go
```diff
@@ -0,0 +1,127 @@
...
+				}
+			}()
+
+			namespaceKargs := []string{"create", "namespace", tt.namespace}
+			_, namespaceErr := cmd.RunKube(log, namespaceKargs, false)
+			if namespaceErr != nil {
+				t.Fatal(namespaceErr)
+			}
+
+			t.Log("Now running appsody init")
```
EnriqueL8: Same as comment above

functest/operator_install_test.go
```diff
@@ -0,0 +1,127 @@
...
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			t.Log("Now running appsody operator command")
+			operatorArgs := append([]string{"operator"}, tt.args...)
+			output, operatorErr := cmdtest.RunAppsody(sandbox, operatorArgs...)
+
+			if !strings.Contains(output, tt.expectedLogs) {
+				t.Fatalf("Expected failure to include: %s but instead receieved: %s. Full error: %s", tt.expectedLogs, output, operatorErr)
```
EnriqueL8: same typo

functest/operator_install_test.go
Kamran64: I put that as the expected log since thats printed when the dryrun succeeds. I could either change this to check that no error has been received or just remove the test altogether if we're thinking of removing the dry run flag. Thoughts?

Issue #897, opened 2020-02-11
Couldn't remove the stacks from repo "dev.local"


brightzheng100:
**Describe the bug**
During the development of Apposody Stack, I found that I **couldn't delete the Stacks** that are under `dev.local` repo, which is the default repo if nothing is specified while issuing `appsody stack package`.

**To Reproduce**

```
$ appsody repo list
NAME      	URL
*dev.local	file:///Users/brightzheng/.appsody/stacks/dev.local/dev.local-index.yaml

$ appsody list dev.local
REPO     	ID                    	VERSION  	TEMPLATES	DESCRIPTION
dev.local	springboot-swagger-jpa	0.1.0    	*default 	A Stack managed by Maven, with Spring Boot, Swagger, JPA

$ appsody stack remove-from-repo "dev.local" springboot-swagger-jpa
******************************************
Running appsody stack remove-from-repo
******************************************
Repository index file not found - unable to remove stack
```

**Expected behavior**
Once issued `appsody stack remove-from-repo <REPO> <STACK>`, the Stack should be removed from repo accordingly.

**Actual behaviour**
Error occurred: 

```
Repository index file not found - unable to remove stack
```

**Environment Details (please complete the following information):**
 - OS: Mac

**If applicable please specify:**
 - CLI version: appsody 0.5.6
 - Stack you are using: A new stack that I'm developing but it applies to any Stack under `dev.local` repo.

CameronMcWilliam:
Closed with #904

PR #904, opened 2020-02-13
Refactored remove-from-repo 

CameronMcWilliam:
Fixes issue #897 .  Refactored the remove-from-repo command to provide more useful error messages and also added tests.

### DIFF
--- a/cmd/stack_addtorepo.go
+++ b/cmd/stack_addtorepo.go
@@ -298,6 +298,11 @@ Run this command from the root directory of your Appsody project.`,
 				return errors.Errorf("Error writing localIndexFile: %v", err)
 			}
 
+			err = generateCodewindJSON(log, indexYaml, localIndexFile, repoName)
+			if err != nil {
+				return errors.Errorf("Could not generate json file from yaml index: %v", err)
+			}
+
 			log.Info.Log("Repository index file updated successfully")
 
 			return nil
--- a/cmd/stack_addtorepo.go
+++ b/cmd/stack_addtorepo.go
@@ -224,8 +224,7 @@ Run this command from the root directory of your Appsody project.`,
 
 			// At this point we should have the indexFile loaded that want to use for updating / adding stack info
 			// find the index of the stack
-			indexYaml, err = findStackAndRemove(log, stackID, indexYaml)
-
+			indexYaml, _, err = findStackAndRemove(log, stackID, indexYaml)
 			if err != nil {
 				return err
 			}
--- a/cmd/stack_addtorepo.go
+++ b/cmd/stack_addtorepo.go
@@ -224,7 +224,11 @@ Run this command from the root directory of your Appsody project.`,
 
 			// At this point we should have the indexFile loaded that want to use for updating / adding stack info
 			// find the index of the stack
-			indexYaml = findStackAndRemove(log, stackID, indexYaml)
+			indexYaml, err = findStackAndRemove(log, stackID, indexYaml)
+
+			if err != nil {
+				return err
+			}
 
 			// get the necessary data from the current stack.yaml
 			stackYaml, err := getStackData(stackPath)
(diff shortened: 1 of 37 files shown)


cmd/stack_package.go
```diff
@@ -493,8 +496,9 @@ func initialiseStackData(stackID string, stackImage string, stackYaml StackYaml)
 	return newStackStruct
 }
 
-func findStackAndRemove(log *LoggingConfig, stackID string, indexYaml IndexYaml) IndexYaml {
+func findStackAndRemove(log *LoggingConfig, stackID string, indexYaml IndexYaml) (IndexYaml, bool, error) {
```
EnriqueL8: I don't believe you need to return an `error`, since we will always return `nil`.

cmd/stack_addtorepo.go
```diff
@@ -224,7 +224,10 @@ Run this command from the root directory of your Appsody project.`,
 
 			// At this point we should have the indexFile loaded that want to use for updating / adding stack info
 			// find the index of the stack
-			indexYaml = findStackAndRemove(log, stackID, indexYaml)
+			indexYaml, _, err = findStackAndRemove(log, stackID, indexYaml)
+			if err != nil {
```
EnriqueL8: We could verify that the `stackExists` doesn't fail here

cmd/stack_removefromrepo_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+	foundStack := -1
+	for i, stack := range indexYaml.Stacks {
+		if stack.ID == "nodejs" {
+			foundStack = i
+			break
+		}
+	}
+	if foundStack != -1 {
+		t.Fatal("Stack found unexpectecdly")
```
EnriqueL8: typo in message

cmd/stack_package.go
CameronMcWilliam: Yeah that was left over from a previous commit, removed it now cheers.

cmd/stack_removefromrepo_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+	foundStack := -1
+	for i, stack := range indexYaml.Stacks {
+		if stack.ID == "nodejs" || stack.ID == "kitura" {
+			foundStack = i
+			break
+		}
+	}
+	if foundStack != -1 {
+		t.Fatal("Stack found unexpectecdly")
```
upLukeWeston: Typo in error message

cmd/stack_removefromrepo.go
```diff
@@ -121,25 +121,14 @@ The updated repository index file is created in  ~/.appsody/stacks/dev.local dir
...
 						return errors.Errorf("Error checking status of %s", localIndexFile)
 					}
-					if exists && useLocalCache {
-						log.Debug.Log(localIndexFile, " exists in the appsody directory and use-local-cache is true")
+					if exists {
+						log.Debug.Log(localIndexFile, " exists in the appsody directory")
 					} else {
-						log.Info.Log("Repository index file not found - unable to remove stack")
-						return nil
+						return errors.Errorf("Repository index file not found - unable to remove stack")
```
upLukeWeston: After speaking to @groeges - it would be better if we kept this as it was, e.g. printing an Info in this case, maybe a Debug in other cases and returning nil from the function.

cmd/stack_removefromrepo.go
```diff
@@ -121,25 +121,15 @@ The updated repository index file is created in  ~/.appsody/stacks/dev.local dir
...
-				if err != nil {
-					return errors.Errorf("Error checking status of %s", localIndexFile)
-				}
-				if exists && useLocalCache {
-					log.Debug.Log(localIndexFile, " exists in the appsody directory and use-local-cache is true")
-				} else {
-					log.Info.Log("Repository index file not found - unable to remove stack")
-					return nil
-				}
+				return errors.Errorf("%v does not exist within the repository list", repoName)
```
upLukeWeston: Would this also need to be changed to printing a debug/info and returning a `nil`? Or is it acceptable to keep it as an error if you're explicitly trying to remove a stack that isn't found?

Issue #909, opened 2020-02-14
appsody build --buildah looks for docker CLI


nastacio:
**Describe the bug**
As context, I was trying to build an appsody project inside https://learn.openshift.com/playgrounds/openshift42/

It looks like `appsody build --buildah` attempted to execute the `docker image ls ...` command, but I believe the --buildah option should  preclude invocations of the docker CLI. Maybe it is just a bad message after an unexpected `buildah` failure (see below) , but the combination of error and ensuing message causes the impression that the `--buildah` option does not prevent appsody from depending on the `docker` CLI.

**To Reproduce**
1. Launch a playground at https://learn.openshift.com/playgrounds/openshift42/
2. Type the following commands
```
curl -sL https://github.com/appsody/appsody/releases/download/0.5.8/appsody-0.5.8-1.x86_64.rpm -o appsody.rpm
curl -sL https://cbs.centos.org/kojifiles/packages/libseccomp/2.4.1/0.el7/x86_64/libseccomp-2.4.1-0.el7.x86_64.rpm -o libseccomp.rpm
yum install -y ./appsody.rpm ./libseccomp.rpm buildah 

git clone https://github.com/nastacio/appsody-nodejs
cd appsody-nodejs
appsody build --buildah
```

**Expected behavior**
I expected it to work, but it failed due to a problem with buildah (bad, but not important) .

**Actual behaviour**
The ensuing error message indicates that `appsody` attempted to invoke `docker` in the error path:

```
Your Appsody project name has been set to appsody-nodejs
Extracting project from development environment
Pulling docker image docker.io/kabanero/nodejs-express:0.2
Running command: buildah pull docker.io/kabanero/nodejs-express:0.2
Error during unshare(CLONE_NEWUSER): Operation not permitted
level=error msg="error parsing PID \"\": strconv.Atoi: parsing \"\": invalid syntax"
level=error msg="(unable to determine exit status)"
[Warning] Docker image pull failed: exit status 1
[Warning] Could not run docker image ls -q for the image: kabanero/nodejs-express:0.2 error: exec: "docker": executable file not found in$PATH Check to make sure docker is available.
[Error] Could not find the image either in docker hub or locally: docker.io/kabanero/nodejs-express:0.2
```

**Environment Details (please complete the following information):**
 - OpenShift 4.2 Playground

**If applicable please specify:**
 - CLI version: appsody 0.5.8
 - Stack you are using:
[...]

Issue #880, opened 2020-02-04
Function checkDockerImageExistsLocally should have a buildah option


chilanti:
**Describe the bug**
This function checks whether the image is present in the local image cache, but it presumes the availability of the `docker` client (it won't work in case it is invoked with `--buildah` and in an environment where `docker` is not installed

**Expected behavior**
Use the appropriate `buildah` command when `--buildah` is set.

PR #911, opened 2020-02-17
Added Buildah pull image option

CameronMcWilliam:
Fixes issues: #909 and #880 , made the `checkDockerImageExistsLocally` option to a more generic `checkImageExistsLocally` allowing for a `buildah pull` instead of a `docker pull`

### DIFF
--- a/cmd/cmdtest/testutils.go
+++ b/cmd/cmdtest/testutils.go
@@ -389,3 +389,11 @@ func Exists(path string) (bool, error) {
 	}
 	return true, err
 }
+
+func GetEnvStacksList() string {
+	stacksList := os.Getenv("STACKS_LIST")
+	if stacksList == "" {
+		stacksList = "incubator/nodejs"
+	}
+	return stacksList
+}
--- a/cmd/cmdtest/testutils.go
+++ b/cmd/cmdtest/testutils.go
@@ -341,7 +341,7 @@ func AddLocalRepo(t *TestSandbox, repoName string, repoFilePath string) (string,
 // The stdout and stderr are captured, printed, and returned
 // args will be passed to the docker command
 // workingDir will be the directory the command runs in
-func RunCmdExec(args []string, cmdName string, t *testing.T) (string, error) {
+func RunCmdExec(cmdName string, args []string, t *testing.T) (string, error) {
 	cmdArgs := []string{cmdName}
 	cmdArgs = append(cmdArgs, args...)
 	t.Log(cmdArgs)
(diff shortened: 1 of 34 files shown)


CameronMcWilliam:
Having to remove the test case I created for it due to our current limitations with Buildah version on Travis.  I have tested this locally with my `appsody buildah` image where `docker` is not present and it is working as intended.

functest/build_test.go
```diff
@@ -37,52 +38,85 @@ type expectedDeploymentConfig struct {
...
+				if tt.buildah {
+					listOutput, listErr := cmdtest.RunBuildahCmdExec([]string{"images", "-q", expectedImageTag}, t)
+					if listErr != nil {
+						t.Fatal(listErr)
+					}
+					if listOutput == "" {
+						t.Errorf("Expected appsody build to create buildah image '%s' but it was not found.", expectedImageTag)
+					}
+					//delete the image
+					deleteImage(expectedImageTag, true, t)
```
EnriqueL8: We could pass `t.Buildah` so that the code is nicer to read and follow instead of `true/false`

functest/build_test.go
```diff
@@ -37,52 +38,85 @@ type expectedDeploymentConfig struct {
...
-		listOutput, listErr := cmdtest.RunDockerCmdExec([]string{"images", "-q", expectedImageTag}, t)
-		if listErr != nil {
-			t.Fatal(listErr)
-		}
-		if listOutput == "" {
-			t.Errorf("Expected appsody build to create docker image '%s' but it was not found.", expectedImageTag)
-		}
+				expectedImageTag := "dev.local/" + sandbox.ProjectName
+				expectedImageTag = strings.Replace(expectedImageTag, "_", "-", -1)
+				if tt.buildah {
```
EnriqueL8: We could reduce duplication of code by creating a var for `[]string{..args..}` and the if statements.

cmd/cmdtest/testutils.go
```diff
@@ -377,6 +377,46 @@ func RunDockerCmdExec(args []string, t *testing.T) (string, error) {
 	return outBuffer.String(), err
 }
 
+// RunBuildahCmdExec runs the buildah command with the given args in a new process
+// The stdout and stderr are captured, printed, and returned
+// args will be passed to the buildah command
+// workingDir will be the directory the command runs in
+func RunBuildahCmdExec(args []string, t *testing.T) (string, error) {
```
EnriqueL8: Just a suggestion, this cmd is very similar  to the docker one. We could add a parameter to `RunDockerCmdExec` for the command `buildah/docker` and rename it to `RunCmdExec`?

PR #913, opened 2020-02-17
Support Open liberty operator

EnriqueL8:
Creating a generic deployment manifest struct which can be used to unmarshall and marshall any kubernetes deployment manifest.
This allows the appsody cli to work with other deployment manifest, instead of just the Appsody Operator CRD.


cmd/utils.go
```diff
@@ -52,6 +52,7 @@ type ProjectConfig struct {
 	Description     string
 	License         string
 	Maintainers     []Maintainer
+	Operator        string
```
kylegc: Is this used anywhere?

EnriqueL8: Oh, I had the idea to add an option for users to specify which operator they want to deploy with. I think that is too much out of scope, so will remove it.

Issue #908, opened 2020-02-13
Allow generation of a CodeWind .json while building stack locally


skoh7645:
**Is your feature request related to a problem? Please describe.**
We need to provide a mechanism and document the process for generating an `index.json` file for CodeWind while developing stacks locally.

This issue will track the CLI changes needed for https://github.com/appsody/stacks/issues/668:
- Update Appsody CLI to have a utility functions that can do this step and call that step during all CLI commands that create/update a repo index (stack package, stack add-to-repo, stack remove-from-repo)

PR #914, opened 2020-02-17
appsody stack: generate json index

skoh7645:
Generates an `index.json` file for CodeWind in the commands:
- `stack package`
- `stack add-to-repo`
- `stack remove-from-repo`

Fixes: https://github.com/appsody/appsody/issues/908

### DIFF
--- a/cmd/stack_package.go
+++ b/cmd/stack_package.go
@@ -76,10 +76,6 @@ type IndexYamlStackTemplate struct {
 	URL string `yaml:"url"`
 }
 
-// type IndexJson struct {
-// 	Stacks []IndexJsonStack
-// }
-
 // struct to convert yaml to json files
 type IndexJsonStack struct {
 	DisplayName  string `json:"displayName"`
@@ -190,7 +186,7 @@ The packaging process builds the stack image, generates the "tar.gz" archive fil
 			// get the necessary data from the current stack.yaml
 			stackYaml, err = getStackData(stackPath)
 			if err != nil {
-				return err
+				return errors.Errorf("Error parsing the stack.yaml file: %v", err)
 			}
 
 			// check for templates dir, error out if its not there
--- a/cmd/stack_package.go
+++ b/cmd/stack_package.go
@@ -77,7 +77,7 @@ type IndexYamlStackTemplate struct {
 }
 
 // struct to convert yaml to json files
-type IndexJsonStack struct {
+type IndexJSONStack struct {
 	DisplayName  string `json:"displayName"`
 	Description  string `json:"description"`
 	Language     string `json:"language"`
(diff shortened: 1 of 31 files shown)


cmd/utils.go
```diff
@@ -2322,3 +2322,39 @@ func RemoveIfExists(path string) error {
...
+			indexJSONStack = append(indexJSONStack, stackJSON)
+		}
+	}
+
+	// Last thing to do is write the data to the file
+	data, err := json.MarshalIndent(&indexJSONStack, "", "	")
+	if err != nil {
+		return err
+	}
+	indexFilePath = strings.ReplaceAll(indexFilePath, ".yaml", ".json")
```
EnriqueL8: We could be stricter here by only replacing once

cmd/utils.go
```diff
@@ -2322,3 +2322,39 @@ func RemoveIfExists(path string) error {
 	}
 	return nil
 }
+
+func generateJSON(log *LoggingConfig, indexYaml IndexYaml, indexFilePath string) error {
```
EnriqueL8: We could rename this function to generateCodewindJSON, because the format is specific to them.

cmd/utils.go
```diff
@@ -2322,3 +2322,39 @@ func RemoveIfExists(path string) error {
...
+			stackJSON := IndexJSONStack{}
+			stackJSON.DisplayName = "Appsody " + stack.Name + " " + template.ID + " template"
+			stackJSON.Description = stack.Description
+			stackJSON.Language = stack.Language
+			stackJSON.ProjectType = "appsodyExtension"
+			stackJSON.ProjectStyle = "Appsody"
+			stackJSON.Location = template.URL
+
+			link := Link{}
+			link.Self = "/devfiles/" + stack.ID + "/devfile.yaml"
```
EnriqueL8: As @CameronMcWilliam mentioned, could we use `filepath.Join`?

cmd/stack_package.go
```diff
@@ -76,6 +76,21 @@ type IndexYamlStackTemplate struct {
...
+	DisplayName  string `json:"displayName"`
+	Description  string `json:"description"`
+	Language     string `json:"language"`
+	ProjectType  string `json:"projectType"`
+	ProjectStyle string `json:"projectStyle"`
+	Location     string `json:"location"`
+	Links        Link   `json:"links"`
+}
+
+type Link struct {
```
EnriqueL8: we could just rename this to `Links`

cmd/utils.go
skoh7645: Using `filepath.Join` on Windows would result in backslashes, which we don't want. See [here](https://github.com/kabanero-io/codewind-templates/blob/master/devfiles/index.json) for a valid Codewind index

upLukeWeston:
The documentation will need to be updated to reflect this new feature.

PR #918, opened 2020-02-19
[WIP] Add CRIU support for appsody to create startup optimised images

bharathappali:
This PR is a work in progress for the CRIU support in appsody

What's `CRIU` ? 
`CRIU` (Checkpoint Restore In Userspace) is a linux tool for checkpointing (freezing) a process and writing all its content to files which were later used to restore the process to its same state where its checkpointed (frozen)

Why `CRIU` in appsody ?
As appsody takes care of the  creation, build and deployment of containerized application CRIU can add a great benefit in terms of improving startup time for the applications.

*Note: To make `CRIU` to be implemented we need [support from stack](https://github.com/bharathappali/spring-criu-stack/blob/master/image/project/java-spring-boot2-build.sh) as well (We should have a mechanism to detect if stack is suitable for CRIU.  it can be like reading the stack metadata or checking for a specific file in stack to ensure that its CRIU suitable)* 

The proposal is to make a test run for the application in build process to start and we poll for a specific trigger (that application is ready to be checkpointed) from its output/logs or wait for certain period of time (say 10-20 secs based on the runtime and application) and checkpoint the application save all its file contents (checkpointed files) and create a image from that container by commiting it and making the launch command of the image to run CRIU restore when its started.

So by this we get the benefit of faster application startup (If we consider tomcat startup it might be like bringing down from 20~40 sec to 1~3 sec)

You can get more details on CRIU from this blog by @ashu-mehra https://openliberty.io/blog/2020/02/12/faster-startup-Java-applications-criu.html 

Sample stack to try CRIU : https://github.com/bharathappali/spring-criu-stack

CRIU can be used for any runtime. currently i'm working to have a `java-spring-boot2` stack CRIU ready, Will port the changes to other stacks as well.

Note: Running the application CRIU image will need some linux capabilities. Currently I'm working on the deployment for OpenShift and kubernetes by adding a `app-criu-deploy.yml` in the stack. Will be posting the updates on issue comments and will remove the `WIP` tag in the PR when its ready for review. 

Would like to have your feedback on this proposal :)

Signed-off-by: bharathappali <bharath.appali@gmail.com>

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -44,6 +44,7 @@ type buildCommandConfig struct {
 	knativeFlagPresent   bool
 	namespaceFlagPresent bool
 	namespace            string
+	criu 				bool
 }
 
 //These are the current supported labels for Kubernetes,
@@ -116,6 +117,7 @@ Run this command from the root directory of your Appsody project.`,
 	buildCmd.PersistentFlags().StringVar(&config.pullURL, "pull-url", "", "Remote repository to pull image from.")
 	buildCmd.PersistentFlags().BoolVar(&config.knative, "knative", false, "Deploy as a Knative Service")
 	buildCmd.PersistentFlags().StringVarP(&config.appDeployFile, "file", "f", "app-deploy.yaml", "The file name to use for the deployment configuration.")
+	buildCmd.PersistentFlags().BoolVar(&config.criu, "criu", false, "Makes appsody to build a startup optimized image")
 
 	buildCmd.AddCommand(newBuildDeleteCmd(config))
 	buildCmd.AddCommand(newSetupCmd(config))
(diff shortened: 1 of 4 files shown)


vijaysun-omr:
Is CRIU support in appsody "experimental" in some way ? i.e. would we support a user who falls victim to one of the limitations we listed in the blog post you referred to ?

bharathappali:
@vijaysun-omr yes as of now I would like to propose it as an "experimental" feature that a user can give a try, A warning message or a documentation should be added if we add this feature in appsody.

dinogun:
@vijaysun-omr was talking to @ebullient at DevNexus about adding an experimental flag to the appsody cli to handle this.

PR #921, opened 2020-02-20
Created application-name parameter for appsody init

### LINKED ISSUE #919: init: ability to specify application name

Appsody allow the user to specify an `application-name: ` field in their `.appsody-config.yaml`. If the user adds this field, it is used by downstream tooling like the Appsody Operator and kAppNav.

However, currently the `application-name: ` field has to be added manually by the developer, after the Appsody project has been created.

The developer may know the name of their complete application at `appsody init` time. In this case, they should be able to specify this as a parameter like this: `appsody init --application-name my-app nodejs-express` (for example).

This should cause the generated `.appsody-config.yaml` to contain:

```
application-name: my-app
```

**Additional context**

This is important when Appsody is being used to scaffold/generate a complete microservices scenario. For example, if `appsody init` is being run multiple times for different microservices which are going to be part of one complete application.

In this case, tooling/scripting will want to do something like (pseudo-code):

```
function generate_complete_application(String appName = "my-app", List microservices) {
    for microservice in microservices {
        run("appsody init --application-name $appName $microservice")
    }
}
```

upLukeWeston:
When executing an `appsody init`, the user can specify an `--application-name` which will then be added to the projects `.appsody-config.yaml`.

Fixes #919

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -92,7 +92,7 @@ Use 'appsody list' to see the available stacks and templates.`,
 	initCmd.PersistentFlags().BoolVar(&config.noTemplate, "no-template", false, "Only create the .appsody-config.yaml file. Do not unzip the template project. [Deprecated]")
 	defaultName := defaultProjectName(rootConfig)
 	initCmd.PersistentFlags().StringVar(&config.projectName, "project-name", defaultName, "Project Name for Kubernetes Service.")
-	initCmd.PersistentFlags().StringVar(&config.applicationName, "application-name", "", "Specifies the greater application which this project will belong to.")
+	initCmd.PersistentFlags().StringVar(&config.applicationName, "application-name", "", "Specifies the greater application which this project belongs to.")
 	return initCmd
 }
 
@@ -101,13 +101,13 @@ func initAppsody(stack string, template string, config *initCommandConfig) error
 	if noTemplate {
 		config.Warning.log("The --no-template flag has been deprecated.  Please specify a template value of \"none\" instead.")
 	}
-	valid, err := IsValidParamName(config.projectName, "project-name")
+	valid, err := IsValidProjectName(config.projectName)
 	if !valid {
 		return err
 	}
 
 	if config.applicationName != "" {
-		valid, err := IsValidParamName(config.applicationName, "application-name")
+		valid, err := IsValidApplicationName(config.applicationName)
 		if !valid {
 			return err
 		}
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -33,9 +33,10 @@ import (
 
 type initCommandConfig struct {
 	*RootCommandConfig
-	overwrite   bool
-	noTemplate  bool
-	projectName string
+	overwrite       bool
+	noTemplate      bool
+	projectName     string
+	applicationName string
 }
 
 // these are global constants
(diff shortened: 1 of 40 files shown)


cmd/init.go
```diff
@@ -90,7 +91,8 @@ Use 'appsody list' to see the available stacks and templates.`,
 	initCmd.PersistentFlags().BoolVar(&config.overwrite, "overwrite", false, "Download and extract the template project, overwriting existing files.  This option is not intended to be used in Appsody project directories.")
 	initCmd.PersistentFlags().BoolVar(&config.noTemplate, "no-template", false, "Only create the .appsody-config.yaml file. Do not unzip the template project. [Deprecated]")
 	defaultName := defaultProjectName(rootConfig)
-	initCmd.PersistentFlags().StringVar(&config.projectName, "project-name", defaultName, "Project Name for Kubernetes Service")
+	initCmd.PersistentFlags().StringVar(&config.projectName, "project-name", defaultName, "Project Name for Kubernetes Service.")
+	initCmd.PersistentFlags().StringVar(&config.applicationName, "application-name", "", "Specifies the greater application which this project will belong to.")
```
EnriqueL8: I would just say `this project belongs to` instead of `will`

cmd/stack_create.go
```diff
@@ -65,7 +65,7 @@ The stack name must start with a lowercase letter, and can contain only lowercas
 
 			stackPath := filepath.Join(config.ProjectDir, stack)
 
-			match, err := IsValidProjectName(stack)
+			match, err := IsValidParamName(stack, "project-name")
 			if !match {
 				return err
```
EnriqueL8: Could we not wrap this error here with info on the `project-name`, instead of passing it everytime?

EnriqueL8: Like we do in https://github.com/appsody/appsody/pull/921/files#diff-4d78da15d8ee41e84a70fdd87c0481e5R61

functest/debug_test.go
```diff
@@ -55,7 +54,7 @@ func TestDebugSimple(t *testing.T) {
 		runChannel := make(chan error)
 		containerName := "testDebugSimpleContainer" + strings.ReplaceAll(stackRaw[i], "/", "_")
 		go func() {
-			_, err := cmdtest.RunAppsody(sandbox, "debug", "--name", containerName, "-P")
+			_, err := cmdtest.RunAppsody(sandbox, "debug", "--name", containerName)
```
EnriqueL8: Should we do this change in this PR? Or did you just miss add it to the commit?

upLukeWeston: Second answer. It's from where my master was ahead of appsody/master!

cmd/utils.go
```diff
@@ -517,6 +525,30 @@ func saveProjectNameToConfig(projectName string, config *RootCommandConfig) erro
...
+	}
+
+	appsodyConfig := filepath.Join(config.ProjectDir, ConfigFile)
+	v := viper.New()
+	v.SetConfigFile(appsodyConfig)
+	err = v.ReadInConfig()
+	if err != nil {
+		return err
+	}
+	v.Set("application-name", applicationName)
```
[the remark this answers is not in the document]
upLukeWeston: I wasn't sure if I should do some checks and only write one if it isn't already present within the config, but that works too!

PR #923, opened 2020-02-21
Added stack deprecation support 

CameronMcWilliam:
Pair programmed with @skoh7645 ! 
 
Related to stack:issue [653](https://github.com/appsody/stacks/issues/653). 

Stack maintainers can now add a `deprecated: <deprecation string>` field in their `stack.yaml` to specify a stack being deprecated, which will appear as a new field in the `appsody-config.yaml`.

We have requested the format to be used to be `"<date> - <deprecation reason>"` (suggestions are welcome on this, if needs to be improved)

The CLI will display warning messages on:

- `init`
- `run`
- `debug`
- `build`
- `deploy`
- `extract`


cmd/build.go
```diff
@@ -224,6 +224,14 @@ func build(config *buildCommandConfig) error {
...
 	}
 
+	deprecated, depErr := GetDeprecated(config.RootCommandConfig)
+	if depErr != nil {
+		return depErr
+	}
+	if deprecated != "" {
+		config.Warning.logf("Stack deprecated: %v", deprecated)
+	}
+
```
skoh7645: You could put this inside the `GetDeprecated` function so you only need to handle the error

cmd/utils.go
```diff
@@ -478,6 +478,22 @@ func getProjectName(config *RootCommandConfig) (string, error) {
...
+	appsodyConfig := filepath.Join(config.ProjectDir, ConfigFile)
+	v := viper.New()
+	v.SetConfigFile(appsodyConfig)
+	err := v.ReadInConfig()
+	if err != nil {
+		return err
+	}
+	if v.Get("deprecated") != nil {
+		deprecatedMessage = v.Get("deprecated").(string)
+		log.Warning.logf("Stack deprercated: %v", deprecatedMessage)
```
groeges: Spelling mistake: "deprercated" should be "deprecated"

groeges:
The warning message isn't very obvious when running the commands and can easily be missed. I had to really look for the message even though I knew it was there. Is there any way the message can be made more visible to the user when running the commands.

CameronMcWilliam:
> The warning message isn't very obvious when running the commands and can easily be missed. I had to really look for the message even though I knew it was there. Is there any way the message can be made more visible to the user when running the commands.

Could look at if can change the colour of the text?

cmd/stack_create.go
```diff
@@ -100,87 +98,39 @@ The stack name must start with a lowercase letter, and can contain only lowercas
...
-
-			err = yaml.Unmarshal(source, &repoFile)
-			if err != nil {
-				return errors.Errorf("Error parsing the repository.yaml file: %v", err)
+				return err
 			}
 
-			// get specificed repo and unmarshal
-			repoEntry := repoFile.GetRepo(repoID)
+			stackEntryURL := stackEntry.SourceURL
```
EnriqueL8: you will need to verift that `stackEntry` is not nil, because in the case of dryrun the function `getStackIndexYaml` will return a nil pointer and this will fail. The tests gave it away

groeges:
@CameronMcWilliam  With these changes, when building a stack that doesn't have `deprecated` in the stack.yaml, should the generated `.appsody-config.yaml` have a `deprecated` element in it? Mine has this:
```
deprecated: ""
stack: docker.io/groeges/node-red:0.1
```

CameronMcWilliam:
> @CameronMcWilliam With these changes, when building a stack that doesn't have `deprecated` in the stack.yaml, should the generated `.appsody-config.yaml` have a `deprecated` element in it? Mine has this:
> 
> ```
> deprecated: ""
> stack: docker.io/groeges/node-red:0.1
> ```

This was an old method, I thought we had removed it, have you tried packaging on the latest version?

groeges:
> This was an old method, I thought we had removed it, have you tried packaging on the latest version?

Validated that with the latest version of code the `deprecated` element is NOT put into the `.appsody-config.yaml`

cmd/stack_package.go
```diff
@@ -386,18 +389,21 @@ The packaging process builds the stack image, generates the "tar.gz" archive fil
 				// create a config yaml file for the tarball
 				configYaml := filepath.Join(templatePath, templates[i], ".appsody-config.yaml")
 				log.Debug.Log("configYaml is: ", configYaml)
-
 				g, err := os.Create(configYaml)
 				if err != nil {
-					return errors.Errorf("Error trying to create file: %v", err)
```
EnriqueL8: Why did we change this error message?

cmd/stack_package.go
```diff
@@ -386,18 +389,21 @@ The packaging process builds the stack image, generates the "tar.gz" archive fil
...
-					return errors.Errorf("Error trying to write: %v", err)
+					return err
+				}
+				v.Set("stack", namespaceAndRepo+":"+semver["majorminor"])
+				err = v.WriteConfig()
+				if err != nil {
+					return err
 				}
-
 				g.Close()
```
EnriqueL8: Could we close the `File` from `os.Create` before reading and setting values with Viper? Just worried this might cause an error

Issue #685, opened 2019-11-21
Delete volume dependencies after running tests


EnriqueL8:
**Is your feature request related to a problem? Please describe.**
Our tests create volumes for the dependencies of every stack tested, and are not deleted.

**Describe the solution you'd like**
Clean up these volumes once the tests have finished. We could use `docker volume prune` since we stop the containers during the tests.

neeraj-laad:
We should not delete other volumes by running `prune`. We need to selectively delete the volumes that we know appsody will create.

PR #925, opened 2020-02-24
testutils.go: clean up docker volumes created by tests

skoh7645:
Added docker volumes clean up to the sandbox cleanup function

Fixes: #685

### DIFF
--- a/cmd/cmdtest/testutils.go
+++ b/cmd/cmdtest/testutils.go
@@ -407,12 +407,12 @@ func getTestDepDockerVolumes(t *testing.T) []string {
 		t.Logf("Error listing docker volumes: %v", err)
 	}
 	volumes := strings.Fields(output)
+	r, _ := regexp.Compile("test([a-z]+)-([0-9]+)-deps")
+	if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/cmdtest/testutils.go
```diff
@@ -397,3 +398,21 @@ func GetEnvStacksList() string {
...
 }
+
+func cleanUpTestDepDockerVolumes(t *testing.T) []string {
+	var testDepVolumes []string
+	output, err := RunCmdExec("docker", []string{"volume", "ls", "--format", "{{.Name}}"}, t)
+	if err != nil {
+		t.Logf("Error listing docker volumes: %v", err)
+	}
+	volumes := strings.Fields(output)
+	for _, volume := range volumes {
```
upLukeWeston: Instead of looping through all of the volumes you can specifically target `t.ProjectName`. That way we wouldn't need to `ls` either - I am not sure if you'd need to verify if the volume exists before attempting to delete it, though; if you execute `docker volume rm <invalid>` it would throw a `No such volume`, but you could probably just print that and move on.

Issue #910, opened 2020-02-16
Change of namespace for "appsody deploy" fails


nastacio:
**Describe the bug**
I deployed an application to the default namespace, then to a different namespace, then attempted to deploy it again against the default namespace, which failed

**To Reproduce**
1. Assuming kubectl pointing to a valid cluster:
```
appsody init incubator/nodejs-express
appsody deploy
appsody deploy delete
appsody deploy -n istio-system
appsody delete
appsody deploy    << deploys against namespace istio-system, which may be a design decision
appsody delete
```
2. Attempt to put the application back in `default` namespace
```
appsody deploy -n default
```

**Expected behavior**
The application should be deployed against the Kubernetes "default" namespace

**Actual behaviour**
The application deployment failed with:
```
Found deployment manifest /Users/nastacio/github.com/icpa-opentracing/nodejs-tracing/app-deploy.yaml
[Error] the namespace "istio-system" from the deployment manifest does not match the namespace "default" passed as an argument.
```

**Environment Details (please complete the following information):**
 - OS: macOS Catalina

**If applicable please specify:**
 - CLI version: appsody 0.5.8
 - Stack you are using:
incubator/nodejs-express:0.4.2, though I don't think it matters

**Screenshots**
n/a

**Additional context**
Editing app-deploy.yaml and replacing `namespace: istio-system` with `namespace: default` worked around the problem.

EnriqueL8:
We decided to go for the `kubectl` approach discussed in #464. 
If a user specifies a namespace arg and the deployment manifest does not contain a namespace value, we write that value passed to the deployment manifest to be used in subsequent calls to `appsody deploy`.

nastacio:
The problem is not that design decision, it ia the fact that subsequent calls with a different namespace do not modify the value in the file and trigger an error.

This is a major problem if app-deploy.yaml is checked into git, which is often the case.

PR #929, opened 2020-02-25
deploy: modify namespace with every call of a different namespace

skoh7645:
Ensure`app-deploy.yaml` gets updated with every call of `--namespace`

Fixes: #910


cmd/deploy.go
```diff
@@ -101,10 +101,15 @@ Run this command from the root directory of your Appsody project.`,
 				manifestNamespace := appsodyApplication.Namespace
 				if manifestNamespace != "" {
 					if namespace != "" && manifestNamespace != namespace {
-						return errors.Errorf("the namespace \"%s\" from the deployment manifest does not match the namespace \"%s\" passed as an argument.", manifestNamespace, namespace)
+						config.Info.Logf("Overriding namespace %s in the deployment manifest to: %s", manifestNamespace, namespace)
+						appsodyApplication.Namespace = namespace
+						err = writeAppsodyApplication(appsodyApplication, configFile)
+						if err != nil {
+							return err
+						}
```
EnriqueL8: We could move all the logic from build into here in order to have everything in one place. I think originally we put it in build so we only write once to the deployment manifest. I think putting it in here is better because in the case where users in the past passed the `--no-build` flag, the namespace won't have been updated.

functest/deploy_test.go
```diff
@@ -172,14 +172,13 @@ func TestDeployNamespaceMismatch(t *testing.T) {
 	t.Logf("Running appsody deploy with namespace: %s ...", secondNamespace)
 	output, err := cmdtest.RunAppsody(sandbox, "deploy", "--generate-only", "-n", secondNamespace)
 
+	expectedOutput := "Overriding namespace " + firstNamespace + " in the deployment manifest to: " + secondNamespace
+	if !strings.Contains(output, expectedOutput) {
+		t.Errorf("Did not get expected output: %s", expectedOutput)
+	}
```
EnriqueL8: Just a suggestion, we could check that the deployment manifest contains the correct namespace

EnriqueL8:
Since #913 has been merged, we can update this PR and resolve any conflicts.

PR #935, opened 2020-03-02
Improve operator error message

EnriqueL8:


neeraj-laad:
Do we need to call out `Not installing Appsody Operator.`? If I am an app developer who is developing app based on a stack that provides a diff operator, I might not even be aware of Appsody operator..

Issue #882, opened 2020-02-05
Stack registry errors on every command


EnriqueL8:
**Describe the bug**
When running an appsody command with verbose, the new stack registry path looks for the stack registry in the config file. This fails if the current directory is not an appsody project, for example during a `stack package` command. See:

![image](https://user-images.githubusercontent.com/30461857/73848678-79edc380-4820-11ea-9dac-c5f1b1536b4e.png)

**To Reproduce**
1. Go to an empty directory
2. Run `appsody -v`
3. See error

**Expected behavior**
This messages should not show up on commands that do not need this check. We should only run this check on init, run, debug...

**If applicable please specify:**
 - CLI version: 0.5.8

kylegc:
This is happening while cobra is initializing the command flag, before we even know which command is being run.

PR #936, opened 2020-03-03
flags: remove logs from initialising stack registry flag

skoh7645:
When running an appsody command with verbose, cobra creates flags before it knows which command to run, this leads to many repeated logs. The only way (I know) to avoid this, is to avoid creating logs inside the flag initialisation and to bring them outside the `addStackRegistryFlag()` function

Fixes: #882

### DIFF
--- a/cmd/dev_common.go
+++ b/cmd/dev_common.go
@@ -20,12 +20,12 @@ import (
 	"os/exec"
 	"os/signal"
 	"os/user"
+	"path/filepath"
 	"regexp"
 	"runtime"
 	"strings"
 	"sync"
 	"syscall"
-	"path/filepath"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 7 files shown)


cmd/dev_common.go
```diff
@@ -80,7 +81,11 @@ func addStackRegistryFlag(cmd *cobra.Command, flagVar *string, config *RootComma
 	defaultRegistry := getDefaultStackRegistry(config)
 	stackRegistryInConfigFile, err := getStackRegistryFromConfigFile(config)
 	if err != nil {
-		config.Debug.Logf("Error retrieving the stack registry from config file: %v", err)
+		if _, ok := err.(*NotAnAppsodyProject); ok {
```
EnriqueL8: We could rechange this to only have an if

skoh7645: I've changed this now. I'm not too sure about the boolean idea, I can give it a try later if we decide to go with that

PR #939, opened 2020-03-04
Run with gid=0 when APPSODY_USER_RUN_AS_LOCAL=true

groeges:
Signed-off-by: Steven Groeger <groeges@uk.ibm.com>


groeges:
@neeraj-laad Do we still want to include this PR or shall I close it, in view of the latest announcements re Appsody. If we want to merge it I will update the PR so it is not out of date with master branch.

PR #944, opened 2020-03-09
Tests for appsody stack validate

### LINKED ISSUE #747: Code coverage analysis for stack_validate.go
- [ ] Test stack validate on a valid stack with no parameters
- [ ] Test stack validate on an invalid stack
- [ ] Test stack validate on a packaged stack using the --no-package flag
- [ ] Test stack validate on a stack with the --no-lint flag
- [ ] Test stack validate on a stack with a custom --image-namespace
- [ ] if !check after check, err := Exists("templates")
       Test stack validate on a stack with a missing templates folder

I believe tests 1 and 2 will cover all the cases for errors

upLukeWeston:
Fixes #747 

* Altered `RunAppsodyCmdExec` to no longer rely on `executable, _ := os.Executable()` for finding the root of the commands execution, because when the function was being called from tests, it was returning the `/usr/local/bin/go` executable rather than `appsody` - instead it now mimics the `RunAppsody` function is `testutils.go`.

* Added tests for `appsody stack validate`, which assert if the `--no-lint`, and `--no-package` flags work as intended.

    Both tests will currently use the `dev.local/appsodyvalidating-starter` docker image, which means only one can be parallel without causing errors. In order to change this, I'd have to rework the `stack validate` command itself

### DIFF
--- a/cmd/stack_validate_test.go
+++ b/cmd/stack_validate_test.go
@@ -1,0 +1,58 @@
+// Copyright © 2019 IBM Corporation and others.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package cmd_test
+
+import (
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/appsody/appsody/cmd/cmdtest"
+)
+
+func TestStackValidateNoLintFlag(t *testing.T) {
+	sandbox, cleanup := cmdtest.TestSetupWithSandbox(t, true)
+	defer cleanup()
+	sandbox.ProjectDir = filepath.Join(sandbox.TestDataPath, "starter")
+
+	args := []string{"stack", "validate", "--no-lint"}
+	output, err := cmdtest.RunAppsody(sandbox, args...)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


EnriqueL8:
@upLukeWeston seems like the tests are failing in travis - will wait for that to pass before reviewing

EnriqueL8:
Need to resolve conflicts

skoh7645:
@upLukeWeston the tests are failing on macOS and Windows because they need docker- should move this test file under `/functest`. `TestStackValidateNoPackageFlag` is failing because the stack needs to be packaged at least once before using the `--no-package` flag. `--no-package` is meant to mean "don't repackage this stack again before running validatation tests"

cmd/stack_validate_test.go
```diff
@@ -0,0 +1,69 @@
...
+	args := []string{"stack", "validate", "--no-lint"}
+	output, err := cmdtest.RunAppsody(sandbox, args...)
+	if err != nil {
+		t.Fatalf("Stack validate failed unexpectedly: %v", err)
+	}
+	if !strings.Contains(output, "Total PASSED: 5") && !strings.Contains(output, "Total FAILED: 0") {
+		t.Error("Stack validate did not have expected PASS/FAIL amounts")
+	}
+	if strings.Contains(output, "PASSED: Lint for stack") {
+		t.Error("Stack validate --no-package still ran packaging process")
```
CameronMcWilliam, line 38: Stack validate --no-lint still ran lint process*

cmd/stack_validate_utils.go
```diff
@@ -16,78 +16,52 @@ package cmd
...
-	executable, _ := os.Executable()
+	rootConfig.ProjectDir = workingDir
 
-	cmdArgs := []string{executable}
-	if rootConfig.Verbose {
-		cmdArgs = append(cmdArgs, "-v")
-	}
-	cmdArgs = append(cmdArgs, args...)
-	rootConfig.LoggingConfig.Info.log(cmdArgs)
+	// // Buffer cmd output, to be logged if there is a failure
```
CameronMcWilliam, line 35: Double `//`

Issue #699, opened 2019-11-25
Code coverage analysis for build.go


kylegc:
- [ ] func checkDockerBuildOptions & if buildOptions != ""
    test the runs build with --docker-options
    at least one with good options and one with bad options

- [ ] extractErr := extract(extractConfig)
    test an error on extract when calling build
    verify the error message from extract is printed

- [ ]  if config.pushURL != ""
    test build with --push-url that is invalid and verify the error

- [ ]  if execError != nil after DockerBuild
    simulate docker build error

PR #950, opened 2020-03-11
Tests for appsody build

upLukeWeston:
Resolves potions of #699.

- Test with an invalid `tag` associated with `push-url`
- Test with an invalid `push-url`
- Test for checking that `app-deploy-yaml` already exists before writing a new one


functest/build_test.go
```diff
@@ -413,36 +413,78 @@ func verifyImageAndConfigLabelsMatch(t *testing.T, deploymentManifest cmd.Deploy
...
+				// If an error is not returned, the test should fail
+			} else {
+				t.Error("Build with missing tag did not fail as expected")
+			}
+		})
+	}
 }
 
-func TestBuildMissingTagFail(t *testing.T) {
+func TestBuildConfigAlreadyExists(t *testing.T) {
```
EnriqueL8: Would just renamed this to `TestBuildDeploymentConfigAlreadyExists`

functest/build_test.go
```diff
@@ -413,36 +413,78 @@ func verifyImageAndConfigLabelsMatch(t *testing.T, deploymentManifest cmd.Deploy
...
+func TestInvalidBuild(t *testing.T) {
+
+	var knativeFlagTests = []struct {
+		testName    string
+		args        []string
+		expectedLog string
+	}{
+		{"Missing Tag ", []string{"--push"}, "Cannot specify --push or --push-url without a --tag"},
+		{"Invalid Tag with Push URL", []string{"--push-url", "i.am.not.a.real.url", "--tag", "£"}, "invalid argument \"i.am.not.a.real.url/£\" for \"-t, --tag"},
+		{"Invalid Push URL", []string{"--push-url", "i.am.not.a.real.url", "--tag", "notgonna/work"}, "Could not push the docker image"},
```
EnriqueL8: Maybe the message here should be invalid push url? could be difficult to verify that the url is valid.

upLukeWeston: I'm not sure which message you're referring to? The message is already 'invalid push url'?

EnriqueL8: I mean the expect log being `Could not push the docker image`

cmd/build.go
```diff
@@ -155,19 +155,21 @@ func build(config *buildCommandConfig) error {
...
+		// check to see if the pushURL provided is in the expected format
+		// Valid examples would be:
+		// https://www.appsody.co.uk
+		// http://www.appsody.co.uk
+		// www.appsody.co.uk
+		// appsody.co.uk
+		matched, err := regexp.MatchString("^(http:\\/\\/www\\.|https:\\/\\/www\\.|http:\\/\\/|https:\\/\\/)?[a-z0-9]+([\\-\\.]{1}[a-z0-9]+)*\\.[a-z]{2,5}(:[0-9]{1,5})?(\\/.*)?$", config.pushURL)
+		if err != nil {
+			return err
+		}
```
EnriqueL8: We would need to put this function in utils, and write specific tests for it.I tested it against `default-route-openshift-image-registry.apps-crc.testing` which should match but it didn't. I think maybe we can put this change in a different PR and merge the build tests for now?

upLukeWeston: I was thinking about this yeah, but wanted to keep it separate from the `build_test` stuff.  I'll extrapolate it into a function and write some test cases for it.

skoh7645:
Following @EnriqueL8's review, this LGTM

PR #952, opened 2020-03-12
Do not warn if there are no custom stack variables

sam-github:
Custom stack variable are optional, and are unused except for the java
stacks (ATM) but every stack developer will see this inscrutable message
when linting or packaging their stack. Also, note that the message
doesn't use "custom stack variables" (from the docs) or
"templating-data" (the stack yaml key name), making it a bit hard to
understand what it refers to, so clarify the format warning.

- https://appsody.dev/docs/stacks/develop#custom-stack-variables


sam-github:
The tests don't complete locally for me (ubuntu x64), not sure why, so will have to wait for Travis.

Kamran64:
@sam-github - Thanks for raising this. The purpose of the linter is to also provide suggestions that a creator can add to enhance their stack. I agree, a warning log is a bit strong here as it makes it seem like the stack creator is doing something wrong. Perhaps we can change it to an Info.log?

Kamran64:
Also the travis build is failing because there are tests that run against a stack with invalid templating values, which expect a particular log message for an error. Since this PR changes the error log, the expected log is different from the actual, which means the test fails.

sam-github:
Ok, added info log:
```
appsody stack lint   
LINTING nodejs-express
Linting Dockerfile-stack: /home/sam/w/appsody/stacks/incubator/nodejs-express/image/Dockerfile-stack
LINTING stack.yaml: /home/sam/w/appsody/stacks/incubator/nodejs-express/stack.yaml
No custom stack variables used (see https://appsody.dev/docs/stacks/develop/#custom-stack-variables)
TOTAL ERRORS: 0
TOTAL WARNINGS: 0
LINT TEST PASSED
```

PR #959, opened 2020-03-18
Retrieve git labels via commit info

Kamran64:
Previously, the git labels relied on a remote upstream being set, so for example, in a tekton pipeline where there is no upstream, not all labels could be extracted properly. More info in https://github.com/appsody/appsody-buildah/issues/16. This PR retrieves this info via the latest commit instead.


cmd/git_utils.go
```diff
@@ -99,31 +103,67 @@ func GetGitInfo(config *RootCommandConfig) (GitInfo, error) {
...
+	kargs := []string{"status", "-sb"}
+	statusOutput, gitErr := RunGit(config.LoggingConfig, config.ProjectDir, kargs, config.Dryrun)
+	if gitErr != nil {
+		return gitInfo, gitErr
+	}
+
+	statusOutput = strings.Trim(statusOutput, trimChars)
+	statusOutputLines := strings.Split(statusOutput, lineSeparator)
+
+	value := strings.Trim(statusOutputLines[0], trimChars)
```
EnriqueL8: I would still run `git status -sb` first, and then go to `branch contains` and check that the `origin/master` appears because it could be a remote which is not `origin` or `upstream` or the 1st.

cmd/git_utils.go
```diff
@@ -99,31 +103,67 @@ func GetGitInfo(config *RootCommandConfig) (GitInfo, error) {
...
+
+	if gitInfo.Upstream == "" {
+		gitInfo.Commit.Pushed = false
+		gitRemoteOutput, err := RunGitRemote(config.LoggingConfig, config.ProjectDir, lineSeparator, config.Dryrun)
+		if err != nil {
+			return gitInfo, err
+		}
+
+		for _, remotes := range gitRemoteOutput {
+			if strings.Contains(remotes, "origin") {
```
EnriqueL8: we might want to be more strict with this condition, maybe equals. Also would rename `remotes` to `remote`.

cmd/git_utils.go
Kamran64: Okay, so we should still keep the original method of getting the upstream from `git status -sb` and only resorting to the other methods if we can't retrieve it from there?

EnriqueL8: Yup, since for now we are not adding the commit url

Kamran64: Cool, added in the latest commit :)

cmd/git_utils.go
```diff
@@ -129,19 +137,55 @@ func GetGitInfo(config *RootCommandConfig) (GitInfo, error) {
...
 		changesMade = true
-
 	}
 	gitInfo.ChangesMade = changesMade
 
-	errMsg := ""
+	if gitInfo.Upstream == "" {
+		config.Debug.log("Failed to retrieve upstream from git status -sb. Looking via latest commit")
+		outputLines, err := RunGitBranchContains(config.LoggingConfig, gitInfo.Commit.SHA, config.ProjectDir, lineSeparator, config.Dryrun)
+		if err != nil {
```
EnriqueL8: What happens when the command doesn't return an error, and doesn't return any remotes? This would mean that `ouputLines` is empty, and `ouputLines[0]` could fail? Maybe we can do this check for empty within the `RunGitBranchContains` function, and return an error.
Regarding `errMsg` , if this fails we add a message. Then we do a `git remote` and if that is succesful we get the url, but the `errMsg` is not empty so [Line 199](https://github.com/appsody/appsody/pull/959/files#diff-5675d25a846ed226e9ba4514bd7eef60R199) will still return an error for the whole `GetGitInfo` function.

cmd/git_utils.go
```diff
@@ -129,19 +137,55 @@ func GetGitInfo(config *RootCommandConfig) (GitInfo, error) {
...
+					gitInfo.Upstream = upstream
+				}
+			}
+			gitInfo.Upstream = strings.TrimSpace(gitInfo.Upstream)
+		}
+	}
+
+	if gitInfo.Upstream == "" {
+		gitInfo.Commit.Pushed = false
+		gitRemoteOutput, err := RunGitRemote(config.LoggingConfig, config.ProjectDir, lineSeparator, config.Dryrun)
```
EnriqueL8: I would also add a check in `RunGitRemote` to add an error if it's empty, not very common but could happen if the user did a `git init` and didn't add a remote manually.

cmd/git_utils_test.go
```diff
@@ -117,27 +117,39 @@ func TestGetGitInfoWithNotAGitRepo(t *testing.T) {
...
-	expectedError := "does not have any commits yet"
-	if err == nil || !strings.Contains(err.Error(), expectedError) {
-		t.Errorf("Should had flagged error: %v", expectedError)
-	}
+			_, gitErr := cmd.RunGit(loggingConfig, sandbox.ProjectDir, []string{"init"}, false)
+			if gitErr != nil {
+				t.Error(gitErr)
+			}
+
+			_, err := cmd.GetGitInfo(config)
```
EnriqueL8: Here we haven't added a remote, so we could add a test to add a remote and then check that information is being extracted.

cmd/git_utils.go
Kamran64: I've added a bool value, `noRemoteFound` that is set to true if no remote has been found and false if one has. Then at the end of the function, a message is appended to `errMsg` if `noRemoteFound` is still true.

cmd/git_utils.go
```diff
@@ -129,20 +139,59 @@ func GetGitInfo(config *RootCommandConfig) (GitInfo, error) {
...
 	if outputLength > 1 {
 		changesMade = true
-
 	}
 	gitInfo.ChangesMade = changesMade
 
-	errMsg := ""
+	if gitInfo.Upstream == "" {
+		config.Debug.log("Failed to retrieve upstream from git status -sb. Looking via latest commit")
+		outputLines, err := RunGitBranchContains(config.LoggingConfig, gitInfo.Commit.SHA, config.ProjectDir, lineSeparator, config.Dryrun)
```
EnriqueL8: After running `git status -sb`, we only know that this branch is tracking an remote upstream branch. We do not know that the commit at the head of the local branch is in fact in the remote branch. I suggest that we always run `git branch contains`, and in the case where the upstream has already been set we verify that it is outputted by the command. If the upstream hasn't been set by `git status -sb`, we continue as per https://github.com/appsody/appsody/pull/959/files#diff-5675d25a846ed226e9ba4514bd7eef60R161-R168

cmd/git_utils.go
```diff
@@ -99,25 +108,26 @@ func GetGitInfo(config *RootCommandConfig) (GitInfo, error) {
 
 	config.Debug.log("git version: ", version)
 
-	kargs := []string{"status", "-sb"}
-
-	output, gitErr := RunGit(config.LoggingConfig, config.ProjectDir, kargs, config.Dryrun)
+	gitInfo.Commit, gitErr = RunGitGetLastCommit(gitInfo.RemoteURL, config)
```
EnriqueL8: Add setURL function 
```
func (commitInfo *CommitInfo) setURL(URL string) {
	if URL != "" {
		commitInfo.URL = StringBefore(URL, ".git") + "/commit/" + commitInfo.SHA
	}
}
```

Issue #509, opened 2019-10-23
"appsody run" should allow user-specified volume mounts


nastacio:
**Is your feature request related to a problem? Please describe.**
Service instances in Kubernetes are invariably mapped to [Kubernetes secrets](https://kubernetes.io/docs/concepts/configuration/secret/) , which are made available to applications in different ways, such as environment variables and volume mounts.

`appsody run` can accept environment variables with `--docker-options=--env-file...` or  `--docker-options=-e...` , but there is no way to hand it files containing secrets pulled out of a container.

**Describe the solution you'd like**
I don't want to prescribe the final design, but at the most basic level, I would like for `appsody run` to stop blocking "-v" mounts inside `--docker-options`.

I would welcome a consistency check between secret references in app-deploy.yaml and options specified for `appsody run`, such as:

```
WARNING: Your deployment spec has an env var named "ABC" and your run did not specify "ABC".
```

**Describe alternatives you've considered**
There are no alternatives. If an application needs to access a secret during runtime, the lack of user-specified mounts dictates the usage of secrets as environment variables during development, and therefore, virtually dictates the same approach when the application is running in production.

**Additional context**
Incidentally, the Appsody Application Operator allows the usage of secrets as both environment variables and as volume mounts when the application runs in a cluster.

kylegc:
Appsody depends on a number of mounts for itself and the stack to work properly. In order to allow user provided mounts, we would have to ensure they do not conflict with the other mounts appsody depends on. We will investigate this further.

GregDritschler:
@kylegc @nastacio I would like to second the motion to allow user-specified mounts.  I tried to use the git-commit-id-plugin maven plugin to capture git repo information in my build.  It does not work because the .git folder in the project is not mounted into the container.  From what I understand this is a popular maven plugin so I imagine others will eventually have this issue.

https://github.com/git-commit-id/maven-git-commit-id-plugin

Of course the problem also could be solved by just having `appsody run` and `appsody build` automatically mount the .git folder.

kylegc:
@GregDritschler thanks for the info. Would you please open a new issue for this? I think this is something that should be addressed by the stack itself. Make sure to specify which stack you are using. Thanks

GregDritschler:
I did some more investigation and found out this problem is specific to java-microprofile.  I opened https://github.com/appsody/stacks/issues/504.

nastacio:
[...]

PR #962, opened 2020-03-19
[WIP] dev_common.go: allow user specified mounts

skoh7645:
Allow user specified mounts in `appsody <run|debug|test> --docker-options`, but throw an error if the user mount interferes with the stack mounts
This should be rebased on top of #946, once it goes into master.

Fixes: #509

### DIFF
--- a/cmd/dev_common.go
+++ b/cmd/dev_common.go
@@ -44,20 +44,30 @@ type devCommonConfig struct {
 	dockerOptions   string
 }
 
-func checkDockerRunOptions(options []string) error {
-	//runOptionsTest := "(^((-p)|(--publish)|(--publish-all)|(-P)|(-u)|(--user)|(--name)|(--network)|(-t)|(--tty)|(--rm)|(--entrypoint)|(-v)|(--volume)|(-e)|(--env))((=?$)|(=.*)))"
-	runOptionsTest := "(^((--help)|(-p)|(--publish)|(--publish-all)|(-P)|(-u)|(--user)|(--name)|(--network)|(-t)|(--tty)|(--rm)|(--entrypoint)|(-v)|(--volume))((=?$)|(=.*)))"
-
+func checkDockerRunOptions(options []string, config *RootCommandConfig) error {
+	runOptionsTest := "(^((--help)|(-p)|(--publish)|(--publish-all)|(-P)|(-u)|(--user)|(--name)|(--network)|(-t)|(--tty)|(--rm)|(--entrypoint))((=?$)|(=.*)))"
 	blackListedRunOptionsRegexp := regexp.MustCompile(runOptionsTest)
-	for _, value := range options {
+
+	for ind, value := range options {
 		isInBlackListed := blackListedRunOptionsRegexp.MatchString(value)
 		if isInBlackListed {
 			return errors.Errorf("%s is not allowed in --docker-options", value)
-
+		}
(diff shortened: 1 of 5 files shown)


cmd/dev_common.go
```diff
@@ -45,20 +45,30 @@ type devCommonConfig struct {
...
 	blackListedRunOptionsRegexp := regexp.MustCompile(runOptionsTest)
-	for _, value := range options {
+
+	for ind, value := range options {
 		isInBlackListed := blackListedRunOptionsRegexp.MatchString(value)
 		if isInBlackListed {
 			return errors.Errorf("%s is not allowed in --docker-options", value)
-
+		}
+		if value == "-v" || value == "--volume" {
```
Kamran64: Could we be allowing the `--mount` flag also? In the [docker docs](https://docs.docker.com/storage/volumes/#choose-the--v-or---mount-flag#choose-the--v-or---mount-flag) it sounds like a more powerful alternative. Although I suppose since `-v` is what we use everywhere this is fine.

groeges:
Reviewed and found 2 issues
1) Help text for run|test|debug still states that `-v | --volume` are not supported
2) The check for duplicate volume paths is too prescriptive. It fails if volume.path is contained in the user volume path, so having a volume path of /test/mydir and a user volume path or /test/mydir2 fails as  the volume path is contained within the user volume path. The check needs to ensure the paths are the same.

cmd/dev_common.go
```diff
@@ -45,20 +45,30 @@ type devCommonConfig struct {
...
+			project, _, err := p.ensureProjectIDAndEntryExists(config)
+			if err != nil {
+				return err
+			}
+			for _, volume := range project.Volumes {
+				userSpecifiedMount := options[ind+1]
+				if strings.Contains(userSpecifiedMount, volume.Path) {
+					return errors.Errorf("User specified mount %s is not allowed in --docker-options, as it interferes with the stack specified mount %s", userSpecifiedMount, volume.Path)
+				}
+			}
```
groeges: The check for strings.Contains is too prescriptive and causes valid user specified mounts to fail. 
How about something like this:
```
			for _, volume := range project.Volumes {
				userSpecifiedMount := options[ind+1]
				userSpecifiedMountSplit := strings.Split(userSpecifiedMount, ":")
				if len(userSpecifiedMountSplit) != 2 {
					return errors.Errorf("User specified mount %s is not in the correct format.", userSpecifiedMount)
				}
				if userSpecifiedMountSplit[1] == volume.Path {
					return errors.Errorf("User specified mount %s is not allowed in --docker-options, as it interferes with the stack specified mount %s", userSpecifiedMount, volume.Path)
				}
			}
```

cmd/dev_common.go
EnriqueL8: Good idea, but `--mount` has a lot more features and will need more robust parsing.. 
```
 --mount 'type=volume,src=<VOLUME-NAME>,dst=<CONTAINER-PATH>,volume-driver=local,volume-opt=type=nfs,volume-opt=device=<nfs-server>:<nfs-path>,"volume-opt=o=addr=<nfs-address>,vers=4,soft,timeo=180,bg,tcp,rw"'
```
I would hold off for now

cmd/dev_common.go
groeges: Latest changes look OK

PR #963, opened 2020-03-19
Reduce duplication in List command

EnriqueL8:
Worked with @upLukeWeston to clean up the list command.


cmd/repo.go
```diff
@@ -578,20 +584,13 @@ func (r *RepositoryFile) getRepositories(log *LoggingConfig) (IndexOutputFormat,
 	return indexOutput, nil
 }
 
-func (r *RepositoryFile) getRepository(log *LoggingConfig, repoName string) (IndexOutputFormat, error) {
+func (r *RepositoryFile) getRepository(config *RootCommandConfig, index *IndexYaml, repoName string) (IndexOutputFormat, error) {
```
upLukeWeston: Does the config still need to be passed in here?

upLukeWeston (dismissed):
Looks good to go in, just the one comment which could be changed, but it's not a major point.

cmd/repo.go
EnriqueL8: Good catch, will update now

upLukeWeston (approved):
Ready to update and merge!

Issue #957, opened 2020-03-17
Add stack image digest as a label on application image


neeraj-laad:
Appsody adds traceability labels to indicate which stack version is used to build the stack. However, we use the tag as a way to uniquely identify the stack image
```
"dev.appsody.stack.tag": "docker.io/appsody/nodejs-express:0.4.5"
```

We would like to see the actual image digest for the image added as an additional label. I did not find an OCI spec label for this, so suggesting we use: 

```
"dev.appsody.stack.digest": "<image digest>"
```

EnriqueL8:
Reopening issue as we are reverting this change. 

Buildah path not working yet. See: https://github.com/containers/buildah/issues/2074

PR #969, opened 2020-03-27
Add image digest as a label

Kamran64:
Adds the digest of the image as a label (only if the image actually has one)

Fixes: https://github.com/appsody/appsody/issues/957

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -1074,8 +1074,7 @@ func getStackLabels(config *RootCommandConfig) (map[string]string, error) {
 
 	imageAndDigest := data[0]["RepoDigests"].([]interface{})
 	if len(imageAndDigest) > 0 { //Check that the image has a digest
-		imageAndDigestStr := fmt.Sprintf("%v", imageAndDigest[0])
-		digest := strings.Split(imageAndDigestStr, "@")
+		digest := strings.Split(imageAndDigest[0].(string), ":")
 		labels[appsodyStackKeyPrefix+"digest"] = digest[1]
 	}
 
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -1072,12 +1072,10 @@ func getStackLabels(config *RootCommandConfig) (map[string]string, error) {
 		}
 	}
 
-	imageAndDigest := data[0]["RepoDigests"].([]interface{})
-	if len(imageAndDigest) > 0 { //Check that the image has a digest
-		imageAndDigestStr := fmt.Sprintf("%v", imageAndDigest[0])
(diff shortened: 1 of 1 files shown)


cmd/utils.go
```diff
@@ -1060,6 +1060,13 @@ func getStackLabels(config *RootCommandConfig) (map[string]string, error) {
 		}
 	}
 
+	imageAndDigest := data[0]["RepoDigests"].([]interface{})
```
EnriqueL8: I think we should be looking at the `id` field given my the inspect command. I build a stack by running a `stack package` and my `RepoDigests` was empty.

Kamran64: I suppose this makes more sense if we're looking at uniquely identifying the images since locally built images won't generate a digest. Added in the latest commit.

Kamran64: @EnriqueL8 I've just reverted that commit to add the image id instead of the digest. Spoke with @neeraj-laad and we need the digest. Since locally built images don't generate one we'll just skip adding the label for those cases.

EnriqueL8: Makes sense, I would only add the `RepoDigests` then.

cmd/utils.go
```diff
@@ -1072,6 +1072,13 @@ func getStackLabels(config *RootCommandConfig) (map[string]string, error) {
...
+	if len(imageAndDigest) > 0 { //Check that the image has a digest
+		imageAndDigestStr := fmt.Sprintf("%v", imageAndDigest[0])
+		digest := strings.Split(imageAndDigestStr, "@")
+		labels[appsodyStackKeyPrefix+"digest"] = digest[1]
```
EnriqueL8: Do we want the `sha256:` prefix in the digest?

cmd/utils.go
```diff
@@ -1072,6 +1072,13 @@ func getStackLabels(config *RootCommandConfig) (map[string]string, error) {
 		}
 	}
 
+	imageAndDigest := data[0]["RepoDigests"].([]interface{})
+	if len(imageAndDigest) > 0 { //Check that the image has a digest
+		imageAndDigestStr := fmt.Sprintf("%v", imageAndDigest[0])
```
EnriqueL8: since here you are not formatting, youn can use `string(imageAndDigest[0])` within `strings.Split()`

cmd/utils.go
EnriqueL8: @Kamran64 apparently we need to add the `sha256:` prefix.. sorry!!

Kamran64: @EnriqueL8 Np, readded in the latest commit

PR #977, opened 2020-04-06
Add label for image digest

Kamran64:
Adds a label for the digest of the image being pulled down.

Fixes: https://github.com/appsody/appsody/issues/957

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -1094,7 +1094,7 @@ func getStackLabels(config *RootCommandConfig) (map[string]string, error) {
 	}
 
 	if config.Buildah {
-		buildahDigest, err := getBuildahDigest(labels["dev.appsody.stack.tag"], config)
+		buildahDigest, err := getBuildahDigest(labels["dev.appsody.stack.id"], imageName, config)
 		if err != nil || buildahDigest == "" {
 			config.Warning.log(err)
 			return labels, nil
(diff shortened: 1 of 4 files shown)


cmd/utils.go
```diff
@@ -1082,9 +1093,51 @@ func getStackLabels(config *RootCommandConfig) (map[string]string, error) {
...
+			return labels, nil
+		}
+		labels[appsodyStackKeyPrefix+"digest"] = buildahDigest
+		config.Debug.log("Digest label successfully added")
+	}
+
 	return labels, nil
 }
 
+func getBuildahDigest(idKey string, image string, config *RootCommandConfig) (digest string, err error) {
```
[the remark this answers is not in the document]
Kamran64: Thanks, I did think about this but although there's more safeguarding with this method, it also compromises on "thoroughness" i.e. it's less reliable to get the correct digest than the Docker method. Unfortunately the format of the `buildah inspect` command changed slightly a couple versions back for the digest meaning we have to resort to this less simple method.

cmd/utils.go
```diff
@@ -1082,9 +1093,51 @@ func getStackLabels(config *RootCommandConfig) (map[string]string, error) {
...
+	splitImage := strings.Split(image, ":") //Split the image to extract the image name and version separately
+	if len(splitImage) != 2 {
+		return "", errors.New("Error retrieving image name and tag used for build. The image digest will not be added as a label")
+	}
+	imageName := splitImage[0]    //The image name minus the version
+	imageVersion := splitImage[1] //The version of the image used
+	cmdName := "buildah"
```
EnriqueL8, line 1116: Can you add a comment here specifying the limation of the buildah version and why we have to use this command instead?

cmd/utils.go
```diff
@@ -1082,9 +1093,51 @@ func getStackLabels(config *RootCommandConfig) (map[string]string, error) {
...
 	return labels, nil
 }
 
+func getBuildahDigest(idKey string, image string, config *RootCommandConfig) (digest string, err error) {
+	splitImage := strings.Split(image, ":") //Split the image to extract the image name and version separately
+	if len(splitImage) != 2 {
+		return "", errors.New("Error retrieving image name and tag used for build. The image digest will not be added as a label")
+	}
+	imageName := splitImage[0]    //The image name minus the version
+	imageVersion := splitImage[1] //The version of the image used
```
EnriqueL8, line 1115: I would rename this to `imageTag` because the convention for registries is tag.

functest/build_test.go
```diff
@@ -229,6 +230,57 @@ func deleteImage(imageName string, cmdName string, t *testing.T) {
...
+
+	var containerConfig map[string]interface{}
+	var buildahData map[string]interface{}
+	err = json.Unmarshal([]byte(inspectOutput), &buildahData)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	containerConfig = buildahData["config"].(map[string]interface{})
+	labelsMap := containerConfig["Labels"].(map[string]interface{})
```
EnriqueL8, line 276: You might need to check that these are not nil

PR #979, opened 2020-04-09
Added no-sever field to stack.yaml

CameronMcWilliam:
Added a `no-server` field to the `stack.yaml` that allows stack creators to specify if their does not provide a server.

Fixes issue [602](https://github.com/appsody/stacks/issues/602).


skoh7645:
We would need to add the `no-server` field to the json struct and `generateCodewindIndex()`, so it appears in the `json` index that we generate

PR #982, opened 2020-04-14
Skip verification of server certificates

groeges:
Signed-off-by: Steven Groeger <groeges@uk.ibm.com>

This PR is the first step in supporting accessing secure endpoints. 

This change will allow the index.yaml and templates to be retrieved from a secured endpoint.

The change doesn't support accessing docker images that are in a secured private registry. 
In order to access the secure private registry the use may need to do a `docker login` and / or copy yhe certificates into their local certificate store so that docker can use it.

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -18,6 +18,7 @@ import (
 	"bufio"
 	"bytes"
 	"compress/gzip"
+	"crypto/tls"
 	"unicode"
 
 	"encoding/json"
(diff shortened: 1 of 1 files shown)


groeges:
@neeraj-laad Do we still want to include this PR or shall I close it, in view of the latest announcements re Appsody.

EnriqueL8:
Closing for now

PR #990, opened 2020-04-23
Show controller version when appsody version is called

CameronMcWilliam:
Using the command `appsody version` will now show both the CLI and controller version for Appsody.  It will also check if the version has been overrides with the two env vars `APPSODY_CONTROLLER_VERSION` and `APPSODY_CONTROLLER_IMAGE` and display the appropriate info.

Also added some unit tests to cover these new capabilities.


CameronMcWilliam:
> Looks good but the help text should be updated to say that this command now also outputs the controller version.

Forgot about that!

cmd/version_test.go
```diff
@@ -34,8 +35,8 @@ func TestVersion(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	if !strings.Contains(output, cmd.VERSION) {
-		t.Fatal("Output does not contain version")
+	if !strings.Contains(output, cmd.VERSION) && !strings.Contains(output, cmd.CONTROLLERVERSION) {
```
EnriqueL8, line 38: What are the values of `cmd.VERSION` and `cmd.CONTROLLERVERSION` here? Does `cmd.VERSION` default to `vlatest` and `cmd.CONTROLLERVERSION` to `latest`?

cmd/version_test.go
```diff
@@ -54,3 +55,39 @@ func TestVersionTooManyArgs(t *testing.T) {
...
+
+	os.Setenv("APPSODY_CONTROLLER_IMAGE", "appsody/new-controller:0.1.1")
+
+	args := []string{"version"}
+	output, err := cmdtest.RunAppsody(sandbox, args...)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if !strings.Contains(output, cmd.VERSION) && !strings.Contains(output, "appsody/new-controller:0.1.1") {
```
EnriqueL8, line 90: wouldn't we want `0.1.1`? Since we split by `:` and take the version?

CameronMcWilliam: Yeah, testing that because there's a `Log.Info.log` that prints out that it is using the new controller image.  I should test that we get `appsody-controller 0.1.1` too

Kamran64:
@EnriqueL8 - Can this be merged?

EnriqueL8:
Sorry missed this ping, we could merge this

Issue #993, opened 2020-04-30
Do not expose debug ports when running in k8s (in Codewind)


makandre:
**Describe the bug**

When an Appsody project is run from Codewind on k8s, the debug port is being exposed.  Take `nodejs-express` stack for example:

In `app-deploy.yaml`
```yaml
        ports:
        - containerPort: 9229
        - containerPort: 3000
```

In `app-service.yaml`
```yaml
  ports:
  - name: port-0
    port: 9229
    targetPort: 9229
  - name: port-1
    port: 3000
    targetPort: 3000
```

This is problematic for 2 reasons:

1. exposing debug port is a security issue
2. this is interfering with Codewind remote debug support; we are trying to do a kube port-forward to the port and we believe it is failing because the port is already "taken" by being already exposed

**To Reproduce**

**Expected behavior**

**Actual behaviour**
What is the actual behaviour.

**Environment Details (please complete the following information):**

**If applicable please specify:**
 - CLI version:
 - Stack you are using:

**Screenshots**

**Additional context**

Kamran64:
From some initial investigation, we need a way to determine the debug port for each stack if we want to do this. There was an issue raised a while ago: https://github.com/appsody/stacks/issues/261 to cover this. My initial thoughts were that:
- We could just set an environment variable in each stack (`APPSODY_DEBUG_PORT`) with this information but (to summarise the issue thread), what if a stack requires multiple debug ports? 

- The other option is to define this information inside the `stack.yaml`. Possibly, a `ports` field with other values nested beneath them e.g. `debugPort`. The CLI could then use this information from the index to not expose the port when generating the deployment yaml.

- Do all stacks that support `appsody debug` have a `launch.json`? If so, the port for the debugger is usually in that file. The CLI could read this file during `appsody run` if it exists but I'm not sure how reliable this is.

I'm struggling to think of a way for the CLI to determine this information without us updating all the stacks which is where I think most of the effort doing this will come from. If point 3 is valid, this could be a potential solution.

@makandre @neeraj-laad - What do you think?

makandre:
I don't know the pros/cons of the options you've described, but have a question.

`launch.json` file is in a template, not the stack image itself right?  If so, couldn't we have the debug port in the `.appsody-config.yaml` itself?  All templates have that I assume; it's also something Codewind could read if necessary w/o downloading the stack image (which may help solve https://github.com/appsody/stacks/issues/261 too)

Kamran64:
[...]

PR #994, opened 2020-05-07
Don't expose debug port when running in k8s

Kamran64:
Fixes: https://github.com/appsody/appsody/issues/993

To utilise this, the stack must have an `APPSODY_DEBUG_PORT` environment variable defined in the `Dockerfile-stack` e.g. for the `nodejs-express` stack `APPSODY_DEBUG_PORT=9229`.

To test the changes (since stacks don't currently have this defined)
- Clone appsody/stacks repo or run e.g. `appsody stack create nodejs-express --copy nodejs-express`
- Add `ENV APPSODY_DEBUG_PORT=9229` to the `Dockerfile-stack`
- Run `appsody stack package`
- `export APPSODY_K8S_EXPERIMENTAL=TRUE`
- `appsody init` the newly packaged stack and `appsody run` (Ensure you've built the CLI locally and are using that instead of your installed version of Appsody)

### DIFF
--- a/cmd/dev_common.go
+++ b/cmd/dev_common.go
@@ -379,14 +379,15 @@ func commonCmd(config *devCommonConfig, mode string) error {
 			return portsErr
 		}
 
-		var debugPortArray []string
-		debugPorts, debugPortErr := GetEnvVar("APPSODY_DEBUG_PORT", config.RootCommandConfig)
-		if debugPortErr != nil || debugPorts == "" {
-			config.Debug.log("No debug port found. Continuing...")
-		} else {
-			debugPortArray = strings.Split(debugPorts, ", ") //Split the string if multiple debug ports exist
-			for _, debugPort := range debugPortArray {
-				debugPortExists := InArray(portList, debugPort) //Determine whether all ports specified in env var have actually been exposed
+		var debugPort string
+		codeWindProjectID := os.Getenv("CODEWIND_PROJECT_ID")
+
+		if codeWindProjectID != "" && mode == "debug" {
+			debugPort, debugPortErr := GetEnvVar("APPSODY_DEBUG_PORT", config.RootCommandConfig)
+			if debugPortErr != nil || debugPort == "" {
+				config.Debug.log("No debug port found. Continuing...")
+			} else {
+				debugPortExists := InArray(portList, debugPort) //Determine whether port specified in env var has actually been exposed
 				if !debugPortExists {
 					return errors.Errorf("Port: %s specified in APPSODY_DEBUG_PORT could not be found in ports list", debugPort)
 				}
@@ -411,7 +412,7 @@ func commonCmd(config *devCommonConfig, mode string) error {
 			return err
 		}
 		config.Debug.Logf("Docker env vars extracted from docker options: %v", dockerEnvVars)
-		deploymentYaml, err := GenDeploymentYaml(config.LoggingConfig, config.containerName, platformDefinition, controllerImageName, portList, debugPortArray, projectDir, dockerMounts, dockerEnvVars, depsMount, dryrun)
+		deploymentYaml, err := GenDeploymentYaml(config.LoggingConfig, config.containerName, platformDefinition, controllerImageName, portList, debugPort, projectDir, dockerMounts, dockerEnvVars, depsMount, dryrun)
 		if err != nil {
 			return err
 		}
@@ -422,7 +423,7 @@ func commonCmd(config *devCommonConfig, mode string) error {
 		if err != nil {
 			return err
 		}
-		serviceYaml, err := GenServiceYaml(config.LoggingConfig, config.containerName, portList, debugPortArray, projectDir, dryrun)
+		serviceYaml, err := GenServiceYaml(config.LoggingConfig, config.containerName, portList, debugPort, projectDir, dryrun)
 		if err != nil {
 			return err
 		}
@@ -431,7 +432,6 @@ func commonCmd(config *devCommonConfig, mode string) error {
 		if err != nil {
 			return err
 		}
-		codeWindProjectID := os.Getenv("CODEWIND_PROJECT_ID")
 		if codeWindProjectID == "" {
 			port := getIngressPort(config.RootCommandConfig)
 			// Generate the Ingress only if it makes sense - i.e. there's a port to expose
--- a/cmd/dev_common.go
+++ b/cmd/dev_common.go
@@ -379,9 +379,12 @@ func commonCmd(config *devCommonConfig, mode string) error {
 			return portsErr
 		}
 
-		debugPort, debugPortErr := GetEnvVar("APPSODY_DEBUG_PORT", config.RootCommandConfig)
-		if debugPortErr != nil || debugPort == "" {
+		var debugPortArray []string
+		debugPorts, debugPortErr := GetEnvVar("APPSODY_DEBUG_PORT", config.RootCommandConfig)
+		if debugPortErr != nil || debugPorts == "" {
 			config.Debug.log("No debug port found. Continuing...")
+		} else {
+			debugPortArray = strings.Split(debugPorts, " ")
 		}
 
 		projectDir, err := getProjectDir(config.RootCommandConfig)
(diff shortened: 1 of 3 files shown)


makandre:
Getting an error with the service:

```
Running command: kubectl apply -f /codewind-workspace/rm-apnode01/app-service.yaml
[Error] kubectl apply failed: exit status 1: The Service "cw-rmapnode01-d0421ce0-90dc-11ea-afd7-service" is invalid: 
* spec.ports[0].name: Required value
* spec.ports[0].port: Invalid value: 0: must be between 1 and 65535, inclusive
* spec.ports[0].targetPort: Invalid value: 0: must be between 1 and 65535, inclusive
```

The generated `app-service.yaml` has this for the ports which is invalid:

```
ports:
  - port: 0
    targetPort: 0
  - name: port-1
    port: 3000
    targetPort: 3000
```

neeraj-laad:
@Kamran64 As per your proposal, will APPSODY_DEBUG_PORT only have one value or multiple values?

makandre:
@neeraj-laad Kamran said it will support [multiple ports](https://github.com/appsody/appsody/issues/993#issuecomment-624145615) but for my test I only listed one (`9229`)

cmd/dev_common.go
```diff
@@ -379,6 +379,11 @@ func commonCmd(config *devCommonConfig, mode string) error {
 			return portsErr
 		}
 
+		debugPort, debugPortErr := GetEnvVar("APPSODY_DEBUG_PORT", config.RootCommandConfig)
```
EnriqueL8, line 382: If we want to support multiple debug ports, we will need to split the value of the ENV here

cmd/utils.go
```diff
@@ -1306,17 +1306,21 @@ func GenDeploymentYaml(log *LoggingConfig, appName string, imageName string, con
...
-		//KNative only allows a single port entry
-		if i == 0 {
-			yamlMap.Spec.PodTemplate.Spec.Containers[0].Ports = containerPorts
-		}
-		log.Debug.Log("Adding port to yaml: ", port)
-		newContainerPort := new(Port)
-		newContainerPort.ContainerPort, err = strconv.Atoi(port)
-		if err != nil {
-			return "", err
+		if port == debugPort {
```
EnriqueL8, line 1309: And change this condition for multiple ports.

cmd/utils.go
```diff
@@ -1306,17 +1306,21 @@ func GenDeploymentYaml(log *LoggingConfig, appName string, imageName string, con
...
+			if i == 0 {
+				yamlMap.Spec.PodTemplate.Spec.Containers[0].Ports = containerPorts
+			}
+			log.Debug.Log("Adding port to yaml: ", port)
+			newContainerPort := new(Port)
+			newContainerPort.ContainerPort, err = strconv.Atoi(port)
+			if err != nil {
+				return "", err
+			}
+			yamlMap.Spec.PodTemplate.Spec.Containers[0].Ports = append(yamlMap.Spec.PodTemplate.Spec.Containers[0].Ports, newContainerPort)
```
EnriqueL8, line 1322: I believe this is currently failing if we have multiple debug ports, because this code will get executed for all but one of the multiple debug ports.

cmd/utils.go
```diff
@@ -1491,13 +1495,17 @@ func GenServiceYaml(log *LoggingConfig, appName string, ports []string, pdir str
...
+		if port == debugPort {
+			log.Debug.log("Debug port found. Skipping addition to the service file...")
+		} else {
+			service.Spec.Ports[i].Name = fmt.Sprintf("port-%d", i)
+			iPort, err := strconv.Atoi(port)
+			if err != nil {
+				return "", err
+			}
+			service.Spec.Ports[i].Port = iPort
+			service.Spec.Ports[i].TargetPort = iPort
```
EnriqueL8, line 1507: Same comments as above

cmd/dev_common.go
EnriqueL8: We will also need to sanitize and verify these values

cmd/utils.go
```diff
@@ -1491,13 +1495,17 @@ func GenServiceYaml(log *LoggingConfig, appName string, ports []string, pdir str
 	service.Spec.ServiceType = "NodePort"
 	service.Spec.Ports = make([]Port, len(ports))
```
EnriqueL8, line 1496: this line is creating an empty port field in the case where we have a debug port. The len(ports) will include the debugport, but during the for loop we will not set those values.

EnriqueL8, line 1496: This is adding a empty port field when have a debug port, because inside the for loop we will skiping that port and not setting values: 
```
- port: 0
  targetPort: 0
```
Instead only add to this list instead of setting a fixed size.

Kamran64:
Sorry for the late replies here, I was off last week for vacation:
@neeraj-laad - Yes, this PR is supposed to support multiple values for `APPSODY_DEBUG_PORT`. That has now been added as of the latest commit. (Will need to be defined like so: e.g. `APPSODY_DEBUG_PORT="9229 9080"`

@makandre - I've fixed the weird port: 0 errors you were facing following @EnriqueL8's comments. Sorry about that.

I just did a run through and tested several flows on this PR. One of which, I added `ENV APPSODY_DEBUG_PORT=9229` to the `Dockerfile-stack` of a `nodejs-express` stack. On `appsody run` (with k8s mode enabled), my `app-deploy.yaml` looked like this:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test
spec:
  selector:
    matchLabels:
      app: test
  replicas: 1
  template:
    metadata:
      labels:
        app: test
        release: test
    spec:
      serviceAccountName: appsody-sa
      initContainers:
      - image: appsody/init-controller:latest
        imagePullPolicy: IfNotPresent
        name: init-appsody-controller
        volumeMounts:
        - name: appsody-controller
          mountPath: /.appsody
      containers:
      - command:
        - /.appsody/appsody-controller
        image: dev.local/appsody/nodejs-express:0.4
        imagePullPolicy: Always
        name: test
        ports:
        - containerPort: 3000
        volumeMounts:
        - name: appsody-controller
          mountPath: /.appsody
        - name: appsody-workspace
          mountPath: /project/user-app
          subPath: root/go/src/github.com/appsody/appsody/test
      volumes:
      - name: appsody-controller
      - name: appsody-workspace
        persistentVolumeClaim:
          claimName: appsody-workspace
```

And my `app-service.yaml` looked like this:
```
apiVersion: v1
kind: Service
metadata:
  name: test-service
  labels:
    release: test
spec:
  selector:
    app: test
  type: NodePort
  ports:
  - name: port-0
    port: 3000
    targetPort: 3000
```
I believe this is what we want?

cmd/utils.go
```diff
@@ -1391,6 +1396,17 @@ func GenDeploymentYaml(log *LoggingConfig, appName string, imageName string, con
...
+//FindElement returns true if an element is in an array, and false otherwise
+func FindElement(slice []string, value string) bool {
+	for _, item := range slice {
+		if item == value {
+			return true
+		}
+	}
+	return false
+}
+
```
EnriqueL8, line 1409: I believe we have a similar function already in utils?

Kamran64: Thanks, yeah you're right. It was in the `testutils` file so I hadn't seen it. I moved it to the main utils so both files can use it. Also I agree on having it as a comma separated list, that's in the latest commit :)

makandre:
Trying the new commits an the app-deploy.yaml looks correct now:

```yaml
ports:
        - containerPort: 3000
```

But then I get:

```bash
Running command: kubectl apply -f /codewind-workspace/rm-apnode01/app-deploy.yaml
panic: runtime error: index out of range [1] with length 1

goroutine 1 [running]:
github.com/appsody/appsody/cmd.GenServiceYaml(0xc0000f0700, 0x7ffdc858f247, 0x25, 0xc00000c4e0, 0x2, 0x2, 0xc00005d6e0, 0x1, 0x1, 0xc000028574, ...)
	/Users/andrewmak/go/src/github.com/appsody/appsody/cmd/utils.go:1514 +0xdc3
github.com/appsody/appsody/cmd.commonCmd(0xc0000f6a80, 0xb6d451, 0x3, 0xc78080, 0xc0000b2008)
	/Users/andrewmak/go/src/github.com/appsody/appsody/cmd/dev_common.go:425 +0x17ab
```

Kamran64:
@makandre - I haven't managed to replicate that error but I'll take a closer look. Thanks.

Kamran64:
Following our discussions, this is the behaviour that the PR alters:
- `appsody run/test`: No change in docker and k8s scenario

- `appsody debug` = No change in docker. We no longer expose the debug port in Codewind (On the `APPSODY_K8S_EXPERIMENTAL` path and check for the existence of `CODEWIND_PROJECT_ID` so that we know we're in a Codewind environment. (remove debug ports from `app-deploy.yaml` and `app-service.yaml`) (k8s scenario)

This will only work if a stack has declared the `APPSODY_DEBUG_PORT` environment variable. Note, we only support one value for this variable now.

Kamran64:
Closing in favour of: https://github.com/appsody/appsody/pull/1000

Issue #996, opened 2020-05-07
The stack linter is unforgiving when a stack extends another stack


chilanti:
**Describe the bug**
If your stack extends another stack, the `Dockerfile-stack` file is unlikely to contain all the apposdy env vars. Yet the linter flags (and in some cases errors out) on some of the "missing" variables - such as `APPSODY_MOUNTS` and `APPSODY_RUN`.

**To Reproduce**
1. Go to https://github.com/IBM-Blockchain/appsody-stacks
2. Take a look at /incubator/blockchain-client-openliberty/image/Dockerfile-stack
3. Notice that it extends another stack: `FROM kabanero/java-openliberty:0.2`
4. Travis produces these errors:
```
LINTING blockchain-client-openliberty
[Warning] Missing Dockerfile in /home/travis/build/IBM-Blockchain/appsody-stacks/incubator/blockchain-client-openliberty/image/project
Linting Dockerfile-stack: /home/travis/build/IBM-Blockchain/appsody-stacks/incubator/blockchain-client-openliberty/image/Dockerfile-stack
[Error] Missing APPSODY_MOUNTS
[Error] Missing APPSODY_RUN
[Warning] Missing APPSODY_DEBUG
[Warning] Missing APPSODY_TEST
[Warning] Missing APPSODY_DEPS
[Warning] Missing APPSODY_PROJECT_DIR
[Error] APPSODY_WATCH_DIR is defined, but no ON_CHANGE variable is defined
[Error] No APPSODY MOUNTS exists, mount paths can not be validated.
LINTING stack.yaml: /home/travis/build/IBM-Blockchain/appsody-stacks/incubator/blockchain-client-openliberty/stack.yaml
TOTAL ERRORS: 4
TOTAL WARNINGS: 5
[Error] LINT TEST FAILED
```

**Expected behavior**
The Linter should be more lenient when Dockerfile-stack extends another stack, because those vars may be inherited. Produce warnings instead of errors?

**Actual behaviour**
Miserable failures of our builds.

Issue #995, opened 2020-05-11
The valid license check should not prevent stack package working


skoh7645:
Currently our valid license check (which is run as part of the `stack.yaml` linter process during `appsody stack package`) returns an error if the license field is not a valid [SPDX license](https://spdx.org/licenses/). 

This check should be more lenient, and return a warning (rather than an error) as an invalid license should not stop the stack from being packaged.

Example of this problem on slack: https://appsody.slack.com/archives/CKZ50HJET/p1589200010019400

PR #997, opened 2020-05-18
Make appsody stack lint more forgiving

Kamran64:
Makes `appsody stack lint` much more forgiving. It will no longer error out for missing variables in the `Dockerfile-stack`. This has caused problems for stacks that inherit from a parent failing due to missing variables. It also changes the license checker to a warning too.

Fixes: https://github.com/appsody/appsody/issues/995
Fixes: https://github.com/appsody/appsody/issues/996

### DIFF
--- a/cmd/stack_lint_test.go
+++ b/cmd/stack_lint_test.go
@@ -42,17 +42,14 @@ func TestLinterInvalidValues(t *testing.T) {
 		replaceLine  string
 		expectedLog  string
 	}{
-		{"Invalid Run Value", dockerfilePath, "APPSODY_RUN", "Testing", "Missing APPSODY_RUN"},
 		{"Invalid Kill Value", dockerfilePath, "_KILL", "ENV APPSODY_DEBUG_KILL=trued", "APPSODY_DEBUG_KILL can only have value true/false"},
 		{"Invalid Regex Value", dockerfilePath, "ENV APPSODY_WATCH_REGEX='^.*(.xml|.java|.properties)$'", "ENV APPSODY_WATCH_REGEX='['", "error parsing regexp: missing closing ]"},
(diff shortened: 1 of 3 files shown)


cmd/stack_yaml_lint.go
```diff
@@ -130,16 +132,16 @@ func (stackDetails *StackYaml) checkTemplatingData(log *LoggingConfig) (int, int
...
 
 	if err := checkValidLicense(log, stackDetails.License); err != nil {
-		stackLintErrorCount++
-		log.Error.logf("The stack.yaml SPDX license ID is invalid: %v.", err)
+		stackLintWarningCount++
+		log.Warning.logf("The stack.yaml SPDX license ID is invalid: %v.", err)
 	}
 	if valid, err := IsValidKubernetesLabelValue(stackDetails.License); !valid {
-		log.Error.logf("The stack.yaml SPDX license ID is invalid: %v.", err)
+		log.Warning.logf("The stack.yaml SPDX license ID is invalid: %v.", err)
```
EnriqueL8, line 142: do we need to increment the count here?

cmd/stack_lint_test.go
```diff
@@ -93,6 +90,59 @@ func TestLinterInvalidValues(t *testing.T) {
...
+			for i, line := range lines {
+				if strings.Contains(line, tt.containsLine) {
+					lines[i] = tt.replaceLine
+				}
+			}
+			output := strings.Join(lines, "\n")
+			err = ioutil.WriteFile(targetPath, []byte(output), 0644)
+			if err != nil {
+				t.Fatal(err)
+			}
```
EnriqueL8, line 134: I see a lot of common code with the other tests. This could be put into a function in the future

cmd/stack_yaml_lint.go
Kamran64: Good catch, done :)

Issue #561, opened 2019-11-01
Appsody on K8s: support debugging scenario


neeraj-laad:
**Is your feature request related to a problem? Please describe.**
Appsody on K8S is untested with regard to the ability to provide remote debugging and to run `appsody debug`.

**Describe the solution you'd like**
Test and enable the `appsody debug` scenarios.

neeraj-laad:
We have done initial work on this but this depends on how CodeWind plans to support debugging scenarios.

kylegc:
Work started but not merged in PR #624

PR #1000, opened 2020-05-21
Enable other appsody "modes" in appsody k8s and stop exposing the debug port for Codewind

Kamran64:
- Extended from https://github.com/appsody/appsody/pull/994 as it enables an `APPSODY_DEBUG_PORT` variable to detect debug ports.

- This PR allows `appsody debug` and `appsody test` to be run when using `APPSODY_K8S_EXPERIMENTAL`. Previously we weren't propagating the `mode` variable to this block and therefore the controller would always fall back to using the `appsody run` command.

- Note: For Codewind, they only actually need `appsody debug` at this time but due to the nature of the code, both `appsody test` and `appsody debug` will be supported since we're passing the arg to the controller. @makandre - Should we block invocations of `appsody test` when running in Codewind or do you think it will be fine to support both?

YAML structure in `app-deploy.yaml`:
```
        - name: appsody-controller
          mountPath: /.appsody
      containers:
      - args:
        - --mode
        - debug
        command:
        - /.appsody/appsody-controller
        image: docker.io/appsody/nodejs-express:0.4
        imagePullPolicy: Always
```

Fixes: https://github.com/appsody/appsody/issues/561

### DIFF
--- a/cmd/cmdtest/testutils.go
+++ b/cmd/cmdtest/testutils.go
@@ -27,7 +27,7 @@ import (
 	"sync"
 	"testing"
 
-	"github.com/appsody/appsody/cmd"
+	cmd "github.com/appsody/appsody/cmd"
 	"gopkg.in/yaml.v2"
 )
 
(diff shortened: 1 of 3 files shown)


Kamran64:
Following our discussions, this is the behaviour that the PR alters:

appsody run/test/debug = No change in docker. We no longer expose the debug port in Codewind (On the APPSODY_K8S_EXPERIMENTAL path and check for the existence of CODEWIND_PROJECT_ID so that we know we're in a Codewind environment. (remove debug ports from app-deploy.yaml and app-service.yaml) (k8s scenario)

This will only work if a stack has declared the APPSODY_DEBUG_PORT environment variable. Note, we only support one value for this variable now.

PR #1001, opened 2020-05-28
Lint APPSODY_DEBUG_PORT environment variable

Kamran64:
Adds the `APPSODY_DEBUG_PORT` variable to the linter as a warning.

### DIFF
--- a/cmd/stack_lint_dockerfile_stack.go
+++ b/cmd/stack_lint_dockerfile_stack.go
@@ -69,6 +69,7 @@ func lintDockerFileStack(log *LoggingConfig, stackPath string) (int, int) {
 	dockerfileStack := getENVDockerfile(log, stackPath)
 
 	variableFound := false
+	appsodyDebugFound := false
 	variable := ""
 
 	for i := 0; i < len(optionalEnvironmentVariables); i++ {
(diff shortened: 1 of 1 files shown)


cmd/stack_lint_dockerfile_stack.go
```diff
@@ -79,8 +79,12 @@ func lintDockerFileStack(log *LoggingConfig, stackPath string) (int, int) {
 			}
 		}
 		if !variableFound {
-			log.Warning.log("Missing ", variable)
-			stackLintWarningCount++
+			if optionalEnvironmentVariables[i] == "APPSODY_DEBUG_PORT" {
+				log.Warning.log("Missing APPSODY_DEBUG_PORT. If your stack supports the appsody debug command, please define this variable.")
+			} else {
+				log.Warning.log("Missing ", variable)
+				stackLintWarningCount++
```
EnriqueL8, line 86: you may want to increment the counter in both cases?

cmd/stack_lint_dockerfile_stack.go
```diff
@@ -79,8 +79,12 @@ func lintDockerFileStack(log *LoggingConfig, stackPath string) (int, int) {
 			}
 		}
 		if !variableFound {
-			log.Warning.log("Missing ", variable)
-			stackLintWarningCount++
+			if optionalEnvironmentVariables[i] == "APPSODY_DEBUG_PORT" {
+				log.Warning.log("Missing APPSODY_DEBUG_PORT. If your stack supports the appsody debug command, please define this variable.")
```
EnriqueL8, line 83: You could check for `APPSODY_DEBUG` to be set? And only through the error if so? you could also concatinate the errmg and pass it to `log.Warning.log("Missing ", variable, errMsgDebug)`

Kamran64: This sounds like a better way of doing it, done 👍

cmd/stack_lint_dockerfile_stack.go
```diff
@@ -77,9 +78,16 @@ func lintDockerFileStack(log *LoggingConfig, stackPath string) (int, int) {
...
 				variableFound = true
 			}
+			if k == "APPSODY_DEBUG" {
+				appsodyDebugFound = true
+			}
 		}
 		if !variableFound {
-			log.Warning.log("Missing ", variable)
+			errMsg := ""
+			if appsodyDebugFound == true && optionalEnvironmentVariables[i] == "APPSODY_DEBUG_PORT" {
```
EnriqueL8, line 87: you can just do `if appsodyDebugFound` instead of `== true`, sorry for the nitpick

Kamran64: Np, i think the go linter fails (the irony) with that anyway

PR #1009, opened 2020-07-19
Use format option to extract docker version

### LINKED ISSUE #1008: appsody init fails (with Stack Check Requirement) with (at least) Docker 19.03.12
With docker 19.03.12 on macOS 10.15.2, you get the following (chose any stack)

```
appsody init incubator/nodejs-express
Checking stack requirements...
Appsody requirements met
[Error] The required version of Docker to use this stack is >= 17.09.0 - Please upgrade.
[Error] One or more technologies need upgrading to use this stack. Upgrades required: 1
```

What is really odd is this works in Docker 19.03.08.

I don't know what other environments may have this issue.

Here is my Docker config:

```
docker version
Client: Docker Engine - Community
 Azure integration  0.1.7
 Version:           19.03.12
 API version:       1.40
 Go version:        go1.13.10
 Git commit:        48a66213fe
 Built:             Mon Jun 22 15:41:33 2020
 OS/Arch:           darwin/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.12
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.13.10
  Git commit:       48a66213fe
  Built:            Mon Jun 22 15:49:27 2020
  OS/Arch:          linux/amd64
  Experimental:     true
 containerd:
  Version:          v1.2.13
  GitCommit:        7ad184331fa3e55e52b890ea95e65ba581ae3429
 runc:
  Version:          1.0.0-rc10
  GitCommit:        dc9208a3303feef5b3839f4323d9beb36df0a9dd
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

henrynash:
The current code to extract and check the minimum docker version is unsafe, since it assumes that the version is the first x.y.z sequence to appear in the output of the `docker version` command. This is no longer true, for instance with docker 19.03.12 on macOS (where the first version listed in the output turns our to be the Azure integration version!). This causes all `appsody init` commands to fail if the stack in question specifies a minimal Docker version in its stack.yaml.

While there is perhaps a larger conversation to be had about whether the correct goal is to check the client or server version of docker (or both?), this fix simply makes the current attempt to check the docker client version safe.

Fixes #1008

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -2401,7 +2401,12 @@ func CheckStackRequirements(log *LoggingConfig, requirementArray map[string]stri
 			}
 			runVersionCmd = VERSION
 		} else {
-			cmd := exec.Command(strings.ToLower(technology), "version")
+			var cmd *exec.Cmd
+			if strings.ToLower(technology) == "docker" {
+				cmd = exec.Command(strings.ToLower(technology), "version", "--format", "{{.Client.Version}}")
+			} else {
+				cmd = exec.Command(strings.ToLower(technology), "version")
+			}
 			runVersionCmd, err = SeparateOutput(cmd)
 			if err != nil {
 				log.Error.log(err, " - Are you sure ", technology, " is installed?")


scottkurz:
@henrynash @neeraj-laad - I was able to recreate this on Win 10 and then verify this patch resolved the issue.

Docker Desktop v.2.4.0.0
Engine 19.03.13

neeraj-laad (approved):
LGTM

